PR #56, opened 2025-12-28
feat: Implement dynamic OpenRouter fallback for pricing

cantalupo555:
## Summary
This PR implements a dynamic fallback mechanism for model pricing when LiteLLM doesn't yet support newly released models.

## Problem
New AI models (e.g., glm-4.7 released recently) don't have pricing data in LiteLLM yet, causing Tokscale to show $0.00 costs for these models. Waiting for LiteLLM updates (which can take weeks or months) isn't ideal.

## Solution
Implemented dynamic fallback to OpenRouter's endpoints API that:
- Fetches pricing **only when needed** (lazy loading)
- Retrieves pricing from the **model's author provider** (not cheapest)
- Caches results **incrementally** (only models you've used)
- **No manual overrides** - pricing always fetched dynamically

### Code Changes (`packages/cli/src/pricing.ts`)
- Added TypeScript interfaces for OpenRouter endpoints API response
- New `fetchModelEndpoints(author, slug)` function to fetch specific model endpoints
- Modified `fetchOpenRouterPricing(modelID?)` to support lazy loading:
  - With modelID: fetches only that model
  - Without modelID: fetches all mapped models (backward compatibility)
- Updated `getModelPricingWithFallback()` to call fetch with specific model
- Removed `OPENROUTER_PRICING_OVERRIDES` constants (no longer needed)
- Simplified `convertOpenRouterToLiteLLM()` (removed override logic)

### Documentation (`README.md`)
- Updated Features section to mention automatic OpenRouter fallback
- Expanded Pricing section with Dynamic Fallback explanation
- Documented caching behavior (LiteLLM + OpenRouter incremental cache)

### Before
```typescript
// Always fetch ALL models via /api/v1/models (expensive)
fetch("https://openrouter.ai/api/v1/models")
```

### After
```typescript
// Lazy: fetch only when needed
getModelPricingWithFallback("glm-4.7")
  → LiteLLM doesn't have it
  → Fetch: /api/v1/models/z-ai/glm-4.7/endpoints
  → Filter: provider_name === "Z.AI" (author provider)
  → Extract: pricing ($0.60/$2.20 + $0.11 cache)
  → Cache: ~/.cache/tokscale/openrouter-pricing.json
```

## Testing

Tested with glm-4.7 model:
- ✅ Fetches pricing from Z.AI (author provider) - not cheapest
- ✅ Correct values: Input $0.60/M, Output $2.20/M, Cache $0.11/M
- ✅ Cache works (second call uses cached data)
[...]

PR #66, opened 2025-12-30
fix(pricing): use author provider pricing from OpenRouter endpoints API

cantalupo555:
## Summary
- Fix OpenRouter pricing to use author provider prices instead of cheapest aggregated prices
- Fetch `/api/v1/models/{id}/endpoints` API and filter by `provider_name` matching the model author
- Extract complete pricing including `cache_read` from author endpoint

## Problem
The previous implementation used `/api/v1/models` which returns aggregated pricing (cheapest provider). This caused incorrect pricing for models like `glm-4.7` where the cheapest provider (Chutes) has different rates than the author provider (Z.AI).

### Before (using cheapest provider - Chutes)
| Field | Value |
|-------|-------|
| Input | $0.40/M |
| Output | $1.50/M |
| Cache Read | ❌ Not available |

### After (using author provider - Z.AI)
| Field | Value |
|-------|-------|
| Input | $0.60/M |
| Output | $2.20/M |
| Cache Read | $0.11/M ✅ |

## Changes
- Fetch model list from `/api/v1/models`
- For each model with known author, fetch `/api/v1/models/{id}/endpoints`
- Filter endpoints by `provider_name` matching author (e.g., Z.AI, xAI, Google, Anthropic)
- Extract pricing including `input_cache_read` from author endpoint
- Parallel fetching with semaphore (10 concurrent requests)

## Related
Continues the work from #56 (original TypeScript implementation) that was migrated to Rust in #64.

### DIFF
--- a/packages/core/src/pricing/openrouter.rs
+++ b/packages/core/src/pricing/openrouter.rs
@@ -11,9 +11,7 @@ const MAX_RETRIES: u32 = 3;
 const INITIAL_BACKOFF_MS: u64 = 200;
 const MAX_CONCURRENT_REQUESTS: usize = 10;
 
-// =============================================================================
-// Structs for /api/v1/models endpoint (list all models)
-// =============================================================================
+/// Structs for `/api/v1/models` endpoint (list all models).
 
 #[derive(Deserialize)]
 struct ModelListItem {
@@ -25,9 +23,7 @@ struct ModelsListResponse {
     data: Vec<ModelListItem>,
 }
 
-// =============================================================================
-// Structs for /api/v1/models/{id}/endpoints endpoint (author pricing)
-// =============================================================================
+/// Structs for `/api/v1/models/{id}/endpoints` endpoint (author pricing).
 
 #[derive(Deserialize)]
 struct EndpointPricing {
@@ -57,11 +53,10 @@ struct EndpointsResponse {
     data: EndpointData,
 }
 
-// =============================================================================
-// Mapping from model ID prefix to author provider name
-// =============================================================================
-
-/// Maps the model ID prefix (e.g., "z-ai") to the provider_name in endpoints API (e.g., "Z.AI")
+/// Model ID prefix to provider name mapping.
+///
+/// Translates model ID prefixes like `z-ai` to their corresponding
+/// provider names in the endpoints API, such as `Z.AI`.
 fn get_author_provider_name(model_id: &str) -> Option<&'static str> {
     let prefix = model_id.split('/').next()?;
     
@@ -77,7 +72,7 @@ fn get_author_provider_name(model_id: &str) -> Option<&'static str> {
         "qwen" => Some("Alibaba"),
         "cohere" => Some("Cohere"),
         "perplexity" => Some("Perplexity"),
-        "moonshotai" => Some("Moonshot"),
+        "moonshotai" => Some("Moonshot AI"),
         _ => None,
     }
 }


cantalupo555:
## Technical Notes

### ✅ Benefits

1. **Accurate pricing** - Uses author provider rates instead of cheapest aggregated
2. **Cache pricing support** - Extracts `cache_read` and `cache_write` when available
3. **Rate limit protection** - Semaphore limits to 10 concurrent requests
4. **Smart filtering** - Only fetches endpoints for models with known authors

### ⚠️ Trade-offs & Considerations

1. **Performance**: Changes from 1 API call to potentially hundreds (one per model with known author). Mitigated by parallel fetching with semaphore control.

2. **Limited coverage**: Only 12 provider prefixes mapped (`z-ai`, `x-ai`, `anthropic`, `openai`, `google`, `meta-llama`, `mistralai`, `deepseek`, `qwen`, `cohere`, `perplexity`, `moonshotai`). Models from unmapped providers won't have author pricing.

3. **Silent failures**: If the `/endpoints` API fails for a specific model, it returns `None` silently and the model is excluded from results.

4. **No fallback**: If the author endpoint doesn't exist or returns no data, the model is skipped entirely. The previous implementation at least had aggregated pricing as a baseline.

---

These trade-offs are acceptable for the current use case since accurate author pricing is more valuable than having fallback data from cheaper providers.

junhoyeo:
Thanks @cantalupo555! I'll take a fast look

packages/core/src/pricing/openrouter.rs
```diff
@@ -19,14 +40,46 @@ struct ModelPricingResponse {
...
-    data: Vec<Model>,
+struct EndpointsResponse {
+    data: EndpointData,
+}
+
+// =============================================================================
+// Mapping from model ID prefix to author provider name
+// =============================================================================
+
+/// Maps the model ID prefix (e.g., "z-ai") to the provider_name in endpoints API (e.g., "Z.AI")
```
junhoyeo:
```suggestion
/// Model ID prefix to provider name mapping.
///
/// Translates model ID prefixes like `"z-ai"` to their corresponding
/// provider names in the endpoints API, such as `"Z.AI"`.
```

junhoyeo:
I see `bun cli pricing z-ai/glm-4.7` still fetching old data so I think I should do more debugging/research before merging this, thanks tho!

```
  Pricing for: z-ai/glm-4.7
  Matched key: z-ai/glm-4.7
  Source: OpenRouter

  Input:  $0.40 / 1M tokens
  Output: $1.50 / 1M tokens
```

packages/core/src/pricing/openrouter.rs
cantalupo555: @junhoyeo Done 🚀

packages/core/src/pricing/openrouter.rs
```diff
@@ -11,9 +11,7 @@ const MAX_RETRIES: u32 = 3;
 const INITIAL_BACKOFF_MS: u64 = 200;
 const MAX_CONCURRENT_REQUESTS: usize = 10;
 
-// =============================================================================
-// Structs for /api/v1/models endpoint (list all models)
-// =============================================================================
+/// Structs for `/api/v1/models` endpoint (list all models).
```
junhoyeo, line 14: 👍👍 thx

packages/core/src/pricing/openrouter.rs
```diff
@@ -77,7 +72,7 @@ fn get_author_provider_name(model_id: &str) -> Option<&'static str> {
         "qwen" => Some("Alibaba"),
         "cohere" => Some("Cohere"),
         "perplexity" => Some("Perplexity"),
-        "moonshotai" => Some("Moonshot"),
+        "moonshotai" => Some("Moonshot AI"),
```
junhoyeo, line 75: Image

👍👍 Provider Name is `Moonshot AI` good (note: different with model prefix i see)

cantalupo555: @junhoyeo I haven't tested it with the adjustment you suggested, but without that requested change, it's working =D
image

image

PR #68, opened 2025-12-30
fix: add --platform flag to napi build for local development

hellosunghyun:
## Problem

On ARM Mac, running `bun run cli` after `bun run build:core` fails with:

```
Error: Native module required. Run: bun run build:core
```

## Cause

- `napi build --release` only generates `tokscale-core.node`
- `index.js` looks for platform-specific filename (`tokscale-core.darwin-arm64.node`)
- Filename mismatch causes module load failure on local builds

## Fix

Add `--platform` flag to `build` script so it generates platform-specific filenames

```diff
- "build": "napi build --release",
+ "build": "napi build --platform --release",
```

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -28,7 +28,7 @@
   },
   "scripts": {
     "artifacts": "napi artifacts",
-    "build": "napi build --release",
+    "build": "napi build --platform --release",
     "build:platform": "napi build --platform --release",
     "build:debug": "napi build",
     "prepublishOnly": "napi prepublish -t npm",


cantalupo555:
Thanks for fixing this! I had also worked on this same fix in my fork:

- Branch: https://github.com/cantalupo555/tokscale/tree/fix/napi-build-platform-flag
- Commit: https://github.com/junhoyeo/tokscale/commit/c4113fdf2c1e02a0b60bc515adb1df8dcb70dab1

Great timing! 👍

PR #72, opened 2025-12-30
fix(tui): force dark background for light terminal themes

hellosunghyun:
## Summary
- Add `backgroundColor="black"` to TUI root container
- Fix text visibility issue on light terminal themes

## Problem
TUI did not specify a background color, inheriting the terminal's default background. On light themes, gray text (`#666666`, `#AAAAAA`, `dim`) becomes invisible.

## Solution
Force black background on the root container so text is visible regardless of terminal theme.

### DIFF
--- a/packages/cli/src/tui/App.tsx
+++ b/packages/cli/src/tui/App.tsx
@@ -345,7 +345,7 @@ export function App(props: AppProps) {
   };
 
   return (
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
This definitely needed, thanks a lot! @hellosunghyun

PR #74, opened 2025-12-30
fix(ci): commit auto-generated NAPI-RS files in publish workflow

junhoyeo:
## Summary

- Fix publish workflow to commit `index.js` and `index.d.ts` after version bump
- These files are auto-generated by NAPI-RS with hardcoded version strings

## Problem

The publish workflow was not committing the regenerated NAPI-RS files (`index.js`, `index.d.ts`), causing version mismatch between:
- `package.json` (correctly bumped to new version)
- `index.js` (still has old version strings)

## Root Cause

1. Build jobs regenerate `index.js`/`index.d.ts` with new version via `napi build`
2. But only `.node` binaries were uploaded as artifacts
3. Finalize job only committed `package.json` files

## Fix

1. Upload `index.js` + `index.d.ts` as artifact from one build job
2. Download in `publish-core` (so npm gets correct version)
3. Download in `finalize` (so git gets correct version)
4. Add both files to git commit


junhoyeo:
Fixes issue in https://github.com/junhoyeo/tokscale/commit/6214e2b816b6f08a7ca1c25f197a0c551d483690

junhoyeo:
Added https://github.com/junhoyeo/tokscale/commit/4f675e9547b2581fa4fd23c4c2370b8da78e5505

PR #89, opened 2026-01-01
fix: codex double-counting + add fallback suffixes for codex models

johnpyp:
- fixes codex double-counting cached input tokens as normal input tokens, because codex logs report input tokens INCLUDING the cached input tokens
- adds "fallback suffixes" which make it so that a model like `gpt-5.2-codex` is priced the same as `gpt-5.2` if `gpt-5.2-codex` isn't found in LiteLLM yet
  - gpt-5.2-codex isn't available on the api, so it doesn't **technically** have a price, but we can be pretty confident it'll cost the same as gpt-5.2. **If** it doesn't, that price will still be automatically updated in future calculations once litellm has it


junhoyeo:
@johnpyp Thanks for the contribution!

Issue #57, opened 2025-12-28
Inconsistency in cost calculation


cantalupo555:
I noticed another point now, analyzing the profiles on the Leaderboard (tokscale.ai), we find highly inflated values.
This image model is not calculating the “Cache Read” value, as support for this model does not have the `“cache_read_input_token_cost”` parameter defined in LiteLLM.

It is related to PR: 
https://github.com/junhoyeo/tokscale/pull/56
https://github.com/BerriAI/litellm/pull/18476

<img width="753" height="291" alt="Image" src="https://github.com/user-attachments/assets/b9ff3d50-ebb7-426f-9762-15a097df521b" />

junhoyeo:
> Since PR [56](https://github.com/junhoyeo/tokscale/pull/56) has been merged. Would it be possible to combine ways of comparing fields, if LiteLLM does not have OpenRouter, it can provide it.
> 
> Could this new issue be related? [#62](https://github.com/junhoyeo/tokscale/issues/62)
> 
> Perhaps it would be better to wait for the next steps from [@junhoyeo](https://github.com/junhoyeo)

@cantalupo555 First of all, thank you for all the PRs! I haven't published the changes yet because I think the design needs more work. Let me explain the current architecture and the improvements I'm working on:

Tokscale has both TypeScript and Rust components. Session parsing is CPU-intensive (scanning thousands of JSON files), so it runs in Rust for performance.

Current flow:

1. TS fetches LiteLLM pricing and caches to disk
2. TS fetches OpenRouter for specific hardcoded model mappings (e.g., `glm-4.7-free` → `z-ai/glm-4.7`) for models not in LiteLLM
3. TS passes pricing entries to Rust
4. Rust parses sessions and does lookups with fuzzy matching (normalization logic duplicated in both TS and Rust)

Our model pricing resolver works well - model naming is becoming more standardized across the industry, so fuzzy matching handles most cases.

But the issue is that OpenRouter fallback only applies to hardcoded mappings. If a model ID appears in user sessions but isn't in LiteLLM AND isn't in our hardcoded mapping, we don't attempt an OpenRouter fallback.

To maintain a single source of truth and improve efficiency, I'm moving all pricing logic to Rust:
- Rust will fetch LiteLLM pricing (with disk caching)
- Rust will parse sessions and collect unique model IDs
- For any model ID missing from LiteLLM, Rust will dynamically attempt OpenRouter fallback
[...]

PR #91, opened 2026-01-03
feat(pricing): support antigravity-auth plugin prefix stripping

### LINKED ISSUE #90: Antigravity plugin for opencode not calculating price properly
A recent antigravity auth plugin for Opencode (https://github.com/NoeFabris/opencode-antigravity-auth) modified the naming scheme to need appending 'antigravity-' in front of the model name. On tokscale, token counts etc are still counted correctly but pricing is 0. Is it possible to support this change?

AndreDalwin:
## Summary

Adds support for the [opencode-antigravity-auth](https://github.com/NoeFabris/opencode-antigravity-auth) plugin which prefixes model names with `antigravity-` for quota routing.

- Add configurable `STRIPPED_PREFIXES` constant for routing prefixes
- Add `strip_routing_prefix()` with case-insensitive matching  
- Strip prefix before alias resolution in `lookup_with_source()`
- Add 8 new tests for prefix stripping scenarios

## Before/After

| Model | Before | After |
|-------|--------|-------|
| `antigravity-gemini-3-flash` | $0.00 | $9.45 |
| `antigravity-gemini-3-pro` | $0.00 | $3.10 |

Original model names are preserved in the display output.

## Extensibility

To add more routing prefixes in the future:
```rust
const STRIPPED_PREFIXES: &[&str] = &["antigravity-", "future-prefix-"];
```

Fixes #90

### DIFF
--- a/packages/core/src/pricing/lookup.rs
+++ b/packages/core/src/pricing/lookup.rs
@@ -3,18 +3,44 @@ use std::collections::HashMap;
 use std::sync::RwLock;
 
 const PROVIDER_PREFIXES: &[&str] = &[
-    "openai/", "anthropic/", "google/", "meta-llama/", "mistralai/", 
-    "deepseek/", "qwen/", "cohere/", "perplexity/", "x-ai/",
+    "openai/",
+    "anthropic/",
+    "google/",
+    "meta-llama/",
+    "mistralai/",
+    "deepseek/",
+    "qwen/",
+    "cohere/",
+    "perplexity/",
+    "x-ai/",
 ];
 
 const ORIGINAL_PROVIDER_PREFIXES: &[&str] = &[
-    "x-ai/", "xai/", "anthropic/", "openai/", "google/", "meta-llama/",
-    "mistralai/", "deepseek/", "z-ai/", "qwen/", "cohere/", "perplexity/", "moonshotai/",
+    "x-ai/",
+    "xai/",
+    "anthropic/",
+    "openai/",
+    "google/",
+    "meta-llama/",
+    "mistralai/",
+    "deepseek/",
+    "z-ai/",
+    "qwen/",
+    "cohere/",
+    "perplexity/",
+    "moonshotai/",
 ];
 
 const RESELLER_PROVIDER_PREFIXES: &[&str] = &[
-    "azure/", "azure_ai/", "bedrock/", "vertex_ai/", 
-    "together/", "together_ai/", "fireworks_ai/", "groq/", "openrouter/",
+    "azure/",
+    "azure_ai/",
+    "bedrock/",
+    "vertex_ai/",
+    "together/",
+    "together_ai/",
+    "fireworks_ai/",
+    "groq/",
+    "openrouter/",
 ];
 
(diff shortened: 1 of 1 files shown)


packages/core/src/pricing/lookup.rs
```diff
@@ -24,7 +50,14 @@ const MIN_FUZZY_MATCH_LEN: usize = 5;
...
 /// Note: OpenCode Zen uses -xhigh suffix for extra-high quality tier
-const TIER_SUFFIXES: &[&str] = &["-xhigh", "-low", "-high", "-medium", "-free", ":low", ":high", ":medium", ":free"];
+const TIER_SUFFIXES: &[&str] = &[
+    "-xhigh", "-low", "-high", "-medium", "-free", ":low", ":high", ":medium", ":free",
+];
+
+/// Routing/quota prefixes that should be stripped for pricing lookup.
+/// These indicate third-party routing systems (e.g., antigravity-auth plugin) but don't affect base model pricing.
+/// Configurable: Add new prefixes here as they emerge from community plugins.
+const STRIPPED_PREFIXES: &[&str] = &["antigravity-"];
```
cantalupo555, line 61: @AndreDalwin Thanks for the contribution! However, I have concerns about this approach.

This PR introduces hardcoded STRIPPED_PREFIXES constants, which contradicts the architectural direction we've been establishing:

#56 explicitly:
  - Removed OPENROUTER_PRICING_OVERRIDES constants
  - Stated "no manual overrides - pricing always fetched dynamically"
  - Listed as benefit: "No manual pricing constants needed"

#57 already raised concerns about relying on static data that becomes outdated.

#66 continued this approach:
  - Fetches author provider pricing dynamically from OpenRouter endpoints API
  - No hardcoded values

Defining values directly in the file requires a code change and release for each new plugin prefix - exactly what we're trying to move away from.

junhoyeo: @cantalupo555 

Thanks for raising this. I agree with the direction from #56/#66: we should not hardcode pricing values and should fetch pricing dynamically (LiteLLM/OpenRouter). 

This change is different in scope: `STRIPPED_PREFIXES` is not a pricing override. It’s an input normalization step for third-party routing prefixes that external plugins add for quota routing (e.g. `opencode-antigravity-auth` prefixes models with `antigravity-`).

Without stripping, the upstream pricing sources don’t recognize the prefixed model IDs, which leads to correct token counts but $0.00 pricing in Tokscale (#90). 

Also, this plugin is actively used in the ecosystem (e.g. [`oh-my-opencode`](https://github.com/code-yeongyu/oh-my-opencode)’s installation guide explicitly includes `opencode-antigravity-auth`).

cantalupo555: @junhoyeo You're right - this is input normalization, not pricing data.

However, the pricing should be fetched dynamically from OpenRouter/LiteLLM after stripping the prefix, rather than relying on hardcoded values.

I'll submit a follow-up PR to address this. **LGTM** 😅

junhoyeo: @cantalupo555 great to hear! i dont see hardcoded pricing in this pr tho (except the test mocks, which is good imo)

PR #97, opened 2026-01-05
adding copy button for join the leaderboard

makyinmars:
Added copy button for join the leaderboard

image


junhoyeo:
nice catch really

PR #218, opened 2026-02-18
fix(amp): use fallback timestamps for events missing timestamp field

### LINKED ISSUE #88: Amp numbers are very wrong
Image

Notice:
- Gemini has 0 input tokens
- None of them have cached read/write costs
- The input tokens numbers, even without cached read, are incorrect. Here is a screenshot from amp thread dashboard from Dec 7th.
	- Notice 24k Opus non-cached Input tokens > **all** the opus input tokens recorded in tokscale 

Image

Yeachan-Heo:
## Problem

Fixes #88 — Amp token numbers were very wrong / incomplete.

Root cause: `usageLedger` events without a `timestamp` field were silently dropped with `continue`, causing significant undercounting.

## Fix

Fall back in order:
1. Parsed RFC3339 timestamp from event
2. Thread `created` time (ms)
3. File mtime

No event is ever dropped due to missing timestamp. All 178 tests pass.

/cc @junhoyeo

### DIFF
--- a/packages/core/src/sessions/amp.rs
+++ b/packages/core/src/sessions/amp.rs
@@ -94,6 +94,14 @@ pub fn parse_amp_file(path: &Path) -> Vec<UnifiedMessage> {
         Err(_) => return Vec::new(),
     };
 
+    // Get file mtime as last-resort timestamp fallback
+    let file_mtime_ms = std::fs::metadata(path)
+        .ok()
+        .and_then(|m| m.modified().ok())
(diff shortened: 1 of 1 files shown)


junhoyeo:
Hey @Yeachan-Heo — this shipped in `tokscale@v2.0.0`! 🎉

Thanks for adding fallback timestamps for Amp events — this fixed the missing token counts that users were reporting.

PR #241, opened 2026-02-24
fix: show component missing fallback

nejcm:
On windows running the cli.ts fails. The reason is the Show component default fallback when not specified breaks on windows with:
```
440 |   }
441 |   if (anchor && anchor instanceof SlotRenderable) {
442 |     anchor = anchor.getSlotChild(parent);
443 |   }
444 |   if (isTextNodeRenderable(node)) {
445 |       throw new Error(`Orphan text error: "${node.toChunks().map((c) => c.text).join("")}" must have a <text> as a parent: ${parent.id} above ${node.id}`);
                  ^
error: Orphan text error: "" must have a <text> as a parent: box-31 above text-node-82
      at _insertNode
```

This PR adds fallback to Show components without it.
PS: null fallback does not work as solid.js does `fallback || ''`.

---
## Summary by cubic
Fixes Windows CLI crash by adding explicit empty fallbacks to Solid’s Show components to prevent orphan text errors and blank screen. The CLI now renders correctly on Windows.

- **Bug Fixes**
  - Added fallback={<></>} to Show in Header, Footer, Legend, ModelRow, BarChart, OverviewView, and StatsView to avoid Solid’s default empty-string fallback.
  - Consolidated StatsView conditionals to use Show with fallback, preventing orphan text nodes.
  - Minor import/order tweaks and formatting; no behavior changes.

<sup>Written for commit 72e0e0eee21e81f1b0c4dd8d18b33217c3d5942e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/cli/src/tui/components/BarChart.tsx
+++ b/packages/cli/src/tui/components/BarChart.tsx
@@ -191,7 +191,7 @@ export function BarChart(props: BarChartProps) {
           <text dim>{isVeryNarrowTerminal() ? "    0│" : "     0│"}</text>
           <text dim>{getRepeatedString("─", axisWidth())}</text>
         </box>
-        <Show when={dateLabels().length > 0}>
+        <Show when={dateLabels().length > 0} fallback={<></>}>
           <box flexDirection="row">
             <text dim>{isVeryNarrowTerminal() ? "      " : "       "}</text>
             <text dim>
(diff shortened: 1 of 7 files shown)


nejcm:
@junhoyeo Seems this PR is not valid anymore after recent changes pushed to main except maybe the Footer component.
But I am running into issues on main now when running cli.ts file where auth.js and some other imports are not found?

PR #288, opened 2026-03-07
fix: use local dates for daily usage bucketing

### LINKED ISSUE #267: Local timezone isn't being used
Image

It's 2026-03-02 right now. But my daily token usage in the table is split across 2026-03-03 and 2026-03-02 which is incorrect. May be related to #145

IvGolovach:
Fix daily usage bucketing across UTC midnight boundaries.

`UnifiedMessage::date` was being derived in UTC during parsing, and the daily aggregation/filtering code uses that as the day key. That could split a single local day into two buckets when usage crossed UTC midnight.

This switches that date to the system local timezone at parse time, and updates the related local-day behavior for date shortcuts, streaks, contribution ranges, and the TUI `today` state.

Added a regression test for the Los Angeles case where two messages from the same local day were split across 2026-03-02 and 2026-03-03.

Fixes #267

### DIFF
--- a/crates/tokscale-cli/src/commands/wrapped.rs
+++ b/crates/tokscale-cli/src/commands/wrapped.rs
@@ -1,7 +1,7 @@
 use crate::{auth, cursor};
 use ab_glyph::{point, Font, FontArc, GlyphId, PxScale, ScaleFont};
 use anyhow::{Context, Result};
-use chrono::{Datelike, Duration, Local, NaiveDate, Utc};
+use chrono::{Datelike, Duration, Local, NaiveDate};
 use colored::Colorize;
 use image::{imageops::FilterType, Rgba, RgbaImage};
 use imageproc::drawing::draw_filled_circle_mut;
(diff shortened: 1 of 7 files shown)


junhoyeo (approved):
## ✅ APPROVE

Thorough review of all 7 changed files + comparison against main.

### Correctness
- Core fix is sound: absolute timestamp → `chrono::Local` datetime at parse time correctly resolves the UTC midnight boundary split (issue #267)
- All date-related codepaths updated consistently — verified no remaining `Utc::now().date_naive()` affecting day bucketing
- `chrono::Local` is the correct cross-platform API for OS-local time on macOS/Linux/Windows
- DST-safe since conversion is from absolute timestamps (no ambiguous local-time input parsing)

### Consistency
- CLI shortcuts (`--today/--week/--month`) → local date ✅
- TUI `jump_to_today`, daily highlight → local ✅
- Contribution graph window, streak calculations → local ✅
- Wrapped streak → local ✅

### Tests
- Regression test with `TZ=America/Los_Angeles` directly proves the fix (two UTC timestamps crossing midnight but same local day → one bucket, two messages)
- Deterministic test helpers (`build_date_filter_for_date`, `calculate_streaks_for_today`, etc.) reduce flakiness
- Fixture hardening with `XDG_CACHE_HOME` and pricing cache priming improves CI stability

### Backward Compatibility
- No schema/format break — `date` remains `YYYY-MM-DD` string, only bucketing semantics change

### Note (non-blocking)
- Bucket assignment reflects machine timezone at parse time. If a user travels across timezones, re-parsed data uses the new timezone. Acceptable semantic choice for a local CLI tool.

LGTM — clean fix with good test coverage.

junhoyeo:
🚀 Deployed in [`tokscale@v2.0.10`](https://github.com/junhoyeo/tokscale/releases/tag/v2.0.10). Thanks @IvGolovach!

PR #294, opened 2026-03-08
refactor(frontend): centralize personal token auth paths

IvGolovach:
Refactor personal token authentication to use a single shared implementation.

Previously token validation and issuance logic was duplicated across multiple routes (`/api/submit`, device auth polling, and settings token routes) as well as the `validateApiToken()` session helper. That made it possible for validation rules to drift between paths.

This change centralizes token issuance, validation, and token management helpers in `personalTokens.ts` and updates those routes to use the shared implementation.

There are no schema or client-facing API changes. Existing tokens and clients continue to work as before. The only behavior adjustment is consistent expiry handling (`expiresAt <= now`) across all token auth paths.

Added tests covering token issuance, validation, revocation, device auth flow, and `/api/submit` authentication so the shared logic stays consistent across routes.

### DIFF
--- a/packages/frontend/__tests__/api/devicePoll.test.ts
+++ b/packages/frontend/__tests__/api/devicePoll.test.ts
@@ -1,0 +1,154 @@
+import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockState = vi.hoisted(() => {
+  const selectResults: Array<Array<Record<string, unknown>>> = [];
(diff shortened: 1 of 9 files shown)


junhoyeo:
🚀 Deployed in [`tokscale@v2.0.10`](https://github.com/junhoyeo/tokscale/releases/tag/v2.0.10). Thanks @IvGolovach!

Issue #262, opened 2026-03-02
feat: split Synthetic into Octofriend (client) and API Gateway (re-attribution)


junhoyeo:
## Problem

Currently, tokscale lumps two fundamentally different concepts under a single "Synthetic" umbrella:

1. **Octofriend** — a standalone open-source CLI coding assistant ([`synthetic-lab/octofriend`](https://github.com/synthetic-lab/octofriend)) that stores sessions in `~/.local/share/octofriend/sqlite.db`
2. **Synthetic API Gateway** — an OpenAI/Anthropic-compatible API provider at [synthetic.new](https://synthetic.new/) (base URL: `https://api.glhf.chat/v1/`) that users can route *any* coding agent through

These are conceptually different in the same way that "Claude Code" (client) differs from "Anthropic" (API provider). Every other entry in tokscale's client table is a **client** (the tool users interact with), but "Synthetic" is an **API provider** — breaking the naming convention.

### How "Synthetic" works today (`crates/tokscale-core/src/sessions/synthetic.rs`, `lib.rs`)

**Strategy 1 — Re-attribution (API Gateway detection):**
Scans messages from ALL other clients and re-attributes `msg.client` to `"synthetic"` if:
- Model ID has `hf:` prefix (e.g., `hf:deepseek-ai/DeepSeek-V3-0324`)
- Model ID has `accounts/fireworks/` or `accounts/together/` prefix
- Provider ID matches `"synthetic"` | `"glhf"` | `"synthetic.new"` | `"octofriend"`

This **mutates** the message's client field in-memory (not on disk), moving it from its original client (e.g., "opencode") to "synthetic".

**Strategy 2 — Octofriend SQLite parsing:**
Parses `~/.local/share/octofriend/sqlite.db` for token usage data. Currently Octofriend only stores `input_history` (no token data), so this is future-proofed — checks for `messages`, `sessions`, or `token_usage` tables.

### The conflation problem

When a user runs `tokscale` (default, all clients):
- A message from OpenCode that used `synthetic.new` API gets re-attributed from `client: "opencode"` → `client: "synthetic"`
- This means the user's **OpenCode usage disappears** from the OpenCode category and shows up under "Synthetic"
- The user used OpenCode as their tool — the fact that it routed through Synthetic's API is a provider detail, not a client identity change

Compare: if someone uses Claude Code through AWS Bedrock vs Anthropic's API, tokscale doesn't create a separate "Bedrock" client category.

[...]

PR #301, opened 2026-03-09
fix: preserve client identity for synthetic gateway traffic

IvGolovach:
Preserve the original client for Synthetic gateway traffic.

When a message matched Synthetic gateway detection, tokscale was rewriting its `client` to `synthetic`. That makes usage disappear from the actual tool the user worked in (for example OpenCode), even though Synthetic is acting as a gateway/provider rather than the client.

This keeps the original client, still normalizes Synthetic gateway model/provider fields, and keeps the existing `--synthetic` filter working as a compatibility view for Synthetic-routed traffic. The same behavior is applied in the TUI data path.

Related to #262.

Checks:
- `cargo test -p tokscale-core --lib`
- `cargo test -p tokscale-cli tui::data -- --nocapture`

### DIFF
--- a/crates/tokscale-cli/src/tui/data/mod.rs
+++ b/crates/tokscale-cli/src/tui/data/mod.rs
@@ -965,9 +965,7 @@ mod tests {
             "unknown"
         ));
         assert!(!sessions::synthetic::matches_synthetic_filter(
-            "opencode",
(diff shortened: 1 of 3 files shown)


junhoyeo (approved):
Thanks a lot!

PR #327, opened 2026-03-13
perf(core): cache parsed source messages for refresh

IvGolovach:
## Summary
Speed up repeated refreshes by caching parsed local source messages and reusing unchanged results, while hardening the cache and incremental refresh path so correctness stays ahead of performance.

## Why
Refreshing local data currently reparses the full source history on every run, even when most files have not changed. That makes refresh noticeably slow on larger histories.

A cache only helps if it remains correct under partial appends, file rewrites, parser failures, and on-disk cache upgrades. The final shape of this PR therefore includes both the performance path and the correctness work needed to keep cache hits and incremental reuse safe.

## Diff scope
- add a persistent source-message cache for parsed local history
- reuse unchanged file results and incrementally parse appended Codex data when validation passes
- reprice cached messages on read so pricing remains current
- harden cache invalidation, cache schema upgrade behavior, and cache consistency across refreshes
- tighten fallback behavior so uncertain or unsafe incremental states fall back to full parsing instead of returning stale results
- add correctness coverage around cache reuse, parser fallback behavior, and upgrade/repair paths

## Test proof
- `cargo fmt --all`
  - passed
- `cargo test -p tokscale-core --lib`
  - `368 passed; 1 ignored`
- `cargo clippy --workspace --all-features -- -D warnings`
  - passed

## Verification-pack proof
Not applicable — core library and CLI refresh-path changes only.

## Migration notes
Not applicable — no schema or data migration.

## CI context confirmation
CI context names unchanged.

## Rollback plan
- no DB downgrade required

## Known residual risks
- this PR still introduces a broad review surface because the cache performance path and the correctness hardening landed together
- refresh behavior is intentionally conservative in ambiguous cases, so some cache-hit opportunities are sacrificed in favor of correctness

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -1212,6 +1212,31 @@ async fn load_pricing_for_local_parse() -> Option<Arc<pricing::PricingService>> 
     )
 }
 
+fn resolve_local_parse_request(
+    options: &LocalParseOptions,
+) -> Result<(String, Vec<String>), String> {
+    let home_dir = get_home_dir_string(&options.home_dir)?;
+    let clients = options.clients.clone().unwrap_or_else(|| {
+        let mut clients: Vec<String> = ClientId::iter()
+            .filter(|c| c.parse_local())
+            .map(|c| c.as_str().to_string())
+            .collect();
+        clients.push("synthetic".to_string());
+        clients
+    });
+    Ok((home_dir, clients))
+}
+
+fn parse_local_unified_messages_resolved(
+    options: LocalParseOptions,
+    home_dir: &str,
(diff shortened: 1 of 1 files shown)


IvGolovach:
Addressed the latest parser and cache review items in `617eebbf`.

This change preserves valid full-file messages when a later line read fails, keeps SQLite cache invalidation tied to `-wal` instead of transient shared-memory churn, and removes the extra predictable temp-path layer from the fallback cache directory.

The earlier cache-deserialization concern is still bounded on the current head via `with_limit(MAX_CACHE_FILE_BYTES)` in the load path.

@cubic-dev-ai please re-run the review on the latest head.

PR #376, opened 2026-03-31
fix: preserve multi-machine submissions on the leaderboard

haridigresses:
## Summary

This fixes the leaderboard overwrite case for users who submit from more than one machine.

Instead of merging every submit into a single `submissions` row per user, submissions can now be optionally source-scoped:
- submits with `meta.sourceId` write to a `(user_id, source_id)` row
- submits without `sourceId` keep the existing unsourced behavior
- leaderboard/profile/embed reads continue to aggregate across all of a user's submission rows

This keeps the fix modest and aligned with the existing schema and aggregation patterns, without introducing nested per-source instances into `daily_breakdown.source_breakdown`.

## Repro

Before this patch:
1. Submit usage for user `X` from machine A with `claude` activity.
2. Submit usage for the same user `X` from machine B with `claude` activity.
3. Observe that the later submission overwrites the earlier machine's contribution for overlapping days.

Expected behavior:
- both machines contribute to the user total
- re-submitting from one machine replaces only that source's prior contribution
- unsourced/older submissions keep current behavior

## What changed

- added optional `source_id` and `source_name` to `submissions`
- replaced single-row-per-user uniqueness with:
  - one unsourced row per user
  - many sourced rows per user
- updated `/api/submit` to target the correct submission row by source when provided
- updated profile, leaderboard, and embed queries to aggregate across a user's submission rows
- added CLI source metadata for `submit` payloads only
- documented `TOKSCALE_SOURCE_ID` and `TOKSCALE_SOURCE_NAME`
- added focused regression tests for source-scoped submit behavior and embed/profile aggregation

## Validation

Automated:
- `DATABASE_URL=postgres://localhost:5432/tokscale bun run db:generate`
- `bunx vitest run __tests__/api/submitSourceScope.test.ts __tests__/lib/getUserEmbedStats.test.ts __tests__/api/usersProfile.test.ts __tests__/api/submitAuth.test.ts __tests__/lib/getLeaderboardAllTime.test.ts __tests__/lib/getLeaderboard.test.ts`

Both passed.

## Notes

- `TOKSCALE_SOURCE_NAME` is stored per `sourceId`; a later submit with the same `sourceId` and a new non-empty name updates that source's display name.
- I did not build the Rust CLI in this environment because `cargo` was unavailable here.
- `bunx tsc --noEmit` still fails on an unrelated existing import in `packages/frontend/src/components/BlackholeHero.tsx`.

---
## Summary by cubic
Preserves multi‑machine submissions by scoping per source and aggregating totals across a user’s sources. Adds robust CLI source ID persistence with a stale‑lock cleanup fallback and normalizes profile/leaderboard/embed aggregates.

- **Bug Fixes**
  - `submissions` are source‑scoped: rows keyed by `(user_id, source_id)`; unsourced rows remain; empty/whitespace `sourceId`/`sourceName` are ignored; optional source metadata failures are tolerated.
  - `/api/submit` targets the correct per‑source row, updates `sourceName` only when provided, handles missing/invalid source metadata, and computes response metrics after commit.
  - Profile, leaderboard, and embed aggregate by summing across all a user’s rows; `submissionCount` uses sums; `updatedAt` uses the latest submit; embed ranks are normalized from string results.
  - `crates/tokscale-cli` includes `sourceId`/`sourceName` in submit meta, persists a stable per‑machine source ID with a small file lock, and cleans up stale locks by age if needed; supports `TOKSCALE_SOURCE_ID`, `TOKSCALE_SOURCE_NAME`, and `TOKSCALE_API_URL`.

- **Migration**
  - Run DB migration 0005 to add `source_id`/`source_name` and replace single‑row uniqueness with per‑source partial unique indexes.
  - No changes needed for existing unsourced submits; reads aggregate across rows.

<sup>Written for commit 9eda26d6302e1acc30ace7620b00ce3e2265a4cf. Summary will update on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -493,10 +493,17 @@ Environment variables override config file values. For CI/CD or one-off use:
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `TOKSCALE_NATIVE_TIMEOUT_MS` | `300000` (5 min) | Overrides `nativeTimeoutMs` config |
+| `TOKSCALE_API_URL` | `https://tokscale.ai` | Overrides the API base URL used by `login` and `submit` |
+| `TOKSCALE_SOURCE_ID` | auto-generated per machine | Overrides the stable source ID attached to `submit` payloads |
+| `TOKSCALE_SOURCE_NAME` | `CLI on <hostname>` | Overrides the human-readable source name attached to `submit` payloads |
 
 ```bash
 # Example: Increase timeout for very large datasets
 TOKSCALE_NATIVE_TIMEOUT_MS=600000 tokscale graph --output data.json
+
+# Example: submit two local test sources without using two machines
+TOKSCALE_SOURCE_ID=machine-a tokscale submit
+TOKSCALE_SOURCE_ID=machine-b TOKSCALE_SOURCE_NAME="Work Laptop" tokscale submit
 ```
 
 > **Note**: For persistent changes, prefer setting `nativeTimeoutMs` in `~/.config/tokscale/settings.json`. Environment variables are best for one-off overrides or CI/CD.
--- a/README.md
+++ b/README.md
@@ -506,6 +506,8 @@ TOKSCALE_SOURCE_ID=machine-a tokscale submit
 TOKSCALE_SOURCE_ID=machine-b TOKSCALE_SOURCE_NAME="Work Laptop" tokscale submit
 ```
 
+> `TOKSCALE_SOURCE_NAME` is stored per `sourceId`. A later submit with the same `sourceId` and a new non-empty `TOKSCALE_SOURCE_NAME` updates the stored display name for that source.
+
 > **Note**: For persistent changes, prefer setting `nativeTimeoutMs` in `~/.config/tokscale/settings.json`. Environment variables are best for one-off overrides or CI/CD.
 
 ### Headless Mode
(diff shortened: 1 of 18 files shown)


junhoyeo:
Hi @haridigresses — thanks for the thorough work here, including the repro and the regression tests. Closing as superseded, with the technical context:

Main has since fixed the multi-machine overwrite **at a different layer**. Instead of splitting `submissions` into per-source rows keyed by `(user_id, source_id)`, each user keeps a single submission row and the per-machine scoping happens one level down: `daily_breakdown` rows carry a `submitted_device_id` referencing a `submitted_devices` table, and the CLI sends a persistent UUID device id with every submit (#517, #593, #609). Cross-machine totals are summed at write time, and a resubmit from one machine replaces only that machine's day rows — so the leaderboard, profile, and embed read paths needed no changes at all.

That makes this PR's row model structurally incompatible with current main: the submit route locks "the user's submission row" (singular), recomputes totals across the whole submission, and the group leaderboards / rank caches all assume one row per user. The `0005_narrow_thena.sql` migration is also unrescuable — the migration chain is now well past 0014, and the journal can't be retrofitted.

The one residual idea — letting users name a source — lives on as `TOKSCALE_DEVICE_NAME` plus the device rename API (`/api/settings/devices`).

Appreciate the contribution — the failing scenarios you documented are exactly the ones the shipped model is tested against.

Issue #378, opened 2026-03-31
Feature Request: Multi-device stats aggregation for the same account


Fullstop000:
Support aggregating token usage stats from multiple devices under the same Tokscale account, so that profiles, leaderboards, and Wrapped reports reflect total usage across all linked devices.

Related PRs:
- #329 feat: multi-device support with UUID device id and multi device TUI aggregation
- #252 feat: cross-machine usage aggregation with device-level dedup
- #55 feat: cross-machine usage aggregation + automatic sync
- #376 fix: preserve multi-machine submissions on the leaderboard
- #313 fix(submit): preserve per-device cross-machine breakdowns

jakublala:
Yes, please! #442

agustinusnathaniel:
Would love to see this feature landed / catered

pinion05:
## Design proposal: device-aware day/client merge semantics

Related symptom report: #248

### Summary

`tokscale submit` currently has no device identity in the submission payload. The effective merge key is currently:

```txt
userId + date + client
```

This works for re-submitting from one machine, but it can overwrite usage when the same account submits from multiple machines using the same client on the same date.

Recommended target semantics:

```txt
userId + deviceId + date + client
```

Meaning:

- same device + same date + same client: replace, preserving idempotent re-submit behavior
- different device + same date + same client: preserve both and aggregate in user/profile/leaderboard totals

### Current submit payload

The CLI payload is currently shaped around aggregate daily contribution data:

```ts
{
  meta: {
    generatedAt: string,
    version: string,
    dateRange: { start: string, end: string }
  },
  summary: {
    totalTokens: number,
    totalCost: number,
    totalDays: number,
    activeDays: number,
    averagePerDay: number,
    maxCostInSingleDay: number,
    clients: string[],
    models: string[]
  },
  years: Array<{
    year: string,
    totalTokens: number,
    totalCost: number,
    range: { start: string, end: string }
  }>,
  contributions: Array<{
    date: string,
    timestampMs?: number,
    totals: { tokens: number, cost: number, messages: number },
    intensity: number,
    tokenBreakdown: {
      input: number,
      output: number,
      cacheRead: number,
      cacheWrite: number,
      reasoning: number
    },
    clients: Array<{
      client: string,
      modelId: string,
      providerId?: string,
      tokens: TokenBreakdown,
      cost: number,
      messages: number
    }>
  }>
}
```

There is no `deviceId`, `deviceName`, or other machine identity in this payload today.

[...]

PR #388, opened 2026-04-01
fix(submit): preserve source-scoped multi-machine submissions

IvGolovach:
## Summary
- preserve usage from multiple machines by storing submissions per source instead of collapsing all submits into a single row per user
- upgrade the first source-aware submit from a legacy account in place to avoid duplicate totals during cutover
- aggregate profile, embed, and leaderboard reads across all submission rows for the same user

## Why
Tokscale currently stores a single submission row per user. That means a later submit from another machine can overwrite usage for the same client instead of aggregating it, which breaks profiles, leaderboard totals, and public stats for users who submit from more than one device.

## Diff scope
- add `sourceId` and `sourceName` to submit payload validation and CLI submission metadata
- add source-scoped submission storage with additive migration `0005_tan_rumiko_fujikawa`
- upgrade the existing unsourced row in place on the first source-aware submit to avoid legacy duplication
- reject ambiguous unsourced submits with `409` once a user has entered source-scoped mode
- aggregate profile, embed, and leaderboard reads across all submission rows
- add regression tests for scope resolution, profile/embed aggregation, and auth/submit edge cases

## Test proof
- `cargo test -p tokscale-cli`
  - unit tests: `314 passed, 0 failed, 1 ignored`
  - integration tests: `74 passed, 0 failed`
- `cargo clippy -p tokscale-cli --all-features -- -D warnings`
  - passed
- `bun x vitest run __tests__/api/*.test.ts __tests__/lib/*.test.ts`
  - `17 files, 123 tests passed`
- `cargo fmt --all --check`
  - passed
- `git diff --check`
  - passed

## Migration notes
- adds migration `0005_tan_rumiko_fujikawa`
- additive schema change: introduces `source_id` and `source_name` plus source-scoped uniqueness
- legacy single-row accounts are upgraded in place on the first source-aware submit
- no destructive backfill is required up front

## Rollback plan
- rollback would require a follow-up schema migration before removing the new columns or indexes in production

## Known residual risks
- older clients that continue submitting without source identity will receive `409` after the account has entered source-scoped mode
- this PR does not add a new UI for managing or renaming sources beyond the submitted `sourceName`

## Closes
- #378

### DIFF
--- a/crates/tokscale-cli/src/auth.rs
+++ b/crates/tokscale-cli/src/auth.rs
@@ -3,7 +3,9 @@ use serde::{Deserialize, Serialize};
 use std::fs;
 use std::io::IsTerminal;
 use std::io::Write;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::thread;
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 fn home_dir() -> Result<PathBuf> {
     dirs::home_dir().context("Could not determine home directory")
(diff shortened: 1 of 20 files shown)


jakublala:
What is the update on this feature? Would love it!

IvGolovach:
> What is the update on this feature? Would love it!

Thanks for checking in- yes, this is still active, and I really appreciate the interest in this feature.

The implementation from this PR was carried forward into #389, which is the current landing branch for it. It preserves the original authored work from here, and that is the branch to watch for merge progress :)

PR #416, opened 2026-04-12
feat(embed): add isometric 3D contribution graph view (view=3d)

SeaL773:
## Summary
- Adds a `view=3d` query parameter to the existing embed SVG endpoint (`/api/embed/<username>/svg`)
- Renders an isometric 3D contribution graph as pure SVG `<polygon>` elements — no Canvas, no WebGL, no new dependencies
- Supports dark/light themes, matching the existing embed visual language

## Usage

```markdown
[![Tokscale 3D](https://tokscale.ai/api/embed/<username>/svg?view=3d)](https://tokscale.ai/u/<username>)
[![Tokscale 3D Light](https://tokscale.ai/api/embed/<username>/svg?view=3d&theme=light)](https://tokscale.ai/u/<username>)
```

## Implementation Details

**Rendering approach**: 2:1 dimetric projection (matching the frontend's obelisk.js visual style). Each contribution day is a cube with three shaded SVG polygon faces (top/left/right), drawn back-to-front via painter's algorithm. Heights map to the 5 contribution intensity levels.

**Files changed**:
| File | Change |
|------|--------|
| `packages/frontend/src/lib/embed/renderIsometric3DSvg.ts` | New renderer — pure SVG isometric 3D graph |
| `packages/frontend/src/app/api/embed/[username]/svg/route.ts` | Added `view=3d` parameter routing |
| `packages/frontend/__tests__/lib/renderIsometric3DSvg.test.ts` | Tests for the new renderer |

**Zero breaking changes**: Existing embed behavior is completely unchanged when `view` is not `3d`.

## Motivation

The 3D contribution graph on the tokscale.ai frontend is one of the most visually distinctive features, but it can't be embedded in GitHub Profile READMEs (it requires Canvas/obelisk.js). This PR brings that visual as a static SVG image that works anywhere `` tags work.

---
## Summary by cubic
Adds a `view=3d` option to the embed SVG endpoint to render an isometric 3D contributions graph with floating Token Usage and Streaks overlays. Existing 2D embeds are unchanged.

- **New Features**
  - `view=3d` renders an isometric 3D graph (2:1 dimetric; heights map to intensity 0–4). Dark/light themes. Pure SVG, no new deps.
  - Overlays: Token Usage (total cost, tokens, active days, date range, rank) and Streaks (longest/current).
[...]

PR #427, opened 2026-04-12
fix(embed): polish 3D isometric rendering — gaps, animation, shading

SeaL773:
## Summary

Follow-up to #416. Improves the visual quality of the `view=3d` isometric embed and adds a 3D toggle to the embed sharing dialog.

### Rendering polish (`renderIsometric3DSvg.ts`)
- **Cube gaps**: Added 1px separation between cubes (cell=10px, cube=9px) so individual cubes are visually distinct
- **SMIL animation**: Cubes drop in and grow from minimum height on load (2s, plays once)
- **Rendering rewrite**: Switched from `<polygon>` with computed vertices to `<rect>` + CSS transforms (`skewY`/`skewX`/`scale`), producing cleaner edges
- **Fixed face shading**: Corrected brightness ratios — top=100%, left=84%, right=70% (d3 `.darker()` formula). Previously left/right were swapped.
- **Removed stroke bleeding**: `stroke="none"` instead of same-color stroke, which caused cubes to look merged
- **CSS classes for face colors**: Reduces SVG payload size vs inline `fill` on every polygon
- **Fixed cube bottom alignment**: Face rect heights are divided by the Y scale factor (1.15) so all cubes align at the same baseline
- **Compact stats box**: Items without subtext (Streaks) use a shorter row height
- **Streak fix**: Current streak now counts from yesterday if no activity today (identified by cubic)

### Embed sharing dialog (`ProfileEmbedDialog.tsx`)
- Added **View** selector (2D / 3D) as the first option group
- When 3D is selected, the embed URL includes `view=3d` and Layout option is hidden (compact not supported for 3D)
- Live preview updates in real-time when switching views

### DIFF
--- a/packages/frontend/__tests__/lib/renderIsometric3DSvg.test.ts
+++ b/packages/frontend/__tests__/lib/renderIsometric3DSvg.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { describe, expect, it } from "vitest";
 import {
   renderIsometric3DEmbedSvg,
   renderIsometric3DErrorSvg,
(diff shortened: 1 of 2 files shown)


SeaL773:
image
Preview

junhoyeo (approved):

junhoyeo:
@SeaL773 thanks a lot!

Issue #441, opened 2026-04-17
need better validation.


ShivamB25:
https://github.com/stelle-w/stelle-w/issues/1

junhoyeo:
Hi @ShivamB25, thanks for opening this.

Unfortunately this issue doesn't contain any description of what validation gap you're seeing in tokscale — the body only links to an unrelated repo/issue (`stelle-w/stelle-w#1`), which doesn't give us anything to act on here.

If you did hit a real validation problem (e.g. bad/dirty data getting accepted on submit, or a trust/decisioning gap you ran into), please reopen with:
- what you ran (command/version)
- the input that should have been rejected
- what actually happened vs. what you expected

For reference, our current validation work lives in #443 and #445 if you want to check whether your case is already covered. Closing for now since there's nothing actionable, but happy to reopen on a concrete repro.

ShivamB25:
@junhoyeo that was just one of the repos doing the uhh.. false numbers on the leaderboard if you were actually seeing your leaderboard you would have recognized  the name

PR #444, opened 2026-04-18
fix(submit): make replayed history deterministic

ShivamB25:
## Summary
- make replayed submission history deterministic instead of additive
- remove stale replayed days and recompute timestamp metadata correctly
- preserve untouched out-of-scope history while fixing in-scope replacements

## Changes
- replace overlapping in-scope replay slices instead of stacking them
- remove omitted in-scope days from replay windows
- recompute or clear stale day-level earliest timestamps after replay changes
- lock the behavior with focused replay regression tests

## Testing
- [x] replay-focused Vitest coverage
- [x] frontend typecheck
- [x] frontend build
- [x] local runtime replay validation against the app and Postgres

Depends on #443
Refs #441

### DIFF
--- a/packages/frontend/__tests__/api/submit.test.ts
+++ b/packages/frontend/__tests__/api/submit.test.ts
@@ -540,6 +540,60 @@ describe('POST /api/submit - Client-Level Merge', () => {
       );
     });
 
+    it.each([
+      ['gpt-4o-2024-08-06', '2024-08-05', '2024-08-06'],
+      ['gpt-4.1-2025-04-14', '2025-04-13', '2025-04-14'],
+    ])(
+      'hard-rejects separator-based dated snapshots for %s before %s',
+      (modelId, submittedDate, availabilityDate) => {
+        const payload = {
+          meta: { generatedAt: new Date().toISOString(), version: '1.0.0', dateRange: { start: submittedDate, end: submittedDate } },
+          summary: { totalTokens: 150, totalCost: 1.5, totalDays: 1, activeDays: 1, averagePerDay: 1.5, maxCostInSingleDay: 1.5, clients: ['claude' as const], models: [modelId] },
+          years: [],
+          contributions: [{
+            date: submittedDate,
+            timestampMs: Date.parse(`${submittedDate}T10:00:00.000Z`),
+            totals: { tokens: 150, cost: 1.5, messages: 1 },
+            intensity: 1 as const,
+            tokenBreakdown: { input: 100, output: 50, cacheRead: 0, cacheWrite: 0, reasoning: 0 },
+            clients: [{ client: 'claude' as const, modelId, tokens: { input: 100, output: 50, cacheRead: 0, cacheWrite: 0, reasoning: 0 }, cost: 1.5, messages: 1 }],
+          }],
+        };
+
+        const result = validateSubmission(payload);
+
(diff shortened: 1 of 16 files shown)


ShivamB25:
This PR is intentionally stacked.

If it were marked as a standalone ready-to-merge change, reviewers would be looking at a diff that conceptually depends on code still under review in the earlier PR(s). Reviewing it in order keeps the trust-boundary changes, replay semantics, and public-surface changes easier to reason about.

junhoyeo:
Thanks for this work, and apologies that it sat unreviewed for so long — that's on us. The replay planner itself is genuinely well-designed: the focused replay suite (6 tests in `__tests__/lib/submissionReplay.test.ts`) still passes in isolation, and `resolveReplayTimestampMs`'s conservative earliest-timestamp resolution is sound.

The problem is that in the months since April, `main` rebuilt the submit subsystem underneath this PR, and the two have diverged in a way that needs your design input, not just a rebase:

1. **Design fork (the big one).** `main` replaced the old merge path with `mergeClientBreakdownsWithRegressionGuard` (`src/lib/db/helpers.ts`), which *preserves* clients that disappear from a same-device resubmit and *blocks* token reductions — the opposite of this PR's deletive, deterministic replay. Merges are now also device-aware (`submittedDevices` table, per-device day fetch, schemaVersion 2 payloads — #517, #609, #616, #627). The replay planner would need to be re-expressed on top of the device-aware model, and we'd need to decide together how deletive replay coexists with the regression guard (e.g., deletive only within an explicitly declared replay window, guard everywhere else?).
2. **Migration collision.** `0005_concerned_justin_hammer` collides with main's `0005_add_case_insensitive_username_index`, and main is now at `0014_widen_cost_columns`. Any revival should drop the hand-numbered migration and regenerate with drizzle-kit so it slots in at 0015+ with a fresh snapshot.
3. **Conflicts.** A merge with main currently conflicts in 9 files; besides the two mechanical test-fixture ones, `submit/route.ts`, `db/helpers.ts`, `db/schema.ts`, `validation/submission.ts`, and `auth/personalTokens.ts` all sit on rewritten code (main also independently added token rehashing in `0006_rehash_plaintext_personal_tokens`, which overlaps the deferred-upgrade commit here).
4. **Stacking.** This depends on #443, which is itself conflicting and needs an author-side rebase first.

One smaller thing worth knowing for any rework: the replay window is `data.meta.dateRange`, so stale in-scope rows on dates *outside* the client-sent range are never stripped — determinism currently depends on the CLI always sending a complete range. Scoping stale-removal per submitted client (or clamping the window server-side) would close that gap.

Given all of the above, I'm going to **close this one** to keep the stack tractable — not because the goal is wrong, but because the branch is structurally superseded: main's regression-guarded, device-aware merge now covers the worst of the resubmit problem (silent inflation/regression), so the remaining live pieces of your trio are the ones in #443 (trust gating, please rebase first) and #445 (adjudication, stacked on top). To be clear and honest: issue #441 is still open, and main does *not* implement deterministic deletive replay — that part of your idea remains valid. If you want to pursue it, the right shape is a fresh PR against current main (ideally after #443 lands), and I'd be glad to discuss the deletive-replay-vs-regression-guard reconciliation before you invest in the rework. If not, no hard feelings; the test suite and planner design here are a solid reference either way.

PR #445, opened 2026-04-18
feat(frontend): isolate public trust and add adjudication

ShivamB25:
## Summary

- stack the adjudication layer on #443 without mixing review operations into submit ingestion
- add dedicated review list, detail, and adjudication APIs for quarantined submissions
- replay accepted review payloads through the same current-main persistence path used by trusted submits
- keep leaderboard, profile, embed, badge, landing, and list API surfaces synchronized after accepted writes

## Rebase onto current `main` and #443

- retained current `main`'s submission model: competitive rows are trusted writes, while quarantined payloads remain in `submission_reviews`
- extracted the current submit persistence transaction so direct trusted submits and accepted reviews share device-aware merges, first-write race recovery, savepoints, timestamp and active-time aggregation, regression guards, and MCP server persistence
- added `0016_nostalgic_senator_kelly.sql` plus generated snapshot/journal metadata for `reviewed_at`, `reviewed_by_username`, `review_note`, and the submission metadata receipt watermark
- kept review cost storage at `numeric(18,4)` and extended migration checks through journal index 16
- removed obsolete public status-filter/profile-metadata changes from the old branch because the current quarantine architecture no longer stores untrusted data in public submissions

## Adjudication contract

- review APIs fail closed unless `SUBMISSION_REVIEW_API_TOKEN` is configured; browser sessions and ordinary personal tokens are not accepted
- `SUBMISSION_REVIEW_OPERATOR_USERNAME` records the external operator identity, with a bounded fallback when unset
- PATCH accepts only `trusted` or `rejected`, validates UUIDs, bounds request bodies, rejects non-object JSON, and records audit metadata
- accepted payloads are revalidated with current server validation before persistence
- row locking makes adjudication single-use; competitive persistence and the review-state transition commit atomically
- rejection records the decision without creating competitive data
- a partial submission increments `submit_count` when its trusted partition is first written; later approval of its queued partition does not increment it again
- a fully queued submission increments `submit_count` once when approval creates or updates competitive data

## Metadata ordering

- `submissions.metadata_received_at` is a dedicated receipt-order watermark; migration `0016` backfills it from each row's existing `updated_at` before setting the default and `NOT NULL`
- delayed approvals compare the review's latest submission receipt time (`submission_reviews.updated_at` before adjudication) against that watermark
- aggregate usage and device/day data still merge from an older approved review, but stale `cliVersion`, submission hash, session-shape metrics, schema metadata, and MCP server metadata cannot overwrite a newer direct submission
- pending-review retries preserve `competitive_write_applied = true` monotonically with SQL `OR`

## Cache and rank consistency

- compare pre/post all-time `RANK()` snapshots using the same aggregate sources and `numeric(18,4)` cost semantics as current public rank queries
- invalidate every affected user's rank-bearing caches and public paths after acceptance
- normalize and deduplicate mixed-case usernames before cache-tag invalidation
- include profile, API, embed, and badge paths for canonical and lowercase username variants
- preserve async group-leaderboard invalidation for the adjudicated user

## Verification

- [x] Frontend Vitest after latest-main restack: 74 files, 670 tests passed
- [x] Frontend typecheck: `bun run typecheck`
- [x] Rust CLI after `main.rs` conflict resolution: 1004 tests passed, 1 ignored
- [x] Frontend production build: `bun run build`
- [x] Frontend lint: 0 errors (6 pre-existing warnings outside this change)
- [x] Fresh PostgreSQL migration/runtime checker: contiguous journal `0..16`, every migration applied, receipt watermark and competitive-write marker defaults/nullability verified, plus constraints/indexes/concurrency/rollback checks passed
- [x] Live app + PostgreSQL: listed queued reviews, loaded detail, accepted a review into submissions/daily/device/MCP data, rejected a review without competitive data, confirmed repeat adjudication returns 409, and accepted a mixed-case user after rank snapshot/cache changes
- [x] Submit-count runtime: first trusted write persisted `1`; partial review approval kept `1` while adding its queued day
- [x] Metadata-order runtime: queued version `1.0.0`, accepted newer direct version `9.0.0`, then approved the older review; aggregate totals reached 200 while `cli_version = 9.0.0`, MCP metadata stayed `newer-direct-mcp`, and `submit_count = 2`

## Stack

Depends on #443 commit `095a6975` and is intentionally kept as a separate stack layer. #444 remains closed and untouched.

Refs #441

### DIFF
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -4209,6 +4209,8 @@ struct TsDailyContribution {
     token_breakdown: TsTokenBreakdown,
     clients: Vec<TsSourceContribution>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    timestamp_ms: Option<i64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     active_time_ms: Option<i64>,
 }
 
@@ -4360,6 +4362,7 @@ fn to_ts_token_contribution_data(
                         messages: s.messages,
                     })
                     .collect(),
+                timestamp_ms: d.timestamp_ms,
                 active_time_ms: d.active_time_ms,
             })
             .collect(),
@@ -4944,6 +4947,22 @@ struct SubmitResponse {
     details: Option<Vec<String>>,
 }
 
+fn terminal_safe(value: &str) -> String {
+    value
+        .chars()
+        .filter(|character| !character.is_control())
+        .collect()
+}
+
+fn safe_profile_username(value: &str) -> Option<&str> {
+    (!value.is_empty()
+        && value.len() <= 39
+        && value
+            .bytes()
+            .all(|byte| byte.is_ascii_alphanumeric() || byte == b'-'))
+    .then_some(value)
+}
+
 fn submit_confirmation_message(
     trust_state: Option<&str>,
     competitive_write_applied: bool,
@@ -5383,14 +5402,14 @@ fn run_submit_command(
                     });
 
             if !status.is_success() {
-                let error = body
-                    .error
-                    .clone()
-                    .unwrap_or_else(|| "Submission failed".to_string());
+                let error = terminal_safe(&body.error.as_deref().unwrap_or("Submission failed"));
                 eprintln!("\n  {}", format!("Error: {}", error).red());
                 if let Some(details) = body.details {
                     for detail in details {
-                        eprintln!("{}", format!("    - {}", detail).bright_black());
+                        eprintln!(
+                            "{}",
+                            format!("    - {}", terminal_safe(&detail)).bright_black()
+                        );
                     }
                 }
                 println!();
@@ -5412,7 +5431,10 @@ fn run_submit_command(
             }
             println!();
             if let Some(id) = body.submission_id {
-                println!("{}", format!("  Submission ID: {}", id).bright_black());
+                println!(
+                    "{}",
+                    format!("  Submission ID: {}", terminal_safe(&id)).bright_black()
+                );
             }
             if let Some(metrics) = &body.metrics {
                 print_submit_metrics("Competitive summary", metrics);
@@ -5422,8 +5444,14 @@ fn run_submit_command(
             }
             if let Some(username) = body
                 .username
-                .clone()
-                .or_else(|| auth_token.username.clone())
+                .as_deref()
+                .and_then(safe_profile_username)
+                .or_else(|| {
+                    auth_token
+                        .username
+                        .as_deref()
+                        .and_then(safe_profile_username)
+                })
             {
                 println!();
                 println!(
@@ -5441,7 +5469,10 @@ fn run_submit_command(
                 if !warnings.is_empty() {
                     println!("{}", "  Warnings:".yellow());
                     for warning in warnings {
-                        println!("{}", format!("    - {}", warning).bright_black());
+                        println!(
+                            "{}",
+                            format!("    - {}", terminal_safe(&warning)).bright_black()
+                        );
                     }
                     println!();
                 }
@@ -6039,6 +6070,25 @@ mod tests {
     };
 
     #[test]
+    fn terminal_safe_removes_control_characters_from_server_messages() {
+        assert_eq!(
+            terminal_safe("warning\u{1b}[31m\r\nnext\u{85}line"),
+            "warning[31mnextline"
+        );
+    }
+
+    #[test]
+    fn profile_links_reject_control_and_path_injection() {
+        assert_eq!(
+            safe_profile_username("valid-user-25"),
+            Some("valid-user-25")
+        );
+        assert_eq!(safe_profile_username("bad\u{1b}]8;;https://evil"), None);
+        assert_eq!(safe_profile_username("../another-user"), None);
+        assert_eq!(safe_profile_username("percent%2Fpath"), None);
+    }
+
+    #[test]
     fn review_required_response_never_uses_success_confirmation() {
         let response: SubmitResponse = serde_json::from_value(serde_json::json!({
             "success": true,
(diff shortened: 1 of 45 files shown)


ShivamB25:
This PR is intentionally stacked.

If it were marked as a standalone ready-to-merge change, reviewers would be looking at a diff that conceptually depends on code still under review in the earlier PR(s). Reviewing it in order keeps the trust-boundary changes, replay semantics, and public-surface changes easier to reason about.

junhoyeo:
Hi @ShivamB25 — first, thank you for this substantial piece of work, and apologies that it sat unreviewed for so long. The public-trust isolation + adjudication design here is genuinely valuable and is not something main has implemented in the meantime, so we'd love to land it — but the branch can no longer be merged as-is.

**Where things stand**: the merge base (48aa05c7, Apr 17) is now ~350 commits behind main, and a test merge produces 20 conflicted files. The conflicts sit exactly on the subsystems this PR rewrites, and main has since rewritten them too:

- **Leaderboard** (`src/lib/leaderboard/getLeaderboard.ts`): main moved to DB-side ranking/pagination with `RANK() OVER (...)` window functions (#611, #655) to fix pagination drift and competition-rank ties. This branch's all-time path ranks the full verified-user set in memory (`fetchAllTimeLeaderboardRows` → `rankLeaderboardUsers` → `slice`), which both conflicts and regresses those fixes. Suggested approach on rebase: fold the trust filter (`status = VERIFIED` / review-status predicate) into main's existing window-function queries rather than replacing them — `fetchCompetitiveRankedUsernames` in your own `submissionReviews.ts` already shows the right SQL pattern.
- **Submit path** (`src/app/api/submit/route.ts`, `src/lib/submissionPersistence.ts`): main hardened this independently (#616, #691, #557). One carry-over to fix during the rework: `applyTrustedSubmission` does SELECT ... FOR UPDATE then INSERT — FOR UPDATE locks nothing when no row exists, so two concurrent first submits for a new user race to a unique-constraint violation. Please use `INSERT ... ON CONFLICT (user_id) DO UPDATE ... RETURNING` (matching main's current path).
- **Schema & migrations**: main's migration chain is now at `0014_widen_cost_columns.sql`, and your `0005`/`0006` numbers are already taken by unrelated migrations on main. The new `submission_reviews` migration will need to be regenerated as 0015+ against the current `_journal.json`.
- **Profile/embed** (`ProfilePageClient.tsx`, `getUserEmbedStats.ts`, `api/users/[username]/route.ts`): main reworked these for range filters and per-device stats (#785, #702, #675).

A few smaller items from the earlier automated review that are still valid and worth folding in:
- `listSubmissionReviewArtifacts` is typed as returning summaries but maps full Detail objects, so the admin list endpoint ships every submission's full payload/hash — a summary-scoped mapper would fix both the type lie and the response bloat.
- `getUtcDateStringFromTimestamp` can throw a RangeError for schema-valid `timestampMs` above 8.64e15 (max JS Date) — tighten the schema bound or guard `Number.isFinite(date.getTime())`.
- `toLocaleString()` for review timestamps risks SSR/CSR hydration mismatches — prefer a fixed UTC formatter.

Also worth confirming before you invest: the adjudication routes gate on `session.isAdmin`, which doesn't exist on current main — that's fine as a net-new concept, but it's worth a quick alignment on whether admin-only is the intended auth model.

**How this fits with the rest of your stack**: this PR stacks on #443 and #444. I've left rebase notes on #443 (which we want landed close together with this one, so REVIEW_REQUIRED submissions have an adjudication path and nothing gets stranded), and I've closed #444 since main's regression-guarded merge superseded that branch structurally — see the note there for what remains salvageable. So the practical path is: rebase #443 first, then rework this one on top — or, given how much main has moved, possibly collapse the remaining work into a single fresh PR against current main. Happy to review promptly this time either way, and thanks again for the effort you put into this.

ShivamB25:
@junhoyeo #445 is now rebased as a separate layer on the final #443 head. I adapted adjudication to current `main`: accepted reviews use the shared current submit persistence transaction, audit fields are migration `0016`, review APIs use a dedicated fail-closed operator token, stored payloads are revalidated, and row locking keeps replay plus state transition atomic. Rank/cache refresh now mirrors current aggregate rank semantics, handles ties and mixed-case cache keys, and covers profile/API/embed/badge/leaderboard surfaces. Final verification: 71 Vitest files / 646 tests, typecheck, production build, migration checker through `0..16`, and live PostgreSQL accept/reject/idempotency/mixed-case-rank scenarios all passed. The PR description has the full deployment and verification contract. Ready for re-review after #443.

ShivamB25:
@junhoyeo #445 is restacked on final #443 head `3a4d529f`; current head is `674524ad`.

The follow-up fixes two delayed-adjudication invariants:
- partial submissions increment `submit_count` only on their initial trusted write; approval of the queued partition leaves the count unchanged, while fully queued submissions count once on approval;
- older queued payloads cannot overwrite newer direct-submit metadata. `0016` adds `submissions.metadata_received_at`, backfilled from existing `updated_at`, and approvals compare the review receipt watermark before updating CLI version, hash, schema/session-shape fields, or MCP metadata. Aggregate usage still merges.

Regression coverage now includes sticky review upserts, first-write count = 1, stale metadata omission, receipt timestamp forwarding, and delayed A/B ordering. Verification: **73 files / 651 tests passed**, typecheck and production build passed, lint has 0 errors (6 pre-existing warnings), fresh PostgreSQL migrations/checker passed through `0..16`, and live runtime checks confirmed partial approval stays at count 1 plus an older queued v1 approval preserves newer direct v9/MCP metadata while totals merge.

ShivamB25:
@junhoyeo Restacked #445’s five adjudication-only commits onto repaired #443 head `095a6975`; new #445 head is `61c561d6`. The resolved `main.rs` is inherited from #443, so the CLI conflict was resolved once rather than independently in both PRs.

GitHub now reports `MERGEABLE`. Final stacked verification: Rust CLI **1004 passed, 1 ignored**; frontend **74 files / 670 tests passed**; typecheck passed; production build passed; lint has **0 errors** and the same 6 pre-existing warnings.

ShivamB25:
fk this.

Issue #446, opened 2026-04-18
OpenCode SQLite parser overcounts tokens N× when sessions are forked (missing deduplication, regression from #80/#181)


ashiner:
Note that this report was generated using Cluade Opus 4.6. 

**Description**
When using OpenCode with forked sessions, tokscale inflates token counts by N× where N is the number of forks. This causes wildly incorrect cost estimates — in my case a single day jumped from ~$126 to ~$1,032 (8× inflation from 9 fork sessions).

This is the same class of bug fixed in #80 for Claude Code JSONL, but the SQLite parser added in #181 does not have equivalent deduplication.

**Root cause**

When OpenCode forks a session, it copies message rows into new session records in the SQLite DB. The copied rows have different message.id values (new UUIDs) but identical time_created timestamps and identical token counts as the originals.

No new API call is made when forking — the history is copied at the DB level only. A real API call only happens when you send a new message in the forked session, which gets a new unique time_created.

Because the SQLite parser has no deduplication, it sums all rows across all sessions, counting the copied history once per fork.

You can verify this by joining two sessions on time_created — every row matches with identical token counts. A DISTINCT on time_created before aggregating collapses the duplicates and produces the correct numbers, which exactly match tokscale's own earlier exports before the source-message-cache.bin was rebuilt.

**Reproduction** steps

1. Have an OpenCode session with significant history
2. Fork the session one or more times
3. Optionally leave the forks unused (never send a new message in them)
4. Run tokscale or force a cache refresh — token counts for that day are multiplied by the number of forks + 1

**Impact**
In my case: 9 fork sessions, reported cost ~$1,032, actual cost ~$126, ratio 8.2×. The inflation is proportional to the number of forks. Unused forks where no new messages were sent after the fork point contribute 100% phantom tokens.

**Expected** behavior

Messages with identical time_created timestamps across sessions should be counted only once, consistent with how #80 deduplicates by messageId:requestId for Claude Code JSONL.

**Environment**

tokscale 2.0.0, OpenCode (SQLite-based session storage, post-#181), macOS

**Suggested** fix

In the OpenCode SQLite parser, deduplicate assistant messages by time_created before aggregating — equivalent to the messageId:requestId hash set used in #80. Since time_created is recorded at API response time, each unique value represents exactly one real API call.

junhoyeo:
Fix has been made in #451 and deployed in https://github.com/junhoyeo/tokscale/releases/tag/v2.0.27! 🎉
[...]

PR #451, opened 2026-04-21
fix(opencode): deduplicate forked sqlite history

IvGolovach:
## Summary
- deduplicate fork-copied OpenCode SQLite assistant history so forked sessions do not inflate usage, cost, or message counts
- keep the existing JSON ↔ SQLite overlap dedup path intact by preserving cross-source message identity keys
- invalidate stale local source-message cache entries so the new OpenCode dedup semantics apply consistently

## Why
OpenCode forks can copy existing assistant history into new SQLite rows with new database IDs. The current parser treats each SQLite row as a distinct message, so forked history is counted again for each fork. That inflates usage, cost, and message totals for OpenCode users. This PR fixes that parser-level correctness bug without breaking the existing legacy JSON overlap handling.

## Diff scope
- add a parser-local OpenCode SQLite fingerprint that collapses fork-copied assistant history inside `parse_opencode_sqlite`
- keep `UnifiedMessage.dedup_key` for JSON ↔ SQLite overlap dedup, preferring embedded OpenCode message IDs when they exist and falling back to the SQLite row ID otherwise
- add parser-level regression tests for forked SQLite copies, same-timestamp distinct messages, and cross-source JSON overlap
- add core-level regression tests for the aggregate parse path and the local client-count path
- bump the source-message cache schema version so cached OpenCode SQLite parses do not keep the old overcounting behavior
- clean up the cache file-locking calls to keep `clippy -D warnings` green

## Test proof
- `cargo test -p tokscale-core opencode -- --nocapture`
  - `49 passed, 0 failed, 1 ignored`
  - `472 passed, 0 failed, 1 ignored`
  - passed
- `cargo fmt --all --check`
  - passed
- `git diff --check`
  - passed

## Verification-pack proof

## Migration notes
Not applicable - there is no database or user-data migration. This PR only bumps the local source-message cache schema version to invalidate stale cached OpenCode SQLite parse results.

## CI context confirmation
CI context names unchanged.

## Rollback plan
- no DB downgrade required

## Known residual risks
- the new OpenCode SQLite dedup fingerprint is intentionally strict so same-timestamp messages with different payloads still survive, but if OpenCode changes its fork-copy shape substantially in the future, the fingerprint may need to be updated again

## Closes
- #446


junhoyeo (approved):
Nice catch Ivan! Thank you as always 👍

junhoyeo:
deployed in https://github.com/junhoyeo/tokscale/releases/tag/v2.0.27! 🎉

Issue #439, opened 2026-04-15
Incorrect token usage for Gemini-CLI and possibly other AI agents???


whab:
Great tool, but I believe the token usage and cost is incorrectly reported with Gemini-CLI and possibly other AI agents: based on below outputs from Gemini-CLI and tokscale, it looks like Cache Reads are counted twice into Total (and Cost). Am I wrong?

> /stats model                                                                                                                                                                                           
╭─────────────────────────────────────────────────────────╮
│  Model Stats For Nerds                                                                                                                                                                               │
│  Metric                      gemini-3-flash-preview                                                                                                                                            │
│  ─────────────────────────────────────────────────────────  │
│  API                                                                                                                                                                                                               │
│  Requests                    8                                                                                                                                                                              │
│  Errors                          3 (37.5%)                                                                                                                                                                │
│  Avg Latency               2.6s                                                                                                                                                                         │
│  Tokens                                                                                                                                                                                                        │
│  Total                           111,513                                                                                                                                                                     │
│    ↳ Input                     52,955                                                                                                                                                                    │
│    ↳ Cache Reads       57,874 (52.2%)                                                                                                                                                       │
│    ↳ Thoughts                  383                                                                                                                                                                     │
│    ↳ Output                       301                                                                                                                                                                     │
│  Roles                                                                                                                                                                                                           │
│  main                                                                                                                                                                                                            │
│    ↳ Requests                8                                                                                                                                                                            │
│    ↳ Input                      52,955                                                                                                                                                                    │
│    ↳ Output                        301                                                                                                                                                                     │
│    ↳ Cache Reads         57,874                                                                                                                                                                    │
╰─────────────────────────────────────────────────────────╯

% npx tokscale@latest --light --today
  Token Usage Report by Model (Today)
┌───────────────┬─────────────────┬─────────────────────────────┐
│ Client    │ Provider   │ Model                               │ Resolved                          │ Input       │ Output │ Cache Write │ Cache Read│ Total         │ Cost         │
│ Gemini  │ google     │ gemini-3-flash-preview │ gemini-3-flash-preview  │   110,829 │      301 │                  0 │           57,874 │   169,004 │       $0.06 │
├──────────────┼──────────────────┼───────────────┼─────────────┤
│ Total      │                 │                                           │                                           │   110,829 │      301 │                  0 │           57,874 │   169,004 │       $0.06 │
└───────────────┴─────────────────┴─────────────────────────────┘

[...]

PR #452, opened 2026-04-21
fix(gemini): avoid double-counting cached input tokens

IvGolovach:
## Summary
- normalize Gemini cached-input accounting so cached tokens are not counted twice in usage or cost
- fix the Gemini pricing path so cache-read tokens use their own pricing bucket instead of being silently folded into paid input
- invalidate stale local source-message cache entries so existing Gemini files pick up the new parser semantics

## Why
Gemini usage metadata reports prompt/input tokens and cached tokens as separate dimensions, but Tokscale currently stores the raw prompt count as `input` and also stores the cached count as `cache_read`. That inflates Gemini token totals and can also inflate costs by treating cached tokens as full-price input. This PR fixes that correctness bug while preserving legacy session files that already store net input counts.

## Diff scope
- add safe Gemini token normalization in `crates/tokscale-core/src/sessions/gemini.rs`
- always normalize headless/API-style Gemini stats where prompt counts are cache-inclusive
- normalize legacy session JSON only when the reported `total` explicitly proves an inclusive-input shape, so already-net session files are not undercounted
- remove the Gemini-specific pricing special-case that discarded `cache_read` during cost calculation
- bump the local source-message cache schema version to invalidate stale Gemini parse results
- add regression tests for inclusive-cache stats, already-net legacy sessions, clamp edge cases, and Gemini cache-read pricing

## Test proof
- `cargo test -p tokscale-core gemini -- --nocapture`
  - `19 passed, 0 failed`
  - `tokscale-core: 524 passed, 0 failed, 1 ignored`
  - `hermes integration tests: 3 passed, 0 failed`
  - passed
- `cargo fmt --all --check`
  - passed
- `git diff --check`
  - passed

## Verification-pack proof

## Migration notes
Not applicable - there is no database or user-data migration. This PR only bumps the local source-message cache schema version to invalidate stale Gemini parse results.

## CI context confirmation
CI context names unchanged.

## Rollback plan
- no DB downgrade required

## Known residual risks
- historical submissions that were already uploaded with inflated Gemini totals do not self-heal server-side; users would need a fresh local parse and resubmit if they want corrected historical submission data
- legacy session JSON without a trustworthy `total` field is left unchanged on purpose to avoid undercounting already-net input records

## Closes
- #439

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -1474,22 +1474,11 @@ fn apply_pricing_if_available(
         return;
     };
 
-    let calculated_cost = if message.client.eq_ignore_ascii_case("gemini") {
-        let usage = TokenBreakdown {
-            input: message.tokens.input,
-            output: message.tokens.output + message.tokens.reasoning,
-            cache_read: 0,
-            cache_write: 0,
-            reasoning: 0,
-        };
-        pricing.calculate_cost_with_provider(&message.model_id, Some(&message.provider_id), &usage)
-    } else {
-        pricing.calculate_cost_with_provider(
-            &message.model_id,
-            Some(&message.provider_id),
-            &message.tokens,
-        )
-    };
+    let calculated_cost = pricing.calculate_cost_with_provider(
+        &message.model_id,
+        Some(&message.provider_id),
+        &message.tokens,
+    );
 
     if calculated_cost > 0.0 {
         message.cost = calculated_cost;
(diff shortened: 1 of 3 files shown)


junhoyeo:
Using `google-gemini/gemini-cli` source as the reference, PR #452’s direction looks correct.

A quick note on the external review path: I could not complete a live `gemini -p` review because the local Gemini CLI is currently blocked by a billing-disabled project configuration, so I validated this against Gemini CLI’s public source code instead.

## What Gemini CLI itself does
- `packages/core/src/services/chatRecordingService.ts` writes:
  - `input = promptTokenCount`
  - `cached = cachedContentTokenCount`
  - `total = totalTokenCount`
- `packages/core/src/agent/event-translator.ts` maps:
  - `inputTokens = promptTokenCount`
  - `cachedTokens = cachedContentTokenCount`
- `packages/core/src/output/stream-json-formatter.ts` exposes both:
  - raw `input_tokens` from `tokens.prompt`
  - separate `cached`
  - separate derived `input`

That is the key evidence: Gemini CLI keeps prompt/input and cached as separate dimensions while also preserving a total. So if Tokscale stores:
- `input = promptTokenCount`
- `cache_read = cachedContentTokenCount`

and then totals them both, it can double-count unless it normalizes.

## How that maps to Tokscale
- Issue #439 reports exactly that symptom: cache reads appear counted twice in total/cost.
- PR #452 updates `crates/tokscale-core/src/sessions/gemini.rs` to normalize headless/session input vs cache overlap.
- PR #452 also removes the Gemini pricing special-case in `crates/tokscale-core/src/lib.rs`, which looks right because cached reads should use their own pricing bucket rather than being discarded or folded into paid input.

## Review verdict
- Low risk / supports merge for the main fix.
- The external Gemini CLI code supports the assumption that raw prompt/input is not the same thing as net non-cached input.

## Residual risk
- The session-file heuristic in `normalize_gemini_session_input_and_cache(...)` depends on `total` matching one shape versus another. That seems reasonable and safer than always subtracting, but if Gemini CLI changes session serialization again, Tokscale may need another format-specific adjustment.
- Headless normalization looks stronger, because Gemini CLI’s streamed/output stats already distinguish raw prompt vs effective input more explicitly.

So: using `google-gemini/gemini-cli` source as the reference, PR #452 looks justified and should merge.

junhoyeo:
Deployed in https://github.com/junhoyeo/tokscale/releases/tag/v2.0.27! 🎉

Issue #410, opened 2026-04-08
Amp CLI usage data significantly mismatches Amp's official Usage page


chanity256:
# [Bug] Amp CLI usage data significantly mismatches Amp's official Usage page

## Environment

- **Device**: Mac mini (M2, 2023)
- **OS**: macOS 15.5 (Sequoia, Build 24F74)
- **tokscale version**: v2.0.0 (`npx tokscale@latest --amp`)
- **Amp CLI**: latest version, data source at `~/.local/share/amp/threads/`

## Description

When using `npx tokscale --amp` to track Amp (AmpCode) usage, the daily token counts and cost figures **significantly differ** from those shown on Amp's official Usage page (`Settings > Usage`). The discrepancies are not minor rounding differences — the numbers are fundamentally inconsistent, suggesting that tokscale is either missing some Amp thread data, or the cost calculation methodology diverges from Amp's own accounting.

As shown in **amp.png**, token consumption was recorded for AMP on **April 3rd, 6th, 7th, and 8th**. However, after running ⁠npx tokscale --amp, the results displayed in **tokscale.png** indicate that none of these token usages were included in the statistics.

<img width="1123" height="970" alt="Image" src="https://github.com/user-attachments/assets/4048cff2-1907-42b9-b5fc-f8d87df3794f" />

<img width="1280" height="1322" alt="Image" src="https://github.com/user-attachments/assets/377e660c-fbe0-423d-8807-3409a600517d" />

junhoyeo:
Fix has been made in #453 and deployed in https://github.com/junhoyeo/tokscale/releases/tag/v2.0.27! 🎉
Please try again with latest version of Tokscale, thanks!

PR #453, opened 2026-04-21
fix(amp): reconcile partial ledger usage with message fallback

IvGolovach:
## Summary
- reconcile Amp `usageLedger.events` with assistant `thread.messages` so partial ledger data does not silently drop real usage entries
- avoid double-counting when the same Amp usage record appears in both sources
- invalidate stale local source-message cache entries so existing Amp files pick up the new parser behavior

## Why
Tokscale currently returns early as soon as it finds any Amp `usageLedger.events`. That means a partial ledger can suppress real assistant usage that still exists in `thread.messages`, which makes Amp days disappear from local usage reports. This PR fixes that parser-level correctness bug without introducing double-counting when both sources describe the same usage record.

## Diff scope
- replace the current Amp ledger-first early-return path with record-level reconciliation between `usageLedger.events` and assistant `thread.messages` in `crates/tokscale-core/src/sessions/amp.rs`
- match duplicate Amp usage records by model plus token breakdown so the same usage is only counted once while unmatched message-level usage still survives
- preserve better timestamps for matched records by using message-derived timestamps when the ledger entry does not have an explicit timestamp
- fall back to file mtime when Amp message usage exists but the thread does not carry a usable `created` timestamp
- add parser-level regression tests for partial ledger recovery, full-ledger deduplication, missing-ledger-timestamp handling, and missing-thread-created fallback
- add an end-to-end local parse regression test for the user-visible Amp day-recovery path
- bump the local source-message cache schema version so cached Amp parses do not keep the old broken behavior

## Test proof
- branch integrity: `0 behind / 1 ahead` vs `origin/main`
- merge-base: `04207c3ba7f7fa962822ed7f71bef53914ff1c90`
- commits introduced in this PR: `ec3a1e04e0171a352ef41ae88f41f19e510891d2`
- `cargo test -p tokscale-core amp -- --nocapture`
  - `21 passed, 0 failed`
  - `472 passed, 0 failed, 1 ignored`
  - passed
- `cargo fmt --all --check`
  - passed
- `git diff --check`
  - passed

## Verification-pack proof

## Migration notes
Not applicable - there is no database or user-data migration. This PR only bumps the local source-message cache schema version to invalidate stale Amp parse results.

## CI context confirmation
CI context names unchanged.

## Rollback plan
- no DB downgrade required

## Known residual risks
- this PR fixes the missing-entry parser bug for partial Amp ledger data, but if Amp changes how it duplicates or shapes usage between `usageLedger.events` and assistant messages again in the future, the matching heuristic may need another update
- historical reports using old local cache entries are handled by the cache-schema bump, but already-uploaded remote submissions do not self-heal server-side without a fresh local parse and resubmit

## Closes
- #410

---

<a href="https://app.devin.ai/review/junhoyeo/tokscale/pull/453" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


junhoyeo (approved):
Thanks!

junhoyeo:
Deployed in https://github.com/junhoyeo/tokscale/releases/tag/v2.0.27! 🎉

PR #458, opened 2026-04-22
fix(tui): preserve per-tab sort state across tab switches

stevejkang:
Before this change, apply_tab_sort_defaults() unconditionally reset sort to Cost/Descending on every keyboard tab switch (#422). This destroyed any user-configured sort order (e.g. Tokens/Ascending) whenever the user navigated between tabs. Mouse tab clicks did not call apply_tab_sort_defaults() at all, causing keyboard and mouse to behave differently.

The original function (introduced in 0c21da3) only forced Date sort when entering the Hourly tab and left other tabs untouched, but this let the Hourly Date sort leak into other tabs. #422 added an else branch to fix the leak, which introduced the regression.

Replace apply_tab_sort_defaults() with switch_tab() that saves the current tab's sort state before leaving and restores the target tab's saved state on entry. First visit to a tab falls back to its default (Hourly: Date/Desc, others: Cost/Desc). This resolves both the original Date-sort leak and the #422 regression, and unifies keyboard and mouse tab-switch behavior.

---
## Summary by cubic
Preserves per-tab sort state in the TUI across keyboard and mouse tab switches. Fixes resets and leaks so each tab keeps its sort field and direction.

- **Bug Fixes**
  - Replaced apply_tab_sort_defaults with switch_tab, which saves the current tab’s sort and restores the target tab’s saved state.
  - First visit defaults: Hourly = Date/Descending; others = Cost/Descending.
  - Unified handling for Tab/BackTab/Arrow keys and mouse tab clicks through switch_tab to keep behavior consistent.

<sup>Written for commit b8b1fd2156ee42c918bf1396a14c7518beb1a9e4. Summary will update on new commits.</sup>

### DIFF
--- a/crates/tokscale-cli/src/tui/app.rs
+++ b/crates/tokscale-cli/src/tui/app.rs
@@ -28,7 +28,7 @@ pub struct TuiConfig {
     pub initial_tab: Option<Tab>,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Tab {
     Overview,
     Models,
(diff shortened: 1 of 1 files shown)


stevejkang:
hey cubic, are you sleeping?

junhoyeo:
> hey cubic, are you sleeping?

I must have used all of its quota 😭

PR #459, opened 2026-04-22
fix(tui): use fuzzy matching for provider colors

zxmpg41:
## Summary
Custom/namespaced provider IDs (e.g. `api-anthropic-test`, `my-org-google`) now correctly resolve to their brand color palettes instead of falling back to gray

## Problem
`get_provider_shade` used exact string matching (`match provider { "anthropic" => ... }`) to select color palettes. When OpenCode (or other clients) use custom provider IDs that contain the canonical name as a substring (e.g. `api-anthropic-test`, `my-org-google`), none of the known palettes matched and all models rendered in the neutral gray `UNKNOWN_SHADES` ramp.

## Solution
Replace exact `match` arms with guard clauses using `contains()` on the lowercased provider string. This preserves the compact `match` structure while allowing any provider ID containing a known name (like `"anthropic"`, `"google"`, `"gemini"`) to pick up the correct palette. Config overrides via `[colors.providers]` still take priority.

---
## Summary by cubic
Custom or namespaced provider IDs now map to the correct brand colors in the TUI using fuzzy matching, preventing fallback to gray. This makes provider colors consistent even when IDs include prefixes/suffixes.

- **Bug Fixes**
  - Use lowercase contains matching in `get_provider_shade` to handle namespaced IDs.
  - Support common aliases (e.g., "gemini"→Google, "grok"→XAI, "llama"→Meta).
  - Preserve `[colors.providers]` overrides; added tests for fuzzy matching.

<sup>Written for commit f83c07ab93be14bbe991befddfbd745c5a34e4ed. Summary will update on new commits.</sup>

### DIFF
--- a/crates/tokscale-cli/src/tui/ui/widgets.rs
+++ b/crates/tokscale-cli/src/tui/ui/widgets.rs
@@ -70,16 +70,19 @@ pub fn get_provider_shade(provider: &str, rank: usize) -> Color {
     if let Some(base) = TokscaleConfig::load().get_provider_color(provider) {
         return shade_from_base(base, rank);
     }
-    let palette: &[(u8, u8, u8)] = match provider {
-        "anthropic" => &ANTHROPIC_SHADES,
-        "openai" => &OPENAI_SHADES,
-        "google" => &GOOGLE_SHADES,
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
thanks!

PR #465, opened 2026-04-25
feat(cli)!: remove deprecated per-client boolean flags

junhoyeo:
**Breaking change** — targets **v3.2.0**. Removes the deprecated, hidden per-client boolean flags (`--opencode`, `--claude`, `--codex`, …); use the canonical `--client/-c` instead. These flags have been hidden from `--help` and emitting a deprecation warning since #464 (pre-v3.0.0), so the migration window is well past.

> ⚠️ After 3.2.0, `tokscale --opencode` (etc.) hard-errors with `unexpected argument`. Migrate scripts to `--client`.

## Re-cut on current main (2026-06-18)
The original branch was ~2 months stale (19 flags then; main now has **32**, and `build_client_filter` is `home_dir`-aware). This branch was re-cut fresh on current `main`.

## What changed
- Remove all **32** hidden legacy bool fields from `ClientFlags`.
- Simplify `build_client_filter_with_defaults` to the single canonical `--client` path + `defaultClients` fallback (drops the legacy mapping array and `emit_legacy_client_flag_warning`).
- Add `id = "client_filter"` / `value_name = "CLIENTS"` to the `--client` arg (avoids a clap id collision now that the bool fields are gone; keeps the `<CLIENTS>` help placeholder).
- Tests: a table-driven guard asserting **all 32 removed flags now produce a clap parse error**, plus positive `--client opencode | opencode,claude | synthetic` tests. ~61 legacy-flag sites in `cli_tests.rs` migrated to `--client`.
- Remove the README deprecation-notice paragraph.

## Migration
| Before | After |
|---|---|
| `tokscale --opencode` | `tokscale --client opencode` |
| `tokscale --opencode --claude` | `tokscale --client opencode,claude` |
| `tokscale --synthetic` | `tokscale --client synthetic` |

Verified locally: `cargo build`/`clippy`/`fmt` clean; 669 bin + 122 integration tests pass; `--opencode` rejected, `--client opencode` works.

🤖 Re-cut with [Claude Code](https://claude.com/claude-code)


junhoyeo:
## Update: Oracle review — addressed LOW finding + rebased on updated #464

### LOW: legacy-rejection guard was too narrow

The original \`test_client_flags_legacy_long_flags_rejected\` only checked \`--claude\`. Expanded to a table over **all 19 removed flags** with \`--synthetic\` called out explicitly — it's the only legacy flag without a matching \`ClientId\`, so its handling has always been special-cased and is the most likely to regress if anyone resurrects a boolean field. Also added a positive sanity sweep to verify \`--client opencode\`, \`--client claude\`, and \`--client synthetic\` all still parse, so the negative test cannot silently pass via an over-tight parser.

Pushed as \`f1caf3d test(cli): expand legacy-flag rejection guard to all 19 removed flags\`.

### Rebased on updated #464

Branch was rebased after \`308b471\` landed on \`feat/cli-client-filter-flag\`. The rebase pulls in:
- The HIGH-severity \`ClientFilter::default_set()\` fix (no-filter TUI / warm-cache mismatch)
- The MEDIUM-severity README example migration (stale \`--claude\` / \`--opencode\` examples in non-deprecation sections)

So once #464 lands, this branch's diff collapses cleanly to just the removal commits with no carryover noise.

### Numbers (this branch only)

- Tests: **364 unit + 83 integration = 447 pass / 0 fail** (4 fewer unit tests vs. #464 because legacy-only tests were removed; gained 2 from #464's regression suite + 1 expanded guard here)
- Diff vs #464: 6 files, +163 / -287 (net **-124 lines**, slightly improved from before due to the expanded test)

PR remains **DRAFT — DO NOT MERGE until v3.0.0**.

PR #496, opened 2026-05-03
fix(codex): attribute resumed token counts to later model

IvGolovach:
## Summary
Fix Codex resumed-session attribution so early `token_count` events emitted before `turn_context` inherit the later model instead of staying under `unknown`.

Fixes #490.

## Why
Codex 0.124 resumed sessions can replay `event_msg` / `token_count` rows before the first `turn_context` row that carries `payload.model`. The previous parser emitted those early rows immediately with `modelId = unknown`, which also meant pricing stayed at `0` even after the real model appeared later in the same session file.

## Diff scope
- `crates/tokscale-core/src/sessions/codex.rs`: buffer model-less Codex `token_count` messages until the parser observes a model, then flush them with that model; preserve `unknown` only when no model ever appears.
- `crates/tokscale-core/src/lib.rs`: avoid writing Codex source-cache entries when a file still has unresolved model-less token events.
- `crates/tokscale-core/src/message_cache.rs`: bump the source-message cache schema from `8` to `9` so stale cached `unknown` Codex rows are invalidated.

This is a narrow runtime parser/cache change. No frontend, release, workflow, migration, or CLI command behavior changed.

## Branch integrity
- Base branch: `main`
- `origin/main`: `97809e755b58b07e21c0eaceb9005f0ba073de0c`
- Ahead / behind from `git rev-list --left-right --count origin/main...HEAD`: `0 behind / 1 ahead`
- Merge base: `97809e755b58b07e21c0eaceb9005f0ba073de0c`
- `git merge-base --is-ancestor origin/main HEAD`: PASS
- Fast-forward safety: branch contains the current validated base.

## Commit integrity
- Introduced commit: `c3ca2ce9cfc113bc465dff8965f49da6cb83d865 fix(codex): attribute resumed token counts to later model`
- One logical change per commit: yes.
- Ledger-Id trailers: Not applicable — `scripts/ledger/` does not exist in this repository.
- Duplicate Ledger-Id check: Not applicable — `scripts/ledger/` does not exist in this repository.

## Ledger proof
Not applicable — `scripts/ledger/` does not exist in this repository.

## Diff hygiene
`git diff --name-status origin/main...HEAD`:

```text
M	crates/tokscale-core/src/lib.rs
M	crates/tokscale-core/src/message_cache.rs
M	crates/tokscale-core/src/sessions/codex.rs
```

- `git diff --check origin/main...HEAD`: PASS, no output.
[...]

PR #511, opened 2026-05-05
fix(codex): deduplicate forked token count history

### LINKED ISSUE #491: Overestimating token usage when folk.
When folk into local or worktree in Codex, it probably copy or the cache files, and the tokscale will double count the cached token usage (see the highest two bars, it is the time that I folk two projects), while only very limited tokens are actually used when folk.
Suggestion: try to avoid double counting the cached token usage when folk into another branch.
Image

IvGolovach:
## Summary
- Deduplicates Codex `token_count` rows that are replayed from a parent session into forked or worktree sessions.
- Adds a stable Codex token-count deduplication key and applies it in both parsed message paths.
- Preserves object-valued `session_meta.source` metadata for newer Codex session formats and bumps the source-message cache schema.

## Why
Codex fork/worktree sessions can include token history from the parent session before recording new activity. Without cross-file deduplication, Tokscale can count the same usage once from the parent session and again from the forked session.

Fixes #491.

## Test plan
- `cargo test -p tokscale-core codex` (59 passed)
- `cargo test -p tokscale-core` (639 passed, 1 ignored)
- `/opt/homebrew/bin/cargo test --workspace --all-features` (passed)
- `cargo fmt --all -- --check` (passed)
- `/opt/homebrew/bin/cargo-clippy --workspace --all-features -- -D warnings` (passed)
- `git diff --check` (passed)

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -813,8 +813,14 @@ fn parse_all_messages_with_pricing_with_env_strategy(
             )
         })
         .collect();
+    let mut codex_seen: HashSet<String> = HashSet::new();
     for (path, outcome) in codex_outcomes {
-        all_messages.extend(outcome.messages);
+        all_messages.extend(
+            outcome
+                .messages
+                .into_iter()
+                .filter(|message| should_keep_deduped_message(&mut codex_seen, message)),
+        );
         if let Some(entry) = outcome.cache_entry {
             source_cache.insert(entry);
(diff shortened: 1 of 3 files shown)


junhoyeo (approved):
Reviewed via parallel audit — **MERGE WITH NITS**. Strong dedup key (timestamp+provider+model+all 5 token fields), cache schema correctly bumped 10→11, stacks cleanly on #496. Negative-test nit ('twin tokens at different times') will land in a follow-up PR off main.

PR #533, opened 2026-05-10
fix(cursor): require fresh cache for each saved account

junhoyeo:
## Summary
- Makes implicit Cursor auto-sync freshness account-aware so a fresh secondary account CSV cannot mask missing or stale active-account cache data.

## Validation
- \; \
running 5 tests
.....
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 468 filtered out; finished in 0.00s

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 88 filtered out; finished in 0.00s; \.

## Risk
- Narrowly changes the freshness gate for implicit pre-report sync; manual sync remains unaffected.

---
## Summary by cubic
Make Cursor auto-sync freshness account-aware and add a retry cap so a permanently stale secondary account doesn’t force sync every run. Active `usage.csv` must be fresh; secondaries must be fresh or have a recent `usage.last-sync-attempt`.

- **Bug Fixes**
  - Skip implicit sync only when the active cache is fresh and each secondary cache is fresh or the `usage.last-sync-attempt` marker is recent; a missing cache dir or missing active cache is stale.
  - Touch `usage.last-sync-attempt` at the end of every sync (even on partial failures) to avoid repeated syncs when a secondary is broken.
  - Preserve the original fix so a fresh secondary can’t mask a stale/missing active cache, and dedupe expected paths when sanitized account IDs collide; add tests for these cases and the marker behavior.

<sup>Written for commit f9feae3575ceb3b32b8141879818519757945b57. Summary will update on new commits.</sup>

### DIFF
--- a/crates/tokscale-cli/src/cursor.rs
+++ b/crates/tokscale-cli/src/cursor.rs
@@ -13,10 +13,11 @@ use std::time::{Duration, SystemTime};
 /// proceeds against cached data after this timeout instead of stalling forever.
 const CURSOR_HTTP_TIMEOUT: Duration = Duration::from_secs(8);
 
-/// Skip implicit pre-report sync when the active account's `usage.csv` was
-/// modified within this window. Prevents `tokscale models` (and its siblings)
-/// from issuing a Cursor API call on every invocation. The manual `tokscale
-/// cursor sync` command bypasses this — explicit user intent is always honored.
+/// Skip implicit pre-report sync when every expected Cursor account cache file
+/// was modified within this window. Prevents `tokscale models` (and its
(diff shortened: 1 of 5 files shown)


junhoyeo:
Addressed by f9feae3. Added a per-attempt marker (`usage.last-sync-attempt`) so a permanently-stale secondary account doesn't thrash implicit sync on every invocation. Active cache is still required to be fresh — the original bug fix is preserved. New tests cover all three combinations (stale active blocks, fresh active + stale secondary + fresh marker passes, fresh active + stale secondary + no marker triggers retry).

PR #542, opened 2026-05-12
feat(tui): add Minutely tab for per-minute token breakdown

DigitalPhilosopher:
## Summary

Adds a new **Minutely** tab to the TUI that shows token usage aggregated per minute. This complements the existing Hourly tab and is useful for seeing burst patterns during focused coding sessions.

## Changes

- Added `Tab::Minutely` variant with navigation/sorting support
- New `MinutelyUsage` data model with per-minute aggregation logic
- New `minutely.rs` UI renderer for the table view
- Wired into the TUI tab bar and cache layer
- Updated unit tests for tab order, prev/next navigation, and keyboard shortcuts

## Usage

Open the TUI and press `Tab` / `Shift+Tab` or `←/→` to navigate to the **Minutely** tab (`Min`).

## Testing

- `cargo test` passes in `crates/tokscale-cli`
- Tab order, short name, and keyboard navigation covered by existing `mod tests`

---
## Summary by cubic
Adds a new Minutely tab in the TUI for per-minute token and cost breakdowns to help spot short bursts. The tab is gated behind a `minutelyTabEnabled` setting (default off) to avoid extra aggregation work unless enabled.

- **New Features**
  - Added Minutely tab (“Min”) between Hourly and Stats; supports sorting by Date/Tokens/Cost and navigation via Tab/Shift+Tab or ←/→.
  - New `minutelyTabEnabled` in `settings.json` (camelCase, default false). When off: DataLoader skips minute aggregation, the tab is hidden from the header and mouse targets, keyboard/tab cycling skips it, and a requested initial Minutely tab clamps to Overview.
  - Introduced `MinutelyUsage` with per-minute buckets (tokens, cost, messages, turns, clients, per-model stats); uses timestamp when available, falls back to message date; recomputed on load and not cached.
  - New `minutely.rs` table UI with responsive columns, cache read/write and hit-rate metrics, current-minute highlight, scrollbar, and an empty state.
  - Wired into app state, footer count label (“(x minutes)”), and tab bar; tests cover flag defaults/round-trip, tab visibility and navigation, minutely aggregation on/off, minute grouping/splitting, and clamping negative tokens/cost.

<sup>Written for commit 7cad4bc00a6641d7ce69cde6882fa3e03ad680b2. Summary will update on new commits. <a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/542?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/crates/tokscale-cli/src/tui/app.rs
+++ b/crates/tokscale-cli/src/tui/app.rs
@@ -257,13 +257,19 @@ impl App {
             config.since,
             config.until,
             config.year,
-        );
+        )
+        .with_minutely_enabled(settings.minutely_tab_enabled);
 
         let data = cached_data.unwrap_or_default();
         let has_data = !data.models.is_empty();
         let dialog_stack = DialogStack::new(theme.clone());
         let dialog_needs_reload = Rc::new(RefCell::new(false));
-        let current_tab = config.initial_tab.unwrap_or(Tab::Overview);
+        let requested_tab = config.initial_tab.unwrap_or(Tab::Overview);
+        let current_tab = if Self::tab_visible(&settings, requested_tab) {
+            requested_tab
+        } else {
+            Tab::Overview
+        };
         let (sort_field, sort_direction) = Self::default_sort_for_tab(current_tab);
 
         let mut app = Self {
@@ -414,22 +420,22 @@ impl App {
                 return true;
             }
             KeyCode::Tab => {
-                let next = self.current_tab.next();
+                let next = self.next_visible_tab();
                 self.switch_tab(next);
                 self.reset_selection();
             }
             KeyCode::BackTab => {
-                let prev = self.current_tab.prev();
+                let prev = self.prev_visible_tab();
                 self.switch_tab(prev);
                 self.reset_selection();
             }
             KeyCode::Left => {
-                let prev = self.current_tab.prev();
+                let prev = self.prev_visible_tab();
                 self.switch_tab(prev);
                 self.reset_selection();
             }
             KeyCode::Right => {
-                let next = self.current_tab.next();
+                let next = self.next_visible_tab();
                 self.switch_tab(next);
                 self.reset_selection();
             }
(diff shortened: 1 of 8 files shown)


junhoyeo:
Maintainer follow-up — force-pushed this branch with a rebase onto `main`.

After #553 merged (which centralized per-tab sort defaults via `default_sort_for_tab` and made `current_count_label` exhaustive in `tui/ui/footer.rs`), this branch picked up a conflict in `switch_tab` and a missing-arm compile error in `current_count_label` for the new `Tab::Minutely` variant.

What changed during the rebase:

- `tui/app.rs`: kept #553's `default_sort_for_tab(target)` indirection in `switch_tab`, and extended `default_sort_for_tab` itself to map both `Tab::Hourly` and `Tab::Minutely` to `Date/Descending`. This preserves your original "Minutely also defaults to date-desc" intent while staying inside the single-helper pattern.
- `tui/ui/footer.rs`: added the `Tab::Minutely => " ({n} minutes)"` arm so the exhaustive match in `current_count_label` still compiles, plus a matching assertion in `test_current_count_label_matches_active_tab`.
- `tui/ui/minutely.rs`: ran `cargo fmt` on the header `vec!` (single-line vs two-line break) so the Lint job passes.
- `Cargo.lock` is intentionally left at upstream's stale 2.1.0 state — main has the same mismatch with `Cargo.toml` and the release pipeline regenerates it.

Verification on the rebased commit:

- `cargo fmt --check`: PASS
- `cargo clippy -p tokscale-cli --bin tokscale --all-features -- -D warnings`: PASS
- `cargo test -p tokscale-cli --bin tokscale`: 485 passed, 0 failed, 1 ignored

The branch is now `CLEAN` against `main` and CI should pick it up automatically. Thanks for the Minutely tab work.

junhoyeo:
Maintainer follow-up — pushed one more commit (`feat(settings): gate Minutely tab behind opt-in flag, default off`) on top of the rebase.

Why the flag: minute-granularity is a niche/diagnostic feature for most users (daily/hourly cover the common case), and an external review flagged that the per-minute aggregation builds `HashMap<NaiveDateTime, MinutelyUsage>` over every parsed message — for users with months of data that is a meaningful RAM/CPU cost paid even when they never open the tab. The flag (`minutelyTabEnabled` in `settings.json`, default `false`) gates BOTH:

- The DataLoader skips the minute-bucket map entirely when the flag is off (`with_minutely_enabled` builder on construction).
- The header tab strip and click-area registration filter `Tab::Minutely` out, and Tab / BackTab / Left / Right navigation skip it via `App::next_visible_tab` / `prev_visible_tab`.
- `App::new_with_cached_data` clamps `initial_tab` to `Tab::Overview` when the requested tab is invisible under the current flag, so a stale `TuiConfig` pointing at Minutely cannot strand the user on a hidden view.

Users who want the feature flip one line in `~/.config/tokscale/settings.json`:

```json
{ "minutelyTabEnabled": true }
```

Tests added: settings round-trip + default-false coverage, the DataLoader aggregation gate (skip-when-off, run-when-on, same-minute grouping, adjacent-minute split, negative-token clamping), tab navigation with and without the flag, and the initial-tab clamp.

There is one outstanding follow-up from the external review that I am NOT addressing here, to keep this PR's scope tight: when the flag is enabled and the data set is very large, `get_sorted_minutely()` re-sorts the whole list every render frame, which can cause TUI jank for 100k+ minute buckets. With the default-off flag, default users are unaffected; I will open a separate issue to track the sort-caching work for opt-in users.

Verification on the pushed commit:

- `cargo fmt --check`: PASS
- `cargo clippy -p tokscale-cli --bin tokscale --all-features -- -D warnings`: PASS
- `cargo test -p tokscale-cli --bin tokscale`: 495 passed, 0 failed, 1 ignored

Thanks for the Minutely tab — it's a great diagnostic surface for power users. Hope the flag approach feels reasonable.

PR #550, opened 2026-05-13
fix(pricing): require exact matches for provider-scoped models

### LINKED ISSUE #523: [Bug] Pricing: wrong Fireworks model matched for ‎⁠accounts/fireworks/models/deepseek-v4-pro⁠
### Summary

When I look up pricing for the Fireworks DeepSeek model, Tokscale matches the wrong key in the LiteLLM pricing table.

Command:

```zsh
tokscale pricing "accounts/fireworks/models/deepseek-v4-pro"
```

Output (relevant part):

Pricing for: accounts/fireworks/models/deepseek-v4-pro
Matched key: fireworks_ai/accounts/fireworks/models/deepseek-r1-0528-distill-qwen3-8b
  Source: LiteLLM
  Input:  $0.20 / 1M tokens
  Output: $0.20 / 1M tokens

This is not the correct model: ‎⁠deepseek-v4-pro⁠ and
‎⁠deepseek-r1-0528-distill-qwen3-8b⁠ have different pricing on Fireworks.

Expected behavior :

For ‎⁠accounts/fireworks/models/deepseek-v4-pro⁠:

- Tokscale should either:

 ▫ find an exact matching key for this model in LiteLLM/OpenRouter and use that, or

 ▫ if no exact key exists, fail gracefully (e.g. “no pricing found”) instead of
fuzzy‑matching to a different Fireworks model.

Ideally, ‎⁠accounts/fireworks/models/*⁠ would prefer exact or prefix‑exact matches
before falling back to fuzzy matching.

Actual behavior

The fuzzy matching phase in the pricing lookup seems to pick the first Fireworks
entry with a similar name:

- Input model: ‎⁠accounts/fireworks/models/deepseek-v4-pro⁠

- Matched model: ‎⁠fireworks_ai/accounts/fireworks/models/deepseek-r1-0528-distill-qwen3-8b⁠

This leads to an incorrect cost estimate for all my ‎⁠deepseek-v4-pro⁠ usage.

Environment

- Tokscale version: (output of tokscale 2.1.0⁠)

- Install method: ‎⁠npx tokscale@latest⁠

- OS: macOS 26.4.1 (25E253)

Suggested fix (idea)

- Special‑case Fireworks keys so that:

 ▫ exact match on ‎⁠accounts/fireworks/models/deepseek-v4-pro⁠ wins over any fuzzy match

 ▫ if there is no exact entry for that key in LiteLLM/OpenRouter, return “no pricing
found” instead of mapping it to another model like
‎⁠deepseek-r1-0528-distill-qwen3-8b⁠.

 ▫ add the possibility to custom price

IvGolovach:
## Summary

Fixes #523.

Provider-scoped model IDs such as `accounts/fireworks/models/deepseek-v4-pro` now resolve through exact/provider-exact pricing paths only. If Tokscale cannot find an exact Fireworks-scoped price, it reports no pricing instead of stripping the model suffix and fuzzy-matching an unrelated Fireworks DeepSeek model.

## Why

The previous fallback path could shorten structured gateway IDs before fuzzy lookup. For Fireworks this allowed `accounts/fireworks/models/deepseek-v4-pro` to match a different key such as `fireworks_ai/accounts/fireworks/models/deepseek-r1-0528-distill-qwen3-8b`, producing incorrect cost estimates.

## Diff scope

- Detect structured `accounts/<provider>/models/<model>` and `accounts/<provider>/routers/<model>` pricing IDs in the core resolver.
- Allow raw exact matches, reseller-prefixed exact matches, and provider-aware exact matches on the terminal model ID.
- Block suffix stripping, prefix stripping, and broad fuzzy matching for provider-scoped paths when no exact provider-scoped price exists.
- Add core resolver, repricing, and CLI regression coverage for the reported Fireworks command.

## Validation

- `cargo fmt --check`: PASS
- `git diff --check`: PASS
- `cargo test -p tokscale-core provider_scoped_path -- --nocapture`: PASS, 4 passed
- `cargo test -p tokscale-core pricing::lookup -- --nocapture`: PASS, 130 passed
- `cargo test -p tokscale-core`: PASS, 673 passed, 1 ignored
- `cargo test -p tokscale-cli test_pricing_command_does_not_fuzzy_match_provider_scoped_fireworks_model --test cli_tests -- --exact --nocapture`: PASS, 1 passed
- `cargo clippy -p tokscale-core --all-targets -- -D warnings`: PASS

## Rollback

Revert this PR.

## Known residual risks

This intentionally returns "model not found" for provider-scoped gateway IDs when the matching provider-specific price is absent. That is safer than applying a different provider or model price; custom pricing overrides can still address missing upstream data separately.

### DIFF
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -381,6 +381,49 @@ fn write_pricing_cache(base: &Path, timestamp: u64) {
     }
 }
 
+fn write_fireworks_pricing_cache(base: &Path) {
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("system time before unix epoch")
+        .as_secs();
+    let litellm = serde_json::json!({
+        "timestamp": now,
+        "data": {
+            "fireworks_ai/accounts/fireworks/models/deepseek-r1-0528-distill-qwen3-8b": {
+                "input_cost_per_token": 0.0000002,
+                "output_cost_per_token": 0.0000002
+            }
+        }
+    });
+    let openrouter = serde_json::json!({
+        "timestamp": now,
+        "data": {
+            "deepseek/deepseek-v4-pro": {
+                "input_cost_per_token": 0.000001,
+                "output_cost_per_token": 0.000002
+            }
+        }
+    });
+
+    for dir in [
+        base.join(".config/tokscale/cache"),
+        base.join("Library/Caches/tokscale"),
+        base.join(".cache/tokscale"),
+    ] {
+        fs::create_dir_all(&dir).unwrap();
+        fs::write(
+            dir.join("pricing-litellm.json"),
+            serde_json::to_vec(&litellm).unwrap(),
+        )
+        .unwrap();
+        fs::write(
+            dir.join("pricing-openrouter.json"),
+            serde_json::to_vec(&openrouter).unwrap(),
+        )
+        .unwrap();
+    }
+}
+
 fn write_fake_credentials(base: &Path) {
     let creds_dir = base.join(".config/tokscale");
     fs::create_dir_all(&creds_dir).unwrap();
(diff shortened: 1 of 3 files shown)


Saidelocha:
Thanks !

junhoyeo (approved):
LGTM — strict matching is the correct safety posture and is well-covered.

junhoyeo:
@IvGolovach @Saidelocha this has been merged to v3.0.0: https://github.com/junhoyeo/tokscale/releases/tag/v3.0.0 thanks for the contribution!

PR #562, opened 2026-05-17
fix(antigravity): decode chunked identity probe responses

ddarkr:
## Summary
* Fix `tokscale antigravity sync/status` reporting `detected connections: 0` while Antigravity is running because the identity probe did not decode chunked HTTP responses from live language-server RPC endpoints.
* Decode identity probe `Content-Length` and chunked bodies while keeping probe reads capped to the existing 4 KiB trust-boundary limit.
* Add tests for chunked identity responses and oversized probe-body prefixes so the detection path keeps working without widening full RPC limits.

## What was wrong
The live Antigravity `GetAllCascadeTrajectories` identity check can return `Transfer-Encoding: chunked` with the Antigravity marker keys near the start of a larger JSON response. The old identity probe read the response body as raw text without chunk decoding, so it inspected chunk framing instead of JSON and rejected the endpoint, leaving `detected connections` and `detected sessions` at `0` even though the language server was running.

## How this fixes it
The identity probe now parses HTTP status and headers, reads fixed-length responses with a 4 KiB cap, decodes chunked responses into a capped 4 KiB prefix, and falls back to conservative marker-key detection for intentionally truncated JSON prefixes. Full RPC calls continue to use the existing 16 MiB body cap.

## Verification
* `cargo +stable fmt --all`
* `cargo +stable test antigravity`
* `cargo +stable check -p tokscale-cli`
* `target/debug/tokscale --no-spinner antigravity sync` → `detected connections: 2`, `detected sessions: 89`, `cached sessions after sync: 186`
* `target/debug/tokscale --no-spinner antigravity status --json` → `detectedConnections: 2`, `cachedSessions: 186`

---
## Summary by cubic
Fixes Antigravity identity probing so `tokscale antigravity sync/status` correctly detects live connections by decoding chunked and fixed-length HTTP responses and preferring `Transfer-Encoding: chunked` over `Content-Length` when both are present. Keeps the 4 KiB probe cap and the 16 MiB RPC limit.

- **Bug Fixes**
  - Decode `Content-Length` and `Transfer-Encoding: chunked` bodies with a 4 KiB cap; prefer chunked per RFC 7230 in both probe and RPC paths.
  - Detect Antigravity markers in truncated JSON prefixes to handle partial bodies.
  - Add tests for chunked decoding, mixed headers (chunked + `Content-Length`), and oversized bodies.

<sup>Written for commit da02e376dbe6b8eb9f8d77941c43bf1fae0f96ab. Summary will update on new commits. <a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/562?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/crates/tokscale-cli/src/antigravity.rs
+++ b/crates/tokscale-cli/src/antigravity.rs
@@ -1086,15 +1086,18 @@ fn identity_probe_request(port: u16, csrf_token: &str, method: &str) -> Option<S
         return None;
     }
 
+    // RFC 7230 §3.3.3: when Transfer-Encoding is present, Content-Length MUST
+    // be ignored. Check chunked first so a server that sets both headers is
+    // decoded correctly.
+    if chunked {
+        return read_chunked_body_prefix(&mut reader, MAX_IDENTITY_PROBE_BYTES).ok();
+    }
+
     if let Some(length) = content_length {
         let read_length = length.min(MAX_IDENTITY_PROBE_BYTES);
         let mut bytes = vec![0_u8; read_length];
         reader.read_exact(&mut bytes).ok()?;
         return String::from_utf8(bytes).ok();
-    }
-
-    if chunked {
-        return read_chunked_body_prefix(&mut reader, MAX_IDENTITY_PROBE_BYTES).ok();
     }
 
     let mut buffer = String::new();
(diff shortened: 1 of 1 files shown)


junhoyeo:
Maintainer follow-up — pushed one extra commit to this branch.

While reviewing the chunked decoder I noticed both `identity_probe_request` and `rpc_request` check `Content-Length` before `Transfer-Encoding: chunked`. Per RFC 7230 §3.3.3, when both headers are present the receiver must ignore `Content-Length` and decode as chunked; the previous ordering would silently misframe responses from any server that sets both headers (some HTTP stacks add `Content-Length: 0` alongside `Transfer-Encoding: chunked` on the wire).

The follow-up commit swaps the branch order in both code paths and adds a regression test (`identity_probe_request_prefers_chunked_over_content_length`) that serves chunked Antigravity JSON with a bogus `Content-Length: 1` header — under the previous ordering this test fails, under the new ordering it passes.

Everything else in the PR looks great — the core fix, the 4 KiB cap preservation, and the marker fallback for truncated bodies are all solid. Thanks for the careful work and the runtime evidence in the PR description.

PR #564, opened 2026-05-17
feat(tui): open daily detail rows for selected date

xczllgit:
## Summary
- Adds an Enter-driven detail view for the selected date in the Daily tab.
- Renders selected-day usage with model-style detail rows for client/provider/model/workspace breakdowns.
- Updates Esc handling and footer hints so users can return cleanly from the detail view.

## Test Plan
- [x] `cargo test` from `crates/tokscale-cli`
- [x] `cargo fmt --check`

---
## Summary by cubic
Adds a Daily detail view with per-source/provider/model breakdown; Enter opens, Esc/Backspace returns. Closing detail now restores the same day after sort changes, and refresh exits detail if the day disappears.

- **New Features**
  - Enter opens model detail rows for the selected day; columns adapt to narrow/very narrow layouts, with sort indicators for tokens/cost.
  - Sorting in detail view respects tokens/cost; selection and scroll are preserved when entering/exiting.
  - Footer updates with "enter:details" or "esc:back", and the right-side count shows models in the detail view.

- **Bug Fixes**
  - Closing detail re-anchors the Daily list by date, keeping the same day highlighted after sort changes; viewport preserved when possible.
  - Auto-refresh exits detail mode if the selected date is removed to avoid a stuck empty view.

<sup>Written for commit e6cf3d898ffb5f86347a03131d1b65c45b09b45e. Summary will update on new commits. <a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/564?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/crates/tokscale-cli/src/tui/app.rs
+++ b/crates/tokscale-cli/src/tui/app.rs
@@ -4,6 +4,7 @@ use std::rc::Rc;
 use std::time::{Duration, Instant};
 
 use anyhow::Result;
+use chrono::NaiveDate;
 use crossterm::event::{KeyCode, KeyEvent, KeyModifiers, MouseButton, MouseEvent, MouseEventKind};
 use ratatui::layout::Rect;
 use tokscale_core::ClientId;
@@ -12,7 +13,9 @@ use crate::ClientFilter;
 
 use ratatui::style::Color;
 
-use super::data::{AgentUsage, DailyUsage, DataLoader, HourlyUsage, ModelUsage, UsageData};
+use super::data::{
+    AgentUsage, DailyUsage, DataLoader, HourlyUsage, ModelUsage, TokenBreakdown, UsageData,
+};
 use super::settings::Settings;
 use super::themes::{Theme, ThemeName};
 use super::ui::dialog::{ClientPickerDialog, DialogStack};
@@ -129,6 +132,17 @@ pub struct ClickArea {
     pub action: ClickAction,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct DailyDetailRow<'a> {
+    pub source: &'a str,
+    pub provider: &'a str,
+    pub model: &'a str,
+    pub color_key: &'a str,
+    pub tokens: &'a TokenBreakdown,
+    pub cost: f64,
+    pub messages: u64,
+}
+
 #[derive(Debug, Clone)]
 pub enum ClickAction {
     Tab(Tab),
@@ -160,6 +174,9 @@ pub struct App {
     pub scroll_offset: usize,
     pub selected_index: usize,
     pub max_visible_items: usize,
+    pub selected_daily_detail_date: Option<NaiveDate>,
+    daily_list_selected_index: usize,
+    daily_list_scroll_offset: usize,
 
     pub selected_graph_cell: Option<(usize, usize)>,
     pub stats_breakdown_total_lines: usize,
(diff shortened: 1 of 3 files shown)


junhoyeo:
Maintainer follow-up — force-pushed this branch with a rebase onto `main`.

After #553 merged (which centralized per-tab sort persistence via `persist_current_sort()` and extracted the footer count label into `current_count_label`), this branch picked up two conflicts.

What changed during the rebase:

- `tui/app.rs` (`set_sort`): now calls `self.persist_current_sort()` first (preserving #553's invariant that every sort mutation records the active tab's state immediately), then runs your existing Daily-detail-aware selection-reset branch. The behavior is unchanged from your original PR for the Daily-detail and non-Daily-detail paths; the addition is the eager `persist_current_sort()` call, which keeps `tab_sort_state` accurate if the user quits without switching tabs.
- `tui/ui/footer.rs`: rewrote the inline `match app.current_tab { ... }` block in `render_main_row` into a single `current_count_label(app)` call, and moved the Daily-detail arm into the helper itself as `Tab::Daily if app.is_daily_detail_active()` returning `({n} models)`. This preserves the Daily-detail count behavior while routing everything through the single helper introduced by #553. The other tab labels (`days`, `hours`, `models`, `agents`, `Stats => ""`) are unchanged from `main`.
- `Cargo.lock` is intentionally left at upstream's stale 2.1.0 state — main has the same mismatch with `Cargo.toml` and the release pipeline regenerates it.

Verification on the rebased commit:

- `cargo fmt --check`: PASS
- `cargo clippy -p tokscale-cli --bin tokscale --all-features -- -D warnings`: PASS
- `cargo test -p tokscale-cli --bin tokscale`: 487 passed, 0 failed, 1 ignored (includes #553's `test_current_count_label_matches_active_tab` and `test_switch_tab_preserves_daily_sort_after_hourly_roundtrip`, plus the Daily-detail tests this PR adds)

The branch is now `CLEAN` against `main` and CI should pick it up automatically. Thanks for the Daily detail-rows work.

junhoyeo:
Maintainer follow-up — pushed one more commit with two daily-detail state-machine fixes that an external review surfaced:

1. **`close_daily_detail()` now re-anchors by date.** Before this fix, changing the Daily sort while in detail mode (Cost → Date or vice versa) and then pressing Esc would restore the cached `daily_list_selected_index`, which now points at a different day in the re-sorted list. The fix finds the original detail date in the freshly-sorted Daily list and selects that position, falling back to the cached index only when the date is gone.
2. **`update_data()` now drops detail mode when the selected date disappears.** Before this fix, if `update_data()` was called (auto-refresh or manual refresh) and the new `UsageData.daily` didn't include the date the user was viewing, `selected_daily_detail_date` would remain `Some(...)` while `get_sorted_daily_detail_rows()` returned an empty list — leaving the UI on a ghost detail view that couldn't be navigated out of with arrow keys. The fix exits detail mode cleanly in that case.

Tests added for both transitions plus a positive-path test that detail mode is preserved when the selected date survives a refresh.

Verification on the pushed commit:

- `cargo fmt --check`: PASS
- `cargo clippy -p tokscale-cli --bin tokscale --all-features -- -D warnings`: PASS
- `cargo test -p tokscale-cli --bin tokscale`: 490 passed, 0 failed, 1 ignored (includes the 3 new tests plus all existing detail/sort/footer coverage from this PR)

The rest of the PR (Enter/Esc/Backspace wiring, sort logic, footer label, daily detail view) all looked solid in review. Thanks for the daily detail rows work — this is going to be useful.

Issue #560, opened 2026-05-16
Bug: cross-machine submissions overwrite each other instead of merging — per-device totals get clobbered


PeterPonyu:
## Summary

When the same tokscale account is used on multiple machines, running `tokscale submit` on each machine causes the latest submission to **overwrite** the prior machine's records instead of merging them. The leaderboard shows only the most recently-submitted machine's totals, not the sum across all of the user's devices.

## Steps to reproduce

1. On **Machine A** (e.g. a laptop), run:
   ```bash
   tokscale login
   tokscale submit
   ```
   The leaderboard reflects Machine A's usage.
2. On **Machine B** (e.g. a desktop) — same tokscale account / same token from `~/.config/tokscale/credentials.json` — run:
   ```bash
   tokscale login
   tokscale submit
   ```
3. Refresh the leaderboard.

## Expected

The leaderboard total = (Machine A usage) + (Machine B usage), de-duplicated by session/turn ID so the same session isn't double-counted, but additive across distinct sessions/clients.

Equivalent of an upsert-and-union per `{client, session-id, timestamp}` key, not a replace-by-account.

## Actual

The leaderboard shows only Machine B's most recent submission. Machine A's data appears to be either replaced wholesale or silently dropped on the server side. Running `tokscale submit` again on Machine A flips it back, suggesting last-writer-wins on the per-account record rather than additive merge.

## Independent evidence — the community already worked around this

Three forks exist specifically to solve this:

| Fork | Purpose |
|---|---|
| [`jsserve-org/codex-tokscale-hub`](https://github.com/jsserve-org/codex-tokscale-hub) | "Fork of Tokscale with a multi-device merge proxy flow" |
| [`jsserve-org/tokscale-merge`](https://github.com/jsserve-org/tokscale-merge) | "Tokscale merge data between devices" |
| [`gigio1023/tokscale-bundle`](https://github.com/gigio1023/tokscale-bundle) | "Companion CLI for bundling tokscale-discoverable session data across [machines]" |

From `jsserve-org/codex-tokscale-hub` README (verbatim):

> A local aggregation proxy for tokscale. Instead of each device submitting directly to tokscale.ai, every device submits to this hub. The hub merges all device data and pushes a single combined submission upstream.

The fact that this workaround pattern was independently re-implemented by multiple users confirms the bug is reproducible and impactful.

## Impact

[...]

PR #589, opened 2026-05-24
fix(submit): adopt legacy daily rows into device buckets

IvGolovach:
## Summary
- Adopt a user's lone legacy `daily_breakdown` bucket into the first modern submitted device during `/api/submit`, so post-migration device-aware submissions do not duplicate historical totals.
- Keep the current `submitted_devices` architecture instead of reintroducing the older `submissions.source_id` design from #389.
- Add regression coverage for legacy-device adoption and same-date profile aggregation across submitted devices.

## Why
#560 reports cross-machine submissions clobbering or undercounting totals. Current `main` already stores daily rows by submitted device, but migrated historical rows live under the `legacy-default` device bucket. Without a guarded cutover path, the first device-aware submit for an existing user can create a second row for the same day and inflate totals instead of treating the legacy row as that user's initial device bucket.

This PR rescues the still-relevant part of #389 for the current schema: it fixes the legacy-to-device transition without adding a new source-scoped `submissions` model, without dropping `submission_hash`, and without changing the CLI payload shape.

## Diff scope
- `POST /api/submit`: when a non-legacy device submits to an existing account with no rows for that device, move the user's legacy daily rows into that device only if no modern device rows already exist for the submission.
- `TokenContributionData`: expose optional `device` and `timeMetrics` fields so the frontend type matches accepted submit payloads.
- Tests: add targeted API/profile regressions for no-duplication cutover and read-path aggregation across submitted devices.

## Branch integrity
- Base: `origin/main`
- Validated base SHA: `1fa26f571b6f6f0affd1c84796f0d032e3e9ee33`
- Head SHA: `c406ddcbae1a627e6ce02ef70913e3a85e7e1809`
- Ahead/behind: `0 behind / 1 ahead`
- Merge base: `1fa26f571b6f6f0affd1c84796f0d032e3e9ee33`

## Validation
- `bun x vitest run __tests__/api/submitAuth.test.ts __tests__/api/submit.test.ts __tests__/api/usersProfile.test.ts __tests__/lib/getLeaderboard.test.ts __tests__/lib/getGroupLeaderboard.test.ts`: PASS, 5 files and 58 tests
- `bun x eslint src/app/api/submit/route.ts src/lib/types.ts __tests__/api/submitAuth.test.ts __tests__/api/usersProfile.test.ts`: PASS, no output
- `bash scripts/check-version-coherence.sh`: PASS (`Version coherence OK: 2.1.3`)

## Diagnostics
- `bun x tsc --noEmit --pretty false`: not selected as a merge gate for this PR. It currently fails on unrelated pre-existing frontend fixture/asset typing issues outside this diff: missing `totalTokens` / `totalCost` in render SVG test fixtures, and the `BlackholeHero` asset import typing.

## Migration notes
No DB migration changed. The fix uses the existing `submitted_devices` and `daily_breakdown.submitted_device_id` schema.

## Runtime safety
The adoption query is guarded so it only moves legacy rows when no modern submitted-device rows exist for that submission. Ambiguous mixed legacy/modern accounts are preserved rather than reassigned. No invariant regression introduced.

## Rollback plan
Rollback: revert this PR.
DB downgrade: not applicable.
Data repair: not applicable.
Operational caveats: none known.

## Known residual risks
- Remote CI has not run until the PR is opened.
- The legacy adoption path intentionally does not attempt to infer ownership when modern device rows already exist, because that attribution would be ambiguous.

### DIFF
--- a/packages/frontend/__tests__/api/submitAuth.test.ts
+++ b/packages/frontend/__tests__/api/submitAuth.test.ts
@@ -615,4 +615,187 @@ describe("POST /api/submit auth path", () => {
       mode: "merge",
     }));
   });
+
+  it("adopts lone legacy daily rows into the first modern submit device instead of duplicating totals", async () => {
+    mockState.authenticatePersonalToken.mockResolvedValue({
+      status: "valid",
+      tokenId: "token-1",
+      userId: "user-1",
+      username: "alice",
+      displayName: "Alice",
+      avatarUrl: null,
+      isAdmin: false,
+      expiresAt: null,
+    });
+
+    mockState.validateSubmission.mockReturnValue({
+      valid: true,
+      data: {
+        device: {
+          id: "dev_laptop",
+          name: "Laptop",
+        },
+        meta: {
+          version: "2.0.0",
+          dateRange: { start: "2026-04-30", end: "2026-04-30" },
+        },
+        summary: {
+          clients: ["codex"],
+        },
+        contributions: [
+          {
+            date: "2026-04-30",
+            timestampMs: 456,
+            clients: [
+              {
+                client: "codex",
+                modelId: "gpt-5.5",
+                tokens: 15,
+                cost: 0.75,
+                input: 10,
+                output: 5,
+                cacheRead: 0,
+                cacheWrite: 0,
+                reasoning: 0,
+                messages: 1,
+              },
+            ],
+          },
+        ],
+      },
+      errors: [],
+      warnings: [],
+    });
+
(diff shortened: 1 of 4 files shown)


junhoyeo:
Thanks for the fix here. I merged #609, which carries the same legacy daily-row adoption behavior on the newer base and adds the extra regression coverage for the ambiguous case where a modern device already has rows. Since this branch now conflicts with main and is superseded by #609, I’m closing this one to keep the submit queue clear.

PR #678, opened 2026-06-05
fix(submission): raise per-day token/cost ceilings and add warn bands

RedesignedRobot:
## Problem

`validateSubmission` rejects legitimate single-day usage. The per-day caps (`MAX_DAILY_TOKENS = 10B`, `MAX_DAILY_COST = $10k`) are applied per day, per client, and to `maxCostInSingleDay`, so one heavy day fails with a wall of errors:

```
Summary maxCostInSingleDay exceeds 10,000: 11185.52
Daily token total exceeds 10,000,000,000 on 2026-06-05: 12,910,573,877
Daily cost exceeds 10,000 on 2026-06-05: 11185.52
Client codex on 2026-06-05: token total exceeds 10,000,000,000: 12,910,573,877
Client codex on 2026-06-05: cost exceeds 10,000: 11185.52
```

These run server-side, so the CLI just surfaces the rejection and the user is locked out of submitting. Two things make the old limits reachable in a single real day:

1. **cacheRead dominates agentic usage.** Tools re-read large contexts every turn, so `cacheRead` routinely runs 10-100x `output` while carrying little to no cost. A day's token total is mostly cache reads. The example above is 12.9B tokens at ~$0.87/M blended, which is a normal rate, not an inflated one.
2. **Parallel multi-agent runs.** Fanning out dozens of concurrent sessions on one high-end device multiplies a single device's daily volume.

## Changes

Extend the existing two-tier pattern (the token band already worked this way; cost now matches it):

- `MAX_DAILY_TOKENS` 10B -> 50B, `WARN_DAILY_TOKENS` 5B -> 10B
- `MAX_DAILY_COST` $10k -> $50k, plus a new `WARN_DAILY_COST` $10k warn band (logs `[submission] high daily cost`, mirroring the existing high-token warning)
- `MAX_COST_PER_MILLION_TOKENS` unchanged at $10k/M

`WARN_*` logs a structured warning for visibility but does not reject, so real outliers submit and anomalies stay greppable in logs. `MAX_*` still hard-rejects order-of-magnitude data errors. Unlike f6aeca7 (which raised the token cap to 100B with no warn band and was reverted for hiding data errors), this keeps a hard ceiling **and** a warn band. The real fraud backstop, `MAX_COST_PER_MILLION_TOKENS`, is left unchanged on purpose: it rejects "huge tokens + huge cost" fabrications regardless of these per-day caps.

## Test plan

`bun run --cwd packages/frontend test __tests__/api/submit.test.ts` -> 53 passed.

Added to the existing guardrail suite:

- accepts a real cache-read-heavy day (12.9B tokens / $11,185.52, single client) that the old caps rejected
- still rejects a day above the 50B token ceiling (assertion updated to the new value)
- still rejects a day above the $50k cost ceiling

Unchanged guardrails still pass: cost-per-million, tokenless-cost, structural/schema validation.

---
## Summary by cubic
Relaxed daily submission caps and added warn bands to prevent legitimate high-volume single-day usage from being rejected while keeping hard ceilings and clear log visibility.

- **Bug Fixes**
  - Raised per-day ceilings: tokens 10B -> 50B (warn at 10B), cost $10k -> $50k (warn at $10k).
  - Added structured warning for high daily cost; still hard-reject above caps.
  - Kept `MAX_COST_PER_MILLION_TOKENS` at $10k/M.
  - Updated tests: accept a 12.9B tokens / $11,185 day; still reject >50B tokens and >$50k cost.

<sup>Written for commit 9ae3f6bfa9b41244ebd584bd131a45af018b7233. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/678?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/__tests__/api/submit.test.ts
+++ b/packages/frontend/__tests__/api/submit.test.ts
@@ -441,8 +441,7 @@ describe('POST /api/submit - Client-Level Merge', () => {
 
   describe("Submission validation guardrails", () => {
     it("accepts internally consistent high-volume one-day token totals", () => {
-      // Cap is 10B (MAX_DAILY_TOKENS). Use a value just under to confirm the
-      // cap boundary is respected without triggering a rejection.
+      // Well under the warn band (10B). Confirms ordinary high days stay clean.
       const payload = createValidationPayload({
         totalTokens: 8_000_000_000,
         totalCost: 800,
@@ -461,7 +460,29 @@ describe('POST /api/submit - Client-Level Merge', () => {
       expect(result.errors).toHaveLength(0);
     });
 
-    it("rejects one-day fabricated token totals above the submission cap", () => {
+    it("accepts a real cache-read-heavy day above the prior 10B/$10k caps", () => {
+      // Single-client day that the old caps rejected: 12.9B tokens (mostly
+      // cacheRead) and $11,185.52, with a plausible ~$0.87/M blended rate. Sits
+      // in the warn band for both tokens and cost, so it logs but does not reject.
+      const payload = createValidationPayload({
+        totalTokens: 12_910_573_877,
+        totalCost: 11_185.52,
+        tokenBreakdown: {
+          input: 1_000_000_000,
+          output: 700_000_000,
+          cacheRead: 11_000_000_000,
+          cacheWrite: 200_000_000,
+          reasoning: 10_573_877,
+        },
+      });
+
+      const result = validateSubmission(payload);
+
+      expect(result.valid).toBe(true);
+      expect(result.errors).toHaveLength(0);
+    });
+
+    it("rejects one-day fabricated token totals above the hard ceiling", () => {
       const payload = createValidationPayload({
         totalTokens: 1_000_000_000_000,
         tokenBreakdown: {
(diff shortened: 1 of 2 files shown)


IvGolovach:
Thanks a lot for putting this together.

I took a careful look at the PR, the current validation path, and the prior history around high-volume submit failures. We already have sanity checks for fabricated totals, especially the cost-per-million guard, and there was also a previous cap-raise attempt that made it clear we should not simply hide parser/data errors behind very loose limits.

This PR looks like a more balanced version of that idea to me: it keeps hard ceilings, adds warning bands for visibility, and still preserves the density-based guardrail. The focused submit validation tests are passing, and I do not see a technical blocker from my side.

That said, this is a sensitive validation-policy change because it affects what the server accepts from self-reported usage payloads, so I would really like @junhoyeo to take a look before we merge it. Thanks again, this is a useful and thoughtful fix.

PR #720, opened 2026-06-17
docs(readme): sync EN/ja/ko/zh — Command Code, Antigravity CLI, report cmd, Warp/Cline

junhoyeo:
Syncs all four READMEs with **README.md as the canonical lead**, closing gaps surfaced after the recent client batch (#718/#717/#713/#715/#633).

## What changed
- **Command Code** (#717) and **Antigravity CLI** (#713) added to the client support table, multi-platform feature list, source-filtering list, and the `--client` possible-values list — in **all four locales** (were missing entirely).
- **`tokscale report`** (#633) command section documented in **ja/ko/zh** (was English-only). Notes that LLM summarization is **on by default** (apple-fm backend, `--no-summarize` to opt out), the local wiki SQLite DB, and task grouping.
- Back-filled **pre-existing translation drift**: Warp/Oz + Cline client rows, feature-list entries, and the **Warp/Oz Commands** section in ja/ko/zh; fixed a malformed (logo-less) Grok Build row in ko/zh.

## Result
All four files now share an **identical 33-client roster** and an identical `--client` list (verified: client `alt=` sets match EN exactly; row counts 33/33/33/33).

## Note
Machine-translated for ja/ko/zh following the EN source; a native-speaker glance on nuance is welcome, but code/flags/paths are kept verbatim.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Sync EN/ja/ko/zh READMEs with the EN lead, adding missing clients and documenting the task-based `tokscale report`, plus fixing summarizer prerequisites and platform-specific paths. All locales now list 33 clients with matching `--client` values.

- **New Features**
  - Added Command Code and Antigravity CLI to the support table, multi-platform list, source filter list, and `--client` values in all locales.
  - Documented `tokscale report` in ja/ko/zh: summarize on by default, `apple-fm` backend, `--no-summarize` opt-out, local wiki SQLite DB, and task grouping.
  - Expanded Kiro entry with IDE globalStorage snapshots.

- **Bug Fixes**
  - Back-filled translation drift: added Warp/Oz and Cline rows and the Warp/Oz Commands section in ja/ko/zh; fixed malformed Grok Build row in ko/zh.
[...]

PR #721, opened 2026-06-17
feat(report): native Rust apple-fm summarizer via FoundationModels FFI (drop Python)

junhoyeo:
Replaces the `python3 scripts/wiki-summarizer.py` `apple-fm` summarizer backend (added in #633) with **native Rust FFI into Apple's on-device FoundationModels**. Removes the only Python in the repo and the unshipped-script footgun flagged in #720 (npm/bunx users hit a missing `wiki-summarizer.py`).

## Why Python existed / why this is "FFI not pure Rust"
Apple Foundation Models is a **Swift-only** framework (no Obj-C, so `objc2` can't reach it). The *only* way any non-Swift language talks to it is a compiled **C-ABI shim** — which is exactly what the official `python-apple-fm-sdk` does under the hood. This PR does the same thing from Rust instead of from Python.

## How
- **Vendor** Apple's Apache-2.0 [`foundation-models-c`](https://github.com/apple/python-apple-fm-sdk) SwiftPM package (the C-ABI shim) under `crates/tokscale-cli/vendor/` (LICENSE + NOTICE included, per-file copyright headers preserved).
- **`build.rs`** compiles it with `swift build` and links `libFoundationModels.dylib` **only** when the new optional `apple-fm` Cargo feature is set **and** `target_os = macos`. It is a complete no-op otherwise → default/cross-platform builds are untouched.
- **`commands/apple_fm.rs`** — FFI bindings + a safe `summarize()` that:
  - **gates on `FMSystemLanguageModelIsAvailable`** before any generation (Apple's own `fm-c-example` *segfaults* generating against an unavailable model),
  - drives structured generation via `FMLanguageModelSessionRespondWithSchemaFromJSON`, blocks on the background callback with a **60s timeout**, and validates/coerces the result.
- Port the Python `fallback_classify` **heuristic to Rust** as the always-available fallback. `summarize()` returns `None` (never errors) when FM is unavailable / feature-off / non-macOS, and the caller applies the heuristic — so **`apple-fm` stays the default backend and degrades gracefully everywhere** (your "optional + cross-platform" requirement).
- **Delete** `scripts/wiki-summarizer.py`, the `python3` invocation, and `find_summarizer_script`.

## Verification (macOS 26.2 / arm64 / swiftc 6.3.2)
- ✅ Default build (no feature) + `clippy` + `fmt` clean — cross-platform path uses the stub.
- ✅ `cargo build --features apple-fm` compiles, links the dylib, and the binary loads it at runtime (rpath); `clippy --features apple-fm` clean.
- ✅ 10 heuristic + 11 report tests pass.
- ⚠️ **Not exercised: live on-device generation** — Apple Intelligence is disabled on the build machine, so only the availability-gate + heuristic-fallback path ran. **Needs validation on an Apple-Intelligence-enabled Mac** (build with `--features apple-fm`, run `tokscale report`).

## Notes
- Overlaps with #720 (docs) on the README report section — whichever lands second rebases; #720's `apple-fm` "not bundled in npm" caveat becomes obsolete once this lands (no more script).
- Default release builds do **not** enable `apple-fm` yet; enabling it for macOS release artifacts is a packaging follow-up.

---
## Summary by cubic
Replaced the `apple-fm` summarizer with a native Rust FFI to Apple’s on‑device FoundationModels, removing Python and enabling the feature for Apple Silicon release builds. Per‑summary provenance is now accurate, sessions are isolated, and builds fall back to a Rust heuristic when the model isn’t available.

- **New Features**
  - Vendor `foundation-models-c` and build it via `build.rs` only when `apple-fm` is enabled on macOS; statically link `libFoundationModelsStatic.a` and set the Swift runtime rpath so the binary is self‑contained.
  - Safe FFI `summarize()` with an availability check, structured generation, and a 60s timeout; returns `None` when unavailable.
  - Ported the heuristic classifier to Rust and removed `scripts/wiki-summarizer.py`; README updated.
  - CI: Apple Silicon (`macos-26`) builds ship with `--features apple-fm`; x86_64 macOS and other platforms use the heuristic.

- **Bug Fixes**
  - Per‑summary provenance: `fm_version` is set to `"apple-fm-on-device"` only when Apple FM produced that summary.
  - Fresh session per request to avoid cross‑session contamination and stuck sessions.
  - Prevent prompt leaks by building the prompt CString before creating FM handles.

<sup>Written for commit b925b51e1ad5f0c5bfb9d209290f336c4db15a44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/721?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo:
Added release packaging (commit `a790d6b5`): the Apple Silicon binary now ships with `apple-fm`.
- **Static linking**: build.rs links `libFoundationModelsStatic.a` + a system rpath to `/usr/lib/swift`, so the published binary is **self-contained** (otool confirms only Apple's system FoundationModels framework + the OS Swift runtime — both always present on macOS 26). No dylib to ship.
- **CI**: `aarch64-apple-darwin` now builds on the GA `macos-26` runner (Xcode 26 → Swift 6.2 + macOS 26 SDK) with `--features apple-fm`. `x86_64-apple-darwin` is intentionally left without it (Apple Intelligence is Apple-Silicon-only; Intel uses the heuristic).

Validated locally on macOS 26.2/arm64: static link → self-contained binary that loads + runs. Still needs a real `macos-26` CI run + an Apple-Intelligence-enabled machine for live generation.

junhoyeo:
Addressed the Codex + cubic review in `4f315a47` (verified each against the code):
- **Provenance per-summary** (cubic P2 + Codex P2): `SessionSummary` now carries `fm_version`; set `apple-fm-on-device` only when FM produced it, `None` for heuristic (incl. per-session fallbacks). report.rs emits per-summary instead of stamping the batch — heuristic fallbacks are no longer recorded as Apple-FM.
- **Fresh session per input** (Codex P2): `respond_one` creates + releases its own `LanguageModelSession` per session (was reusing one stateful session → transcript contamination). Mirrors the old Python loop.
- **Prompt-handle leak** (cubic P2): build the prompt CString before allocating the composed-prompt handle.
- **rpath** (cubic P1): already resolved by the static-link switch (`a790d6b5`).

Also merged main, which carries an unrelated fix: `test_build_client_filter_all_legacy_flags` was red on main (its hardcoded roster missed `micode`) and was failing Code Coverage on every PR — fixed in main `6a6fd5f4`.

Verified locally (macOS 26.2/arm64): default + `--features apple-fm` builds, clippy, fmt clean; 10 apple_fm + 11 report tests pass.

PR #742, opened 2026-06-21
fix(micode): namespace dedup key by database; guard authoritative cost from repricing

### LINKED ISSUE #740: micode (MiMo Code): cross-database dedup collision + unguarded repricing of authoritative cost
The MiMo Code parser (`crates/tokscale-core/src/sessions/micode.rs`) has two correctness edge cases around dedup and cost.

## 1. Bare message id collides across databases

The dedup key is the bare `message_id` (or `row_id`):

```rust
let dedup_key = message_id.clone().unwrap_or(row_id); // micode.rs:230
```

and the lane shares one `micode_seen` set across every scanned `*.db` (`lib.rs:899`, `914`). Two independent databases under `$XDG_DATA_HOME/micode` that reuse an id (for example both starting their `row_id` sequence at `1`) collapse: the later database's unrelated message is dropped, under-counting its tokens and cost.

**Suggested fix:** namespace the dedup key by the database (or session) before it reaches the cross-file set, e.g. `format!("{db_or_session}:{id}")`.

## 2. Authoritative embedded cost is repriced unconditionally

micode carries an authoritative per-message `cost` (`micode.rs:229`), but the lane routes through `load_or_parse_sqlite_source(..., pricing, ...)` (`lib.rs:898`), which applies pricing unconditionally — `apply_pricing_if_available` overwrites whenever the lookup is non-zero:

```rust
if calculated_cost > 0.0 {
    message.cost = calculated_cost; // lib.rs:2118-2119
}
```

For native MiMo models (absent from the pricing dataset) this is a no-op, but MiMo Code preserves `providerID`/`modelID` like an opencode fork, so a session routed to a priced provider (e.g. `anthropic` / `claude-...`) has its authoritative embedded cost clobbered by a recomputed estimate. The sibling gjc lane guards this exact case with `if msg.cost <= 0.0` (`lib.rs:1218`); micode does not.

**Suggested fix:** apply the same `cost <= 0.0` guard to the micode lane (or carry a cost-present flag, as in the related gjc issue).

## Context

Found while vendoring `tokscale-core` into a downstream menu-bar usage monitor and running a high-recall review over the newer clients. Happy to send a PR.

Nanako0129:
Fixes #740.

Two correctness fixes for MiMo Code.

## 1. Cross-database dedup collision

The dedup key was the bare `message_id` (or `row_id`) and the lane shares one `micode_seen` set across every scanned `*.db`, so two independent databases under `$XDG_DATA_HOME/micode` that reuse an id (e.g. both starting their `row_id` at `1`) collapsed each other's unrelated messages. This namespaces the key by the database path (mirroring `crush`), so independent databases no longer collide. The within-database content-fingerprint dedup is unaffected (it keys on content, not the dedup key), so forked-history collapse still works.

## 2. Authoritative cost repriced unconditionally

micode carries an authoritative per-message `cost`, but the lane routed through `load_or_parse_sqlite_source(..., pricing, ...)`, which reprices unconditionally and overwrites (and persists to the source cache) the embedded cost whenever the model resolves to a price — e.g. a session routed to a priced provider (`anthropic` / `claude-...`) through MiMo Code. This passes `None` to the loader and reprices only messages with no embedded cost (`cost <= 0.0`), mirroring the gjc lane's guard.

`cargo test -p tokscale-core` (964 passed) and `cargo clippy --workspace --all-targets` are clean. Regression test asserts the same id in two databases yields distinct dedup keys.

---
## Summary by cubic
Fix cross-database dedup in MiMo Code and keep authoritative costs intact. Row-id fallback keys are namespaced per `.db`, embedded ids stay global to avoid double-counting across channel databases; pricing only runs when cost is missing.

- **Bug Fixes**
  - Namespace only the row-id fallback by database path; keep embedded message ids un-namespaced so the same message in `mimocode.db` and `mimocode-<channel>.db` collapses correctly. Added regression tests for both namespaced row-ids and shared embedded ids.
  - Stop repricing stored costs by passing `None` to `load_or_parse_sqlite_source` and applying pricing only when `cost <= 0.0`. Tests confirm costs persist across first parse and cache hits.

<sup>Written for commit 1b1a5b0fe03aa0572cd560e723e578d15f17fdf4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/742?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -899,20 +899,34 @@ fn parse_all_messages_with_pricing_with_env_strategy(
     let mut micode_seen: HashSet<String> = HashSet::new();
 
     for db_path in &scan_result.micode_dbs {
+        // Pass `None` so the loader does not reprice: MiMo Code carries an
+        // authoritative per-message cost that unconditional repricing would
+        // overwrite (and persist to the cache). Reprice only messages that had
+        // no embedded cost, mirroring the gjc lane's guard.
         let CachedParseOutcome {
             messages,
             cache_entry,
             ..
-        } = load_or_parse_sqlite_source(db_path, &source_cache, pricing, |path| {
+        } = load_or_parse_sqlite_source(db_path, &source_cache, None, |path| {
             sessions::micode::parse_micode_sqlite(path)
         });
 
-        all_messages.extend(messages.into_iter().filter(|message| {
-            message
-                .dedup_key
-                .as_ref()
-                .is_none_or(|key| micode_seen.insert(key.clone()))
-        }));
+        all_messages.extend(
+            messages
+                .into_iter()
+                .map(|mut message| {
+                    if message.cost <= 0.0 {
+                        apply_pricing_if_available(&mut message, pricing);
+                    }
+                    message
+                })
+                .filter(|message| {
+                    message
+                        .dedup_key
+                        .as_ref()
+                        .is_none_or(|key| micode_seen.insert(key.clone()))
+                }),
+        );
 
         if let Some(entry) = cache_entry {
             source_cache.insert(entry);
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -3788,6 +3788,95 @@ mod tests {
         assert!(messages[0].cost > 0.0);
     }
 
+    /// MiMo Code records carry an authoritative per-message cost. The micode
+    /// lane must NOT reprice a record that already has a cost, even when the
+    /// model has a market price that would compute a different (non-zero) value.
+    /// This must hold on the first parse AND on a subsequent cache hit, since
+    /// the previous bug repriced and persisted the inflated cost to the cache.
+    #[test]
+    #[serial_test::serial]
+    fn test_micode_authoritative_cost_is_not_repriced_on_first_parse_or_cache_hit() {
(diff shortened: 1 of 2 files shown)


junhoyeo (approved):
Thank you, @Nanako0129! 🙏 Great catch on both the authoritative-cost repricing guard and the cross-DB dedup issue.

The **cost guard is correct and I kept it as-is**. On the dedup half I pushed a maintainer touch-up: the PR namespaced the *entire* dedup key by DB path, but since MiMo writes the **same embedded message id** to both `mimocode.db` and `mimocode-<channel>.db` after a channel switch, that caused the message's cost+tokens to be **double-counted**. Fix: embedded ids stay un-namespaced (so they collapse across channel DBs), and **only the row-id fallback** is namespaced by DB path (rowids aren't globally unique). Added regression tests — verified `embedded_message_id_is_not_namespaced_by_database` *fails* on the old code and passes with the fix; 966 tests green, clippy clean.

Approving and merging — thank you for a sharp, well-scoped contribution! 🎉

Issue #739, opened 2026-06-21
gjc: three dedup / cost-attribution edge cases (explicit-zero cost, header-less collision, replay-variant key)


Nanako0129:
The gjc parser (`crates/tokscale-core/src/sessions/gjc.rs`) and its lane have three independent correctness edge cases around cost attribution and message dedup. Grouping them since they share the file; happy to split or send a PR for any one.

## 1. Explicit `usage.cost.total: 0.0` is repriced instead of preserved

The gjc lane reprices only when the embedded cost is absent, using `cost <= 0.0` as the absence proxy (`lib.rs:1218` / `1244`, and the binding-note comment at `lib.rs:1202`). But `embedded_cost` (`gjc.rs:76`) collapses "absent" and "explicit `0.0`" to the same `0.0`:

```rust
match usage.cost.as_ref().and_then(|c| c.total) {
    Some(total) if total.is_finite() && total >= 0.0 => total,
    _ => 0.0,
}
```

So a free/credited turn that genuinely recorded `usage.cost.total: 0.0` for a priced model is treated as missing and overwritten with a calculated price.

**Suggested fix:** carry whether `usage.cost.total` was present (an `Option<f64>` through to the message, or a bool) and reprice on absence rather than on `<= 0.0`.

## 2. Header-less files collide on the `"unknown"` session id

`parse_gjc_file` defaults `session_id` to `"unknown"` when a file has no `{"type":"session",...}` header (`gjc.rs:198`) and builds `dedup_key = "{session}:{msg_id}"` (`gjc.rs:199-200`). The lane shares one `gjc_seen` set across all files (`lib.rs:1210`, `2450`), so two independent header-less files that share a message id (a truncated/rotated log, or a file that begins mid-session) both produce `"unknown:<id>"` and the second is dropped — its tokens and cost vanish.

**Suggested fix:** namespace the fallback session by the file path, or skip cross-file dedup when the session id is the `"unknown"` sentinel.

## 3. Id-less fallback key is not replay-invariant

[...]

PR #743, opened 2026-06-21
fix(gjc): namespace header-less session fallback by file name

Nanako0129:
Addresses the header-less collision in #739 (one of the three items there).

A file with no `{"type":"session",...}` header defaults `session_id` to a shared `"unknown"` (`gjc.rs:198`), and the dedup key is `"{session}:{msg_id}"`. The lane shares one `gjc_seen` set across all files, so two independent header-less files (a truncated/rotated log, or a file that begins mid-session) that share a message id both produce `"unknown:<id>"` and the second is dropped. This falls back to the file name instead, so independent files get distinct sessions and dedup keys. Regression test included.

The other two items in #739 are left as-is for your call, since they touch design intent:

- **explicit `usage.cost.total: 0.0`** would need the parser to carry cost *presence* (an `Option`/flag through to the message) so the lane reprices on absence rather than `<= 0.0` — a `UnifiedMessage`-shape change affecting all clients.
- **id-less fallback `ordinal`**: `derive_dedup_key`'s doc comment says the ordinal is intentional ("distinct messages stay apart"); dropping it for replay-invariance is a tradeoff worth your call.

`cargo test -p tokscale-core` (964 passed) and `cargo clippy --workspace --all-targets` are clean.

---
## Summary by cubic
Fix header-less GJC session collisions by deriving the fallback `session_id` from the file name instead of the shared `unknown`. Independent files now get distinct sessions and dedup keys; adds tests (including header-precedence) and clarifies the header-less fallback scope in comments.

<sup>Written for commit c3a8855674cfbf8113ffd4eeba9873aaa8dbb9c3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/743?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/gjc.rs
+++ b/crates/tokscale-core/src/sessions/gjc.rs
@@ -195,7 +195,15 @@ pub fn parse_gjc_file(path: &Path) -> Vec<UnifiedMessage> {
 
         let cost = embedded_cost(&usage);
 
-        let session = session_id.clone().unwrap_or_else(|| "unknown".to_string());
+        // No `{"type":"session",...}` header in this file: fall back to the file
+        // name rather than a shared `"unknown"`, so two independent header-less
+        // files do not collide on the same session in the cross-file dedup set.
+        let session = session_id.clone().unwrap_or_else(|| {
+            path.file_stem()
+                .and_then(|stem| stem.to_str())
+                .map(str::to_string)
+                .unwrap_or_else(|| "unknown".to_string())
+        });
         let dedup_key = match entry.id.filter(|s| !s.is_empty()) {
             Some(msg_id) => format!("{session}:{msg_id}"),
             None => derive_dedup_key(&session, timestamp, &model, &tokens, ordinal),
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
Thank you, @Nanako0129! 🙏 Solid fix — namespacing the header-less fallback by file stem (instead of a shared `"unknown"`) correctly stops independent header-less files from collapsing in the cross-file dedup set, and it mirrors the `opencode.rs` sibling.

I pushed two small, behavior-preserving touch-ups: a comment documenting that the depth-1/depth-2 replay-collapse guarantee holds for *headered* files (the realistic case), and a complementary test proving the header session id wins over the file stem when a header is present. All gjc tests green, clippy clean.

Approving and merging — thanks! 🎉

PR #769, opened 2026-06-23
fix(tui): cancel pending Codex login child

astkaasa:
## Summary

- Move the Codex login worker and child-process helpers into `tui/codex_login.rs` so the TUI app only owns UI state and cancellation calls.
- Replace the `Option<Child>` slot with an explicit state containing `child` plus `cancelled`, so dismissing before `codex login` is spawned is still visible to the worker.
- Kill and wait a newly spawned login child immediately when it is registered after the slot was already cancelled.

## Root Cause

The previous slot used `None` for both “child has not been spawned yet” and “login was cancelled”. If the user dismissed the login panel before the worker registered the child, the app dropped its slot handle while the worker still owned another clone. The worker could later spawn `codex login`, store it in the slot, and wait forever from the TUI's perspective because the app no longer had a way to kill that child.

## Validation

- `cargo fmt`
- `cargo test -p tokscale-cli codex_login`
- `cargo test -p tokscale-cli dismiss_codex_login`
- `cargo test -p tokscale-cli on_tick_collects_codex_login_events`

---
## Summary by cubic
Fixes a race where dismissing the Codex login panel could leave a hidden `codex login` process running. Cancellation now reliably kills any pending or newly spawned child and surfaces a clear "cancelled" outcome to the UI.

- **Bug Fixes**
  - Use explicit state (`child` + `cancelled`) to tell "not started" vs "cancelled".
  - If cancelled before spawn, any later-registered child is killed immediately and the worker exits as cancelled.
  - Dismiss now calls `cancel_codex_login_child` to kill and wait for the process.

- **Refactors**
  - Moved Codex login worker and helpers to `tui/codex_login.rs`; `app.rs` keeps only UI state and cancellation.
  - Added small test helpers and trimmed `app.rs` substantially.

<sup>Written for commit 8679ebb604e7107fdb6fe30343dedb31ee7e9d48. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/769?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/tui/app.rs
+++ b/crates/tokscale-cli/src/tui/app.rs
@@ -13,7 +13,10 @@ use crate::ClientFilter;
 
 use ratatui::style::Color;
 
-use super::codex_login::CodexLoginOutcome;
+use super::codex_login::{
+    cancel_codex_login_child, run_codex_login_worker, CodexLoginChildSlot, CodexLoginEvent,
+    CodexLoginOutcome,
+};
 use super::data::{
     AgentUsage, DailyUsage, DataLoader, HourlyUsage, MinutelyUsage, ModelUsage, TokenBreakdown,
     UsageData,
@@ -173,12 +176,6 @@ pub enum ClickAction {
     CodexResetAccount { account_id: String },
 }
 
-#[derive(Debug, Clone)]
-enum CodexLoginEvent {
-    Output(String),
-    Finished(CodexLoginOutcome),
-}
-
 fn codex_reset_outcome_label(
     result: &crate::commands::usage::codex::RateLimitResetConsumeResult,
 ) -> String {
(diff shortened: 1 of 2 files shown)


junhoyeo (approved):
Thanks @astkaasa — clean refactor and a genuine fix. Verified the cancellation logic end-to-end: the new `CodexLoginChildState { child, cancelled }` correctly distinguishes "not yet spawned" from "cancelled", `put_codex_login_child` hands the child back to the worker for immediate kill when the slot was cancelled pre-spawn, and the new test covers exactly that race. CI green, cubic clean. 🚀

astkaasa:
> Thanks @astkaasa — clean refactor and a genuine fix. Verified the cancellation logic end-to-end: the new `CodexLoginChildState { child, cancelled }` correctly distinguishes "not yet spawned" from "cancelled", `put_codex_login_child` hands the child back to the worker for immediate kill when the slot was cancelled pre-spawn, and the new test covers exactly that race. CI green, cubic clean. 🚀

Thanks for your reply, let's rock! 🚀

About the usage tab, I only use codex so my work is focused on that. I don't know what it will look like if there are multiple other subscriptions, and some functions (reset) may be only effective for codex subscriptions. So if you or other people have the screenshot of that tab with other subscriptions, please share. Thanks again for your inspiration.

Issue #776, opened 2026-06-25
leaderboard not failing.


Brain2k-0005:
<img width="1800" height="984" alt="Image" src="https://github.com/user-attachments/assets/0208c7d4-43cd-4d44-abaa-db3baef4f286" />

`Uncaught Error: An error occurred in the Server Components render. The specific message is omitted in production builds to avoid leaking sensitive details. A digest property is included on this error instance which may provide additional details about the nature of the error.`

YoannLetacq:
I think I found the likely cause. I tested a few routes :

| Route | Result |
| --- | --- |
| `/leaderboard` (default) | fails |
| `/leaderboard?sortBy=cost` | fails |
| `/leaderboard?sortBy=time` | fails |
| `/leaderboard?period=month` | works |
| `/leaderboard?period=week` | works |
| `/leaderboard?view=groups` | works |
| `/u/<username>` | works for everyone **except one** profile |

The profile in question :
<img width="1256" height="254" alt="Image" src="https://github.com/user-attachments/assets/e5f663d3-9893-4a64-8767-a27eeeeced93" />

Clicking it (https://tokscale.ai/u/grenadeoftacoss) gives a 404. Nice username btw. 

Since the same page works for `month`/`week`/`groups` and every other profile, it's not a render bug — it's a data error on the all-time aggregation. Those failing routes all run `SUM(CAST(total_cost AS DECIMAL(12,4)))`, but `total_cost` was widened to `numeric(18,4)` in migration `0014` (packages/frontend/src/lib/db/migrations/0014_widen_cost_columns.sql). `DECIMAL(12,4)` caps at `99999999.9999`, and the cast is per-row, so a single user with `total_cost >= 1e8` throws `numeric field overflow` and kills the whole query. The 404 is the same thing:
`/api/users/<username>` 500s, and the profile page turns a non-OK response into `notFound()`. The account shows `$8,349.44K` / `8,364,335,497,909,652` tokens. The monthly slice stays under the cap (so the monthly view renders), but its all-time `total_cost` is almost certainly over `1e8`.

In short, the migration widened the columns but not the query casts — the DB accepts costs it can no longer read back.

grenadeoftacoss:
Hi. That profile belongs to yours truly.
Turns out "what if somebody cracks the integer limits" is less of a hypothetical than your SQL expected. Sorry for accidentally discovering the integer ceiling in production. Very funny, I must say.

Transient-Onlooker:
[...]

PR #780, opened 2026-06-26
fix(leaderboard): widen cost casts to match numeric(18,4) columns

YoannLetacq:
Migration 0014 widened submissions.total_cost to numeric(18,4) and daily_breakdown.cost to numeric(14,4), but the aggregation queries still cast to DECIMAL(12,4), which caps at 99999999.9999. The cast runs per row before the SUM, so a single user whose stored total_cost reached 1e8 threw "numeric field overflow" and 500'd the all-time leaderboard, the cost/time sorts, and that user's profile API (which the profile page then surfaced as a 404 via notFound()).

Widen every cost cast to its column's precision: total_cost to DECIMAL(18,4) across the leaderboard, group leaderboard, profile, and embed queries, and daily_breakdown.cost to DECIMAL(14,4) in the submit aggregation. Period (month/week) and group views were unaffected because they don't cast cost.

Refs #776

---
## Summary by cubic
Fixes numeric overflow in cost aggregations by widening casts to match column precision. Prevents 500s on all-time leaderboards, cost/time sorts, and user profile/embed for high-cost users; addresses #776.

- **Bug Fixes**
  - Cast `submissions.total_cost` to DECIMAL(18,4) across leaderboard (incl. search/global stats/user rank), group leaderboard, profile API, and embed queries/ranking.
  - Cast `daily_breakdown.cost` to DECIMAL(14,4) in the submit path for both per-row batch writes and the recomputed SUM.
  - Added a regression test that asserts all all-time cost casts use DECIMAL(width >= 18, 4) to prevent future narrowing.

<sup>Written for commit f6c4b2f0234bbd25429925d6eea172cb965608d5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/780?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


raw34:
I independently hit and fixed this in #782 before spotting this PR — closing mine as a duplicate, since this one predates it and also covers the submit write-path cast I'd scoped out. 👍

One thing worth lifting from #782: this PR's regression test pins the exact literals (`DECIMAL(18,4)` present / `DECIMAL(12,4)` absent) and only covers the all-time list query. cubic flagged that same literal-coupling on my PR — if someone later re-narrows to, say, `DECIMAL(15,4)`, the assertion wouldn't catch it. #782 has wider SQL-shape coverage (all-time list, **search**, **user-rank**, **group**, **profile**, **embed**) with a width-based assertion that fails on any precision `< 18`:

```js
const widths = sqlTexts
  .filter((t) => /CAST\([^)]*(?:total_cost|totalCost)[^)]*AS DECIMAL/.test(t))
  .flatMap((t) => [...t.matchAll(/DECIMAL\((\d+),\s*4\)/g)].map((m) => Number(m[1])));
expect(widths.length).toBeGreaterThan(0);
expect(widths.every((w) => w >= 18)).toBe(true);
```

Happy to PR those on top once this merges, or feel free to lift them from #782.

YoannLetacq:
Thanks @raw34 -- I've lifted your width-based assertion into the all-time regression test. 
Kept it scoped to the all-time path to stay lean for now -- happy to take your broader search / user-rank / group / profile / embed coverage on top once this merges, if you want to open that as a follow-up. 👍

junhoyeo (approved):
Thanks @YoannLetacq — sharp catch and a thorough fix. Migration 0014 widened `total_cost`→`numeric(18,4)` and `daily_breakdown.cost`→`numeric(14,4)`, but the aggregation casts stayed at `DECIMAL(12,4)`, so a single ≥1e8 total overflowed and 500'd the all-time leaderboard, the cost/time sorts, and the profile/embed APIs. Widening every cost cast to its column precision (incl. the submit write-path) is exactly right, and you addressed cubic's P1 promptly.

Also thanks @raw34, who independently hit this in #782 and closed it as a duplicate in favor of this one — I'll open a follow-up that lifts the width-based regression tests from #782 (covering search / user-rank / group / profile / embed) and credits you as co-author. 🚀

junhoyeo:
@raw34 done — opened #787 lifting your width-based cost-cast coverage from #782 (search / user-rank / group / profile / embed shapes) with you as co-author. Thanks for the thorough follow-through and for closing #782 as a dup in favor of @YoannLetacq's fix — both of your contributions are now in. 🙏

PR #782, opened 2026-06-26
fix(leaderboard): prevent numeric overflow in all-time cost aggregation

raw34:
## Problem

The hosted all-time leaderboard (`/leaderboard`, the default view) is currently down: it renders Next.js's server-side error boundary (`Application error: a server-side exception has occurred`, digest `250786165`). Period-filtered views (`?period=month|week|last-month`) work fine.

Reproduced against production with read-only probes:

| Request | Result |
|---|---|
| `GET /api/leaderboard?period=all` (any `sortBy`) | **500** |
| `GET /api/leaderboard?period=month` (any `sortBy`) | 200 |
| `GET /api/leaderboard/user/<name>?sortBy=cost` | **500** |
| `GET /api/leaderboard/user/<name>?sortBy=tokens` | 200 |
| body of `GET /leaderboard` (all-time) | contains digest `250786165` |

## Root cause

Every all-time query casts `submissions.total_cost` — a `decimal(18,4)` column ([`schema.ts`](../blob/main/packages/frontend/src/lib/db/schema.ts#L171)) — down to **`DECIMAL(12,4)`**, whose ceiling is `99,999,999.9999`. `total_cost` is a per-submission lifetime aggregate that grows over time; once it crosses **$100M** for any user, the per-row `CAST` throws `numeric field overflow` and the whole query 500s.

Period-scoped queries aggregate cost in JS (`Number(...)` / `.reduce`) with no SQL cast, so they were unaffected — which is exactly why only the all-time views broke.

The same narrowed cast appears in the group leaderboard, profile, and embed queries, so those surfaces 500 for an affected user too.

## Fix

Widen each affected cast to the column's own precision, `DECIMAL(18,4)`:
- `lib/leaderboard/getLeaderboard.ts` (8 sites)
- `lib/groups/getGroupLeaderboard.ts` (3 sites)
- `app/api/users/[username]/route.ts` (1 site)
- `lib/embed/getUserEmbedStats.ts` (2 sites)

`SUM` promotes numeric scale automatically, so no value the column can legally hold can overflow the per-row cast. Scale stays `4`, so results are byte-for-byte identical for normal-sized costs (no rounding/ordering/type change). No migration needed.

## Tests

Added SQL-shape regression tests asserting cost casts never narrow below the column precision, covering the all-time **list**, **search**, **user-rank**, **group**, **profile**, and **embed** paths. Each was confirmed to fail (RED) against the pre-fix `DECIMAL(12,4)` and pass (GREEN) after widening.

```
4 regression files: 25 passed
full suite: 433 passed (2 failures are pre-existing in clientRegistry.test.ts, unrelated)
eslint: clean on all changed files
```

## Out of scope (follow-up)

`app/api/submit/route.ts` still narrows the per-day `dailyBreakdown.cost` to `DECIMAL(12,4)` / `numeric(10,4)`. Those are write/recalc paths whose per-day inputs sit far below the overflow threshold, so they are not the live failure; left untouched to keep this fix focused on the read surfaces causing the outage.

---
## Summary by cubic
Fixes 500s in the all‑time leaderboard by preventing numeric overflow in cost aggregation. Casts now match the `decimal(18,4)` column across queries, restoring `/leaderboard`, group views, profiles, and embeds.

- **Bug Fixes**
  - Replaced `DECIMAL(12,4)` with `DECIMAL(18,4)` in all-time cost queries for leaderboard list/search/user-rank, group leaderboard, profile API, and embed (including `SUM` and `ORDER BY`).
  - Keeps scale at 4; results and ordering unchanged for normal values; no schema changes.
  - Added regression tests to ensure no narrowed casts across all affected paths.

<sup>Written for commit dbbedbc5b566c8c323b3056d2657a68556c737e8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/782?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/__tests__/api/usersProfile.test.ts
+++ b/packages/frontend/__tests__/api/usersProfile.test.ts
@@ -259,6 +259,50 @@ describe("GET /api/users/[username]", () => {
     expect(body.user.username).toBe("ImLunaHey");
   });
 
+  it("casts total_cost at full column precision in the profile stats query", async () => {
+    mockState.pushSelectResult([
+      {
+        id: "user-alice",
+        username: "alice",
+        displayName: "Alice",
+        avatarUrl: null,
+        createdAt: "2026-01-01T00:00:00.000Z",
+      },
+    ]);
+    mockState.pushSelectResult([
+      {
+        totalTokens: 0,
+        totalCost: 0,
+        inputTokens: 0,
+        outputTokens: 0,
+        cacheReadTokens: 0,
+        cacheCreationTokens: 0,
+        reasoningTokens: 0,
+        submissionCount: 0,
+        earliestDate: null,
+        latestDate: null,
+      },
+    ]);
+    mockState.pushSelectResult([]);
+    mockState.pushSelectResult([]);
+    mockState.pushExecuteResult([]);
+
+    await GET(
+      new Request("http://localhost:3000/api/users/alice"),
+      { params: Promise.resolve({ username: "alice" }) }
+    );
+
+    // submissions.total_cost is decimal(18,4); a narrower cast overflows for a
+    // profile whose lifetime cost has grown past the narrowed ceiling.
+    const costCastWidths = serializeSqlCalls()
+      .filter((text) => /CAST\([^)]*(?:total_cost|totalCost)[^)]*AS DECIMAL/.test(text))
+      .flatMap((text) =>
+        [...text.matchAll(/DECIMAL\((\d+),\s*4\)/g)].map((match) => Number(match[1]))
+      );
+    expect(costCastWidths.length).toBeGreaterThan(0);
+    expect(costCastWidths.every((width) => width >= 18)).toBe(true);
+  });
+
   it("rejects ambiguous case-insensitive username matches", async () => {
     mockState.pushSelectResult([
       {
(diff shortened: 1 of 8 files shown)


raw34:
Thanks @cubic-dev-ai — addressed both findings:

**#1 (embed test negative assertions coupled to `DECIMAL(12,4)`)** — fixed. The two tie-breaker assertions now match precision-agnostically (`CAST\(total_cost AS DECIMAL\(\d+,4\)\)`), so they keep guarding tie-breaker placement instead of going stale on the widened cast.

**#2 (rank `HAVING` round-trips the cost aggregate through JS `Number()`)** — valid in theory, but deferring as out of scope for this fix:

- This `Number()` round-trip in `fetchUserRank` is **pre-existing** (not introduced here); widening the cast only changes the theoretical ceiling.
- float64 represents a 4-decimal cost value exactly up to ~9×10¹⁵ ⇒ ~**$900B of single-user lifetime cost**. The outage this PR fixes occurs at **$100M** (~9000× lower). An off-by-one would additionally require two users within <1 ULP of each other at that scale.
- So it can't manifest at any realistic data scale, and folding a `::numeric`-literal rank-comparison change into this overflow hotfix would broaden the diff/risk.

Happy to follow up in a separate PR if you'd like the rank comparison hardened regardless.

raw34:
Closing as a duplicate of #780, which predates this PR, links the tracking issue #776, and additionally covers the submit write-path cast I'd scoped out. I've offered my extra regression tests (group/profile/embed coverage + width-based assertions) over on #780.

PR #799, opened 2026-07-01
feat(kiro): parse IDE execution files for accurate token estimation

leecoder:
Kiro IDE stores full response text in execution files under globalStorage/<hash>/<dir>/<file>.json rather than in the session JSON (which only has a placeholder like 'On it.'). This PR improves both input and output token estimation for IDE sessions.

## Changes

- Detect execution files by their executionId+actions structure
- **Output**: Extract from actions[say/reasoning].output.message (actual response text)
- **Input**: Read from context.messages[].entries[] (full context including system prompt + history, matching other clients)
- Use .chars().count() consistently (not .len()) for Unicode-safe char counting
- Skip Kiro-internal model names (agent, auto, qdev) and fall back to 'auto'

## Before/After

| Metric | Before | After |
|--------|--------|-------|
| IDE output tokens | 3 (from 'On it.' placeholder) | 146 (from actual reasoning+say text) |
| IDE input tokens | 1 (user message only) | ~4000 (full context with system prompt) |

### DIFF
--- a/crates/tokscale-core/src/sessions/kiro.rs
+++ b/crates/tokscale-core/src/sessions/kiro.rs
@@ -513,17 +513,17 @@ fn parse_kiro_global_storage_file(path: &Path) -> Vec<UnifiedMessage> {
         Err(_) => return Vec::new(),
     };
 
+    if let Some(messages) = try_parse_kiro_execution_file(&value, path) {
+        return messages;
+    }
+
     let file_stem = path
         .file_stem()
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
Thanks @leecoder! Merging — I'll follow up on main with workspace attribution, model extraction, duration wiring, and snapshot/execution dedup for the new execution-file path.

Issue #802, opened 2026-07-01
OMP title-first session files are dropped by Pi parser, causing recent usage to appear as 0


cuipengfei:
## Summary

Tokscale 4.0.7 advertises OMP support by scanning `~/.omp/agent/sessions` into the `pi` client, but current Pi parsing drops many recent OMP sessions because the parser requires the first non-empty JSONL record to be `{"type":"session"}`.

Current OMP session files on my machine are often `title`-first:

```json
{"type":"title", ...}
{"type":"session", ...}
```

Because of that, recent OMP sessions are ignored entirely, while older `session`-first OMP files still parse. This creates a misleading symptom:

- `tokscale --light --client pi` shows some historical data
- `tokscale --week --light --client pi` and `tokscale --month --light --client pi` show `0`
- `tokscale submit --month --client pi` submits `0` unless the files are rewritten first

## Environment

- tokscale: `4.0.7`
- install: npm/bun package `tokscale` -> `@tokscale/cli`
- platform: Linux x64 GNU
- upstream repo: `junhoyeo/tokscale`

## Why this seems to happen

In current source, `crates/tokscale-core/src/sessions/pi.rs` expects the first non-empty line to deserialize as `PiSessionHeader`, then immediately returns `Vec::new()` unless `header.entry_type == "session"`.

That is incompatible with current OMP files that prepend a `title` record before the `session` record.

### 1. Use an OMP session file whose first two non-empty lines are:

```json
{"type":"title", ...}
{"type":"session", ...}
```

### 2. Put it under a Pi-compatible root, for example:

```text
/tmp/repro/.pi/agent/sessions/testproj/orig.jsonl
```

### 3. Run:

```bash
tokscale --light --client pi --home /tmp/repro
tokscale --week --light --client pi --home /tmp/repro
tokscale --month --light --client pi --home /tmp/repro
```

## Actual behavior

All of the above report `0 messages, 0 tokens` for that raw title-first file.

## Control experiment

If I take the exact same file and remove only the leading `{"type":"title", ...}` line, leaving the rest unchanged, Tokscale immediately parses it successfully.

For one real OMP file on my machine, removing only that first `title` line changed the result from:

- `0 messages / 0 tokens`

to:

- `211 messages / 29,519,261 tokens / $16.61`

## Real-world symptom on my machine

My current OMP directory contains a mix of:

- older `session`-first OMP files from `2026-03-11`
[...]

PR #805, opened 2026-07-02
fix(core): handle Oh My Pi JSONL title line before session header

GoldTest:
Oh My Pi (OMP) session files start with a `title` line before the `session` header.

**Before:** The parser returned `Vec::new()` on any non-"`session`" line, causing OMP sessions to be silently skipped.

**After:** Skip non-"session" lines until the actual session header is found.

The scanner already adds `~/.omp/agent/sessions` as a scan root for Pi, but the parser was rejecting OMP files because of the extra title line.

**Changes:**
- `crates/tokscale-core/src/sessions/pi.rs`: Change early return to `continue` when encountering non-"session" lines in the header search phase

---
## Summary by cubic
Fix parsing of Oh My Pi JSONL files by skipping any non-"session" lines (like the leading `title`) until the actual session header. Sessions in `~/.omp/agent/sessions` are now imported instead of being silently skipped.

<sup>Written for commit 77cf795c51daf3f9c7230d848611913a8975f4ce. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/805?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/pi.rs
+++ b/crates/tokscale-core/src/sessions/pi.rs
@@ -1,7 +1,7 @@
-//! Pi (badlogic/pi-mono) session parser
+//! Pi (badlogic/pi-mono) and Oh My Pi session parser
 //!
-//! Parses JSONL files from ~/.pi/agent/sessions/<encoded-cwd>/*.jsonl
-
+//! Parses JSONL files from `~/.pi/agent/sessions/<encoded-cwd>/*.jsonl`
+//! and `~/.omp/agent/sessions/<encoded-cwd>/*.jsonl`.
 use super::utils::file_modified_timestamp_ms;
 use super::{normalize_workspace_key, workspace_label_from_key, UnifiedMessage};
 use crate::provider_identity::inferred_provider_from_model;
@@ -10,7 +10,10 @@ use serde::Deserialize;
 use std::io::{BufRead, BufReader};
 use std::path::Path;
 
-/// Pi session header (first line of JSONL)
+/// Pi/Oh My Pi session header — found in JSONL files as a line whose
+/// `type` field equals `"session"`. OMP files may precede it with a
+/// `title` line (type == "title"), so we scan for the first `"session"`
+/// entry rather than assuming it's line #1.
 #[derive(Debug, Deserialize)]
 pub struct PiSessionHeader {
     #[serde(rename = "type")]
@@ -55,7 +58,10 @@ pub struct PiUsage {
     pub total_tokens: Option<i64>,
 }
 
-/// Parse a Pi JSONL session file
+/// Parse a Pi or Oh My Pi JSONL session file.
+///
+/// OMP session files may start with a `title` line before the actual
+/// `session` header, so we skip non-`"session"` lines until we find it.
 pub fn parse_pi_file(path: &Path) -> Vec<UnifiedMessage> {
     let file = match std::fs::File::open(path) {
         Ok(f) => f,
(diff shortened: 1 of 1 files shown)


George-RD:
Links the same root cause as **#802** and **#807**.

- **#802** — issue report describing the bug (OMP title-first sessions silently dropped)
- **#805** — this PR: minimal one-line fix (`return Vec::new()` → `continue`)
- **#807** — alternative fix: type-probe approach with regression tests, more thorough

Both #805 and #807 fix the same bug. Either would resolve #802.

crates/tokscale-core/src/sessions/pi.rs
```diff
@@ -90,8 +96,10 @@ pub fn parse_pi_file(path: &Path) -> Vec<UnifiedMessage> {
                 Err(_) => return Vec::new(),
```
Mathews-Tom, line 96: This still discards the file for the real-world case in #802. `PiSessionHeader.id` is a required `String` (not `Option`), so a real OMP `title` record — e.g. `{"type":"title","v":1,"title":"...","source":"auto","updatedAt":"..."}` (no `id` field) — fails to deserialize into `PiSessionHeader` right here and hits `Err(_) => return Vec::new()`, before the `continue` on line 101 is ever reached.

Repro (as a test on this branch):
```rust
#[test]
fn test_parse_pi_real_omp_title_record_shape() {
    let content = r#"{"type":"title","v":1,"title":"Comment on GitHub issue","source":"auto","updatedAt":"2026-07-02T18:08:49.723Z"}
{"type":"session","id":"pi_ses_005","timestamp":"2026-07-02T18:07:14.690Z","cwd":"/tmp"}
{"type":"message","timestamp":"2026-07-02T18:08:53.229Z","message":{"role":"assistant","model":"claude-sonnet-5","provider":"anthropic","usage":{"input":2,"output":180,"cacheRead":0,"cacheWrite":70844,"totalTokens":71026}}}"#;
    let file = create_test_file(content);
    let messages = parse_pi_file(file.path());
    assert_eq!(messages.len(), 1); // fails: got 0
}
```

To cover this case, the pre-session line needs a probe that only requires `type` (not the full `PiSessionHeader` schema) before deciding whether to skip it — then run the full `PiSessionHeader` parse once `type == "session"` is confirmed.

junhoyeo:
Hi @GoldTest — thank you so much for jumping on this, and for being the first to diagnose and attempt a fix for the OMP title-line bug (#802)! Your PR correctly identified the root cause: the parser bailing on any non-`session` first line.

Unfortunately, after testing against the real-world record shape, this patch doesn't quite reach the bug: real OMP `title` records (e.g. `{"type":"title","v":1,"title":"...","source":"auto","updatedAt":"..."}`) have no `id` field, and `PiSessionHeader.id` is a required `String`. So the title line fails the `simd_json::from_slice::<PiSessionHeader>` parse and hits the unchanged `Err(_) => return Vec::new()` branch before ever reaching the `entry_type != "session"` check this PR changed to `continue` — the file is still dropped. This was independently confirmed by @aj2666 across 59 real OMP session files (all title-first, none with an `id` on the title record) and by @Mathews-Tom's inline comment above.

We're merging #807, which restructures the same fix to probe only the `type` field before the full header parse, plus regression tests. Closing this one as superseded — but genuinely, thank you for surfacing this and moving first on it. We'd love to see more contributions from you!

PR #814, opened 2026-07-03
feat(kiro): add workspace-session promptLogs input estimation and execution input.data parsing

leecoder:
## Summary

Improve Kiro IDE globalStorage token estimation by adding two new parsing paths:

1. **Workspace-session `promptLogs` parser** — For `workspace-sessions/` snapshot files that have `history` + `sessionId`, sum all `promptLogs[].prompt` string lengths across history entries to estimate total input tokens consumed (each prompt is the full payload sent to the LLM per turn, including cumulative context). This replaces the previous fallback to generic role-text counting which severely undercounted because assistant `message.content` is always just "On it." in these files.

2. **Execution `input.data.messages` addition** — The existing `try_parse_kiro_execution_file` only counted `context.messages` entries as input. Now also includes `input.data.messages` text (the user's current-turn input), which was previously uncounted.

3. **Executions index skip** — Files with `executions` + `version` top-level keys are now early-returned as empty (they're summary metadata, no token usage data).

## Routing order in `parse_kiro_global_storage_file`

```
execution (executionId + actions) → workspace-session (history + sessionId/selectedModel + promptLogs) → executions index (skip) → generic snapshot fallback
```

## What was tested

- `cargo test kiro` — 22 tests pass (including pre-existing execution, snapshot, dedup, and scanner tests)
- Validated against real Kiro IDE dumps from two users (t1000040, t1100403) with workspace-session files containing 44 sessions / 1,402 history entries

---
## Summary by cubic
Improves Kiro IDE globalStorage token estimates by parsing workspace-session `promptLogs`, counting execution input messages, and skipping index files. Also prevents double counting by suppressing workspace-session snapshots covered by execution records.

- **New Features**
  - Parse workspace-session: sum `history[*].promptLogs[*].prompt` for input; count assistant message text for output.
  - Include `execution.input.data.messages` in input totals.
  - Skip executions index files and route: execution → workspace-session → snapshot fallback.

- **Bug Fixes**
  - Use `chars().count()` for string lengths to handle non-ASCII consistently.
  - Suppress `{session_id}:workspace-session` snapshots when a matching execution exists (global session UUID match) to avoid double counting.

<sup>Written for commit 6ed8d015773b69e20192cc4f0894f8c63eecab64. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/814?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/kiro.rs
+++ b/crates/tokscale-core/src/sessions/kiro.rs
@@ -774,7 +774,7 @@ fn try_parse_kiro_workspace_session(
         if let Some(prompt_logs) = entry.get("promptLogs").and_then(|v| v.as_array()) {
             for pl in prompt_logs {
                 if let Some(prompt) = pl.get("prompt").and_then(|v| v.as_str()) {
-                    total_prompt_chars += prompt.len();
+                    total_prompt_chars += prompt.chars().count();
                     prompt_log_count += 1;
                 }
             }
@@ -782,7 +782,7 @@ fn try_parse_kiro_workspace_session(
         if let Some(msg) = entry.get("message") {
             if msg.get("role").and_then(|v| v.as_str()) == Some("assistant") {
                 if let Some(content) = msg.get("content").and_then(|v| v.as_str()) {
-                    assistant_chars += content.len();
+                    assistant_chars += content.chars().count();
                 }
             }
         }
--- a/crates/tokscale-core/src/sessions/kiro.rs
+++ b/crates/tokscale-core/src/sessions/kiro.rs
@@ -833,14 +833,19 @@ fn try_parse_kiro_workspace_session(
 ///
 /// Matching is exact and workspace-scoped on the shared `executionId` (with a
 /// legacy fallback matching an execution's `chatSessionId` against a snapshot
-/// file stem). Anything unmatched is kept — the pass can only remove verified
-/// duplicates, never unrelated usage.
+/// file stem). Workspace-session promptLogs snapshots are matched globally on
+/// the session UUID instead, because they live under a different
+/// `kiro.kiroagent` subdirectory than executions and so never share a
+/// workspace key. Anything unmatched is kept — the pass can only remove
+/// verified duplicates, never unrelated usage.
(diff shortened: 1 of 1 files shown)


junhoyeo:
Thanks @leecoder — this is a genuinely valuable improvement. You found the biggest blind spot in the Kiro globalStorage estimator: `promptLogs[].prompt` (the full compiled request payload per turn) was completely invisible to the old generic snapshot walker, and we confirmed against upstream Continue's schema (`PromptLog { prompt, completion, ... }`) that your interpretation of the field is exactly right. The `input.data.messages` addition and the executions-index skip both check out too.

We pushed one small addition on top of your branch before merging: workspace-session snapshots and execution records can coexist for the same conversation in real trees (execution `chatSessionId` == workspace-session `sessionId`), and `suppress_snapshots_covered_by_executions` didn't yet recognize your new `:workspace-session` dedup key — so a session captured by both stores would have had its cumulative prompt payloads counted twice. The pushed commit teaches the suppression pass to drop a workspace-session snapshot when any of its executions are already counted (matched globally on the session UUID, since the two stores live under different `kiro.kiroagent` subdirectories and never share a workspace key), plus tests for both your parser and the new suppression case.

One follow-up we noted for later (no action needed from you): `kiro_global_storage_workspace` currently attributes these files to a literal "workspace-sessions" workspace rather than the real workspace hash one level deeper — grouping cosmetics only, token totals are unaffected. Thanks again for the careful validation against real dumps!

junhoyeo:
Shipped in [v4.0.11](https://github.com/junhoyeo/tokscale/releases/tag/v4.0.11), included in the latest [v4.1.0](https://github.com/junhoyeo/tokscale/releases/tag/v4.1.0) 🎉 @leecoder

PR #819, opened 2026-07-04
fix(jcode): keep journal corrections turn-neutral so new turns still count

### LINKED ISSUE #818: jcode: journal correction consumes a pending turn-start and under-counts turn_count
`parse_jcode_file` in `crates/tokscale-core/src/sessions/jcode.rs` merges the append-only journal through the same mutable turn-state `context` as the snapshot. `parse_jcode_messages` derives `is_turn_start` from a running `pending_turn_start`:

```rust
// parse_jcode_messages, shared by the snapshot and journal passes
if message.role.as_deref() == Some("assistant") && context.pending_turn_start {
    unified.is_turn_start = true;
    context.pending_turn_start = false;
}
```

and the journal merge then overwrites that flag for a message that replaces a snapshotted id:

```rust
// parse_jcode_file, journal merge
Some(existing_index) => {
    message.is_turn_start = parsed[existing_index].is_turn_start;
    parsed[existing_index] = message;
}
```

## Root cause

A `user` message sets `pending_turn_start` and the next `assistant` message consumes it. When the snapshot ends on a user message, `pending_turn_start` is still true as the journal is merged. If the journal's first entry corrects an *earlier* assistant id (the same dedup id as a snapshotted message), parsing that duplicate through the shared context consumes `pending_turn_start` -- even though the merge immediately discards the duplicate's `is_turn_start` and restores the snapshot's flag. A subsequent brand-new journal assistant then finds `pending_turn_start == false` and is never marked `is_turn_start`, so that session's `turn_count` is under-counted by one.

## Reproduction

Snapshot ends on a user message; the journal first corrects a snapshotted id, then opens a new turn:

| Source | role | id | pending after | is_turn_start (current) | is_turn_start (expected) |
|---|---|---|---|---|---|
| snapshot | user | user_1 | true | -- | -- |
| snapshot | assistant | assistant_snap | false | true | true |
| snapshot | user (last) | user_2 | true | -- | -- |
| journal L1 | assistant | assistant_snap (dup) | **false** | discarded | turn-neutral |
| journal L2 | assistant | assistant_journal (new) | false | **false** | **true** |

The new journal assistant should open the second turn but is not marked, so `turn_count` is 1 instead of 2. Control: drop the journal's first (duplicate) line and the new assistant is correctly marked -- confirming the duplicate's consumption of `pending_turn_start` is the cause.

## Suggested fix

Make a journal message that only *replaces* an already-emitted `dedup_key` turn-neutral: it should not advance the turn-state machine, because its `is_turn_start` is overwritten by the snapshot's flag during the merge anyway. Passing the set of already-known dedup keys into the journal parse (and `None` for the snapshot pass, which stays unchanged) lets a replacement skip both setting and consuming `pending_turn_start`, so the pending turn-start survives for the next brand-new message. An equivalent alternative is to snapshot and restore `pending_turn_start` around a replace in the merge branch, but that only holds when the duplicate and the new turn are on separate journal lines; gating in the parse is robust to both. This changes cached turn flags, so it also needs a `CACHE_SCHEMA_VERSION` bump.

## Context

Found while vendoring `tokscale-core` into a downstream menu-bar usage monitor. PR attached with the fix plus a regression test that fails before the change (the new journal assistant's `is_turn_start` is false, `turn_count` 1 instead of 2); happy to adjust the approach.

Nanako0129:
When a session snapshot ends on a user message, a turn-start is pending as the append-only journal is merged. `parse_jcode_messages` runs the journal through the same mutable turn-state `context` as the snapshot, so a journal entry that merely corrects an already-snapshotted assistant id (a replace, whose `is_turn_start` is overwritten by the snapshot's flag during the merge) consumes that pending turn-start. A subsequent brand-new journal assistant is then never marked `is_turn_start`, under-counting the session's `turn_count` by one.

This makes a journal message that only replaces an already-emitted `dedup_key` turn-neutral: it no longer advances the turn-state machine (its turn flag is discarded anyway), so the pending turn-start survives for the next new message. The snapshot pass passes `None` and is unchanged.

`journal_correction_of_snapshot_message_keeps_pending_turn_start` builds a snapshot ending on a user message, a journal that first corrects the snapshot assistant and then opens a new turn, and asserts both turns are counted; it fails before the change (the new turn's `is_turn_start` is false) and passes after.

Bumps `CACHE_SCHEMA_VERSION` 24 -> 25 so caches holding the under-counted turn flags reparse. (If a sibling schema-touching fix lands first, this needs a trivial rebase to the next number.)

Fixes #818

---
## Summary by cubic
Fixes under-counted turn tracking in Jcode sessions by making journal corrections turn-neutral, so a new assistant message after a user still starts a turn. Also bumps cache schema and includes a minor rustfmt cleanup.

- **Bug Fixes**
  - Use known snapshot `dedup_key`s to detect journal replacements and keep them turn-neutral.
  - Only non-replacement assistant messages consume a pending turn-start; snapshot parsing is unchanged.
  - Adds regression test `journal_correction_of_snapshot_message_keeps_pending_turn_start`.

- **Migration**
  - Bump `CACHE_SCHEMA_VERSION` to 26 to invalidate caches carrying under-counted `is_turn_start`.

<sup>Written for commit b46b13b43f41879ff5c380aa530b3aa5927a3f4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/819?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -25,7 +25,11 @@ use std::time::UNIX_EPOCH;
 // token_usage, so invalidate them.
 // 24: UnifiedMessage now carries cost_source so cached provider-reported costs
 // are not repriced as if they were missing.
-const CACHE_SCHEMA_VERSION: u32 = 24;
+// 25: Jcode journal corrections that only replace a snapshotted message are now
+// turn-neutral, so a following brand-new journal turn is no longer robbed of
+// its is_turn_start; schema-24 caches carry the under-counted turn flags, so
+// invalidate them.
+const CACHE_SCHEMA_VERSION: u32 = 25;
 const CACHE_FILENAME: &str = "source-message-cache.bin";
 const CACHE_LOCK_FILENAME: &str = "source-message-cache.lock";
 const MAX_CACHE_FILE_BYTES: u64 = 256 * 1024 * 1024;
--- a/crates/tokscale-core/src/sessions/jcode.rs
+++ b/crates/tokscale-core/src/sessions/jcode.rs
@@ -170,8 +170,7 @@ fn parse_jcode_messages(
             // following brand-new journal message should have received (an
             // under-count of that session's turn_count). `known_dedup_keys` is
             // None for the snapshot pass, so snapshot parsing is unchanged.
-            let is_replacement = known_dedup_keys
-                .is_some_and(|keys| keys.contains_key(&dedup_key));
+            let is_replacement = known_dedup_keys.is_some_and(|keys| keys.contains_key(&dedup_key));
 
             if !is_replacement && message.role.as_deref() == Some("user") {
                 context.pending_turn_start = true;
--- a/crates/tokscale-core/src/sessions/jcode.rs
+++ b/crates/tokscale-core/src/sessions/jcode.rs
@@ -149,16 +149,34 @@ fn parse_jcode_messages(
     context: &mut JcodeSessionContext,
     fallback_timestamp: i64,
     fallback_id_scope: &str,
+    known_dedup_keys: Option<&std::collections::HashMap<String, usize>>,
 ) -> Vec<UnifiedMessage> {
     messages
         .into_iter()
         .enumerate()
         .filter_map(|(ordinal, message)| {
-            match message.role.as_deref() {
-                Some("user") => context.pending_turn_start = true,
-                Some("assistant") => {}
-                _ => {}
+            let message_id = message
+                .id
+                // Real Jcode messages include stable IDs; this fallback keeps
+                // malformed/custom files parseable without colliding across
+                // snapshot and journal batches.
+                .unwrap_or_else(|| format!("{fallback_id_scope}:{ordinal}"));
+            let dedup_key = format!("jcode:{}:{message_id}", context.session_id);
+
+            // A journal correction that only replaces an already-emitted message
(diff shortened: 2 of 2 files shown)


junhoyeo:
Thanks @Nanako0129 — this is a really well-built fix. The issue write-up in #818, the turn-neutral replacement approach, the regression test that genuinely fails pre-fix (I verified: without the fix it panics at `assertion failed: messages[1].is_turn_start`, exactly the under-count you described), and the correctly-annotated cache schema bump all made this easy to verify. The only thing standing between this and green CI was rustfmt wanting the `is_replacement` binding on one line, so I pushed that one-line formatting commit to your branch rather than bounce it back to you. Merging once CI confirms. Nice catch on a subtle interaction between the snapshot/journal merge and the turn-state machine!

junhoyeo:
Shipped in [v4.0.11](https://github.com/junhoyeo/tokscale/releases/tag/v4.0.11), included in the latest [v4.1.0](https://github.com/junhoyeo/tokscale/releases/tag/v4.1.0) 🎉 @Nanako0129

PR #821, opened 2026-07-04
fix(copilot): prefer the invoke_agent span for the trace fallback agent

### LINKED ISSUE #820: copilot: trace-fallback agent is first-wins, mis-attributing agentless turns on out-of-order OTel export
`collect_trace_contexts` in `crates/tokscale-core/src/sessions/copilot.rs` locks the trace-level fallback agent to the first non-empty `gen_ai.agent.id` it sees in a trace:

```rust
if context.agent_id.is_none() {
    if let Some(agent_id) = first_non_empty_attr(attributes, &["gen_ai.agent.id"]) {
        context.agent_id = Some(agent_id.to_string());
    }
}
```

The comment on this block states the first-wins lock assumes the `invoke_agent` span exports before the per-turn chat spans.

## Root cause

The trace-level agent is only a fallback for records that carry no `gen_ai.agent.id` of their own; per-record ids win at attribution time. But OTel file export order is not guaranteed. A child chat / sub-agent span that carries its *own* `gen_ai.agent.id` can be written before its parent `invoke_agent` span. The first-wins lock then stores the sub-agent id as the trace default, and a later agentless turn in the same trace inherits the sub-agent instead of the intended trace default (the `invoke_agent` span's agent). The parser already distinguishes the `invoke_agent` span via `is_agent_summary_span_record`, so the correct fallback source is identifiable.

## Reproduction

One trace, records exported child-before-parent:

| Order | Record | operation.name | gen_ai.agent.id | own agent | trace fallback |
|---|---|---|---|---|---|
| 1 | chat (sub-agent turn) | `chat` | `github.copilot.reviewer` | reviewer | -- |
| 2 | invoke_agent (parent) | `invoke_agent` | `github.copilot.default` | -- | -- |
| 3 | chat (plain turn) | `chat` | *(none)* | -- | **current: reviewer / expected: default** |

Record 3 carries no agent id, so it falls back to the trace agent. Current output attributes it to `github.copilot.reviewer` (the first-seen id, from record 1); it should be `github.copilot.default` (the `invoke_agent` span). Control: swapping records 1 and 2 (parent first) yields the correct `default` -- confirming export order, not anything causal, decides attribution today.

## Suggested fix

Populate the trace fallback preferring the `invoke_agent` span's agent id over a per-turn span's, rather than first-wins. Tracking whether the stored id came from an `invoke_agent` span (via `is_agent_summary_span_record`) lets a later `invoke_agent` span override an earlier chat-span id, while first-non-empty behavior is kept otherwise; per-record ids still take precedence at attribution time, so a record that names its own agent is unaffected. This changes cached agent attribution, so it also needs a `CACHE_SCHEMA_VERSION` bump.

## Context

Found while vendoring `tokscale-core` into a downstream menu-bar usage monitor with an Agents view. PR attached with the fix plus a regression test that fails before the change (the agentless turn attributes to the sub-agent that exported first); happy to adjust.

Nanako0129:
`collect_trace_contexts` locked the trace-level fallback agent to the first non-empty `gen_ai.agent.id` seen in the trace, assuming the `invoke_agent` span exports before the per-turn chat spans. OTel export order is not guaranteed, so a child chat / sub-agent span that carries its own `gen_ai.agent.id` can be exported before its parent `invoke_agent` span. The fallback then locks onto the sub-agent, and a later agentless turn in the same trace inherits the sub-agent id instead of the trace default.

This prefers the `invoke_agent` span's agent id for the trace fallback (tracked with a new `agent_from_invoke_agent` flag) and otherwise keeps the first non-empty id. Per-record agent ids still take precedence at attribution time, so a record that names its own agent is unaffected.

`test_parse_copilot_cli_trace_agent_prefers_invoke_agent_over_child_span` exports a sub-agent chat span before its parent `invoke_agent` span and asserts a later agentless turn attributes to the trace default; it fails before the change (attributes to the sub-agent) and passes after.

Bumps `CACHE_SCHEMA_VERSION` 24 -> 25 so caches holding the mis-attributed agent reparse. (If a sibling schema-touching fix lands first, this needs a trivial rebase to the next number.)

Fixes #820

---
## Summary by cubic
Fixes mis-attribution of Copilot agent IDs when OTel exports child spans before the parent. Agentless turns now inherit the correct default from the `invoke_agent` span. Fixes #820.

- **Bug Fixes**
  - Prefer the `invoke_agent` span’s `gen_ai.agent.id` for the trace fallback; otherwise keep the first non-empty value. Per-record agent IDs still take precedence.
  - Track fallback source with `agent_from_invoke_agent` to prevent sub-agent spans from hijacking the trace default.
  - Add a regression test to cover out-of-order exports.

- **Migration**
  - Bump `CACHE_SCHEMA_VERSION` from 24 to 25 to invalidate caches with mis-attributed agents; old traces will reparse automatically.

<sup>Written for commit 480cd387db24e279a12fee976a9687c034fdf5cb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/821?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -25,7 +25,11 @@ use std::time::UNIX_EPOCH;
 // token_usage, so invalidate them.
 // 24: UnifiedMessage now carries cost_source so cached provider-reported costs
 // are not repriced as if they were missing.
-const CACHE_SCHEMA_VERSION: u32 = 24;
+// 25: Copilot trace-level fallback agent now prefers the invoke_agent span's
+// agent id over the first-exported span, so agentless turns are no longer
+// mis-attributed to a sub-agent under an out-of-order OTel export; schema-24
+// caches carry the mis-attributed agent, so invalidate them.
+const CACHE_SCHEMA_VERSION: u32 = 25;
 const CACHE_FILENAME: &str = "source-message-cache.bin";
 const CACHE_LOCK_FILENAME: &str = "source-message-cache.lock";
 const MAX_CACHE_FILE_BYTES: u64 = 256 * 1024 * 1024;
--- a/crates/tokscale-core/src/sessions/copilot.rs
+++ b/crates/tokscale-core/src/sessions/copilot.rs
@@ -85,6 +85,7 @@ struct TraceContext {
     session_id: Option<String>,
     session_id_priority: SessionIdPriority,
     agent_id: Option<String>,
+    agent_from_invoke_agent: bool,
 }
 
 struct CopilotUsageCandidate {
@@ -146,6 +147,7 @@ fn collect_trace_contexts(records: &[Value]) -> HashMap<String, TraceContext> {
                 session_id: None,
                 session_id_priority: SessionIdPriority::Missing,
                 agent_id: None,
+                agent_from_invoke_agent: false,
             });
 
         if context.model.is_none() {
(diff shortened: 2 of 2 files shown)


junhoyeo:
Shipped in [v4.0.11](https://github.com/junhoyeo/tokscale/releases/tag/v4.0.11) and included in the latest [v4.1.0](https://github.com/junhoyeo/tokscale/releases/tag/v4.1.0) 🎉 Thanks for the contribution, @Nanako0129!

PR #825, opened 2026-07-05
fix(zcode): subtract cache/reasoning overlap from input/output

Ankairis:
## Problem

ZCode's `model_usage` SQLite table (and the JSONL transcript `usage`/`token_usage` blocks) report `input_tokens` and `output_tokens` as **inclusive** of cache and reasoning tokens, not disjoint from them:

- `input_tokens` already contains `cache_read_input_tokens` + `cache_creation_input_tokens`
- `output_tokens` already contains `reasoning_tokens`

`parse_zcode_sqlite()` and `ZcodeUsage::to_breakdown()` pass these columns straight through into `TokenBreakdown`, and `TokenBreakdown::total()` sums all five buckets as if they were disjoint — so cache and reasoning tokens get double-counted.

## Evidence

Sampled the full `model_usage` table from a real `~/.zcode/cli/db/db.sqlite` (1300+ rows, every provider/model combination — Claude, GLM, GPT, DeepSeek, Qwen — all via the ZCode client):

- `computed_total_tokens == input_tokens + output_tokens` holds with **zero exceptions** across the entire table.
- `output_tokens >= reasoning_tokens` always holds (reasoning is nested inside output, never additional).
- `input_tokens >= cache_read_input_tokens + cache_creation_input_tokens` always holds (both cache buckets are nested inside input).

Sample rows (GLM-5.2 via ZCode):

| input_tokens | cache_read_input_tokens | input − cache_read | output_tokens |
|---|---|---|---|
| 140159 | 139584 | 575 | 139 |
| 139524 | 139264 | 260 | 96 |
| 139252 | 138944 | 308 | 227 |

Sample rows (Claude Sonnet 5 via ZCode) — a row with both cache and reasoning double-counting present:

```
input_tokens=15211, output_tokens=2295, reasoning_tokens=2094,
cache_read_input_tokens=14336, cache_creation_input_tokens=0,
computed_total_tokens=17506

naive total() = 15211+2295+2094+14336+0 = 33936   (94% inflated)
correct total = input+output              = 17506
```

The same Claude model routed through the **`claude` client** (Claude Code's own session files, not ZCode) shows the opposite, cache-exclusive shape (`input` far smaller than `cache_read`), confirming this is a ZCode-client-specific data format issue, not a provider/model quirk.

## Fix

Following the existing pattern in `gemini.rs` (`normalize_gemini_session_input_and_cache` / `subtract_cached_overlap`):

- Added `normalize_zcode_input_and_output()`: when a reported total is available, compares it against the inclusive sum (`input + output`) vs. the fully-additive sum to detect the shape, then subtracts the cache/reasoning overlap out of `input`/`output` accordingly.
- `parse_zcode_sqlite()` now selects `computed_total_tokens` and uses it for detection. For the legacy-schema fallback query (no `computed_total_tokens` column), subtracts unconditionally — every sampled row in a real database is confirmed inclusive.
- `ZcodeUsage::to_breakdown()` (JSONL path) gets the same detection, gated on an optional `total`/`totalTokens` field. When absent, behavior is unchanged, since I don't have confirmed evidence about the shape of JSONL usage blocks that omit a total (see note below).

## Scope note: JSONL path directory/field mismatch

While verifying this, I found that `parse_zcode_file()`'s assumptions (path `~/.zcode/projects/<slug>/<session>.jsonl`, snake_case `usage` fields) don't match a real, current ZCode installation, which stores transcripts at `~/.zcode/cli/agents/<sess>/<agent>/transcript.jsonl` with camelCase fields (`inputTokens`, `outputTokens`, `totalTokens`, `cacheReadTokens`, `reasoningTokens`). I added the camelCase aliases and the `total`/`totalTokens` field to `ZcodeUsage` so the detection now also works against real transcript data, but did not change the directory-scanning logic — that's a separate, pre-existing bug outside this PR's scope, and I don't have enough certainty about which ZCode versions use which layout to safely change the file-discovery code here.

## Testing

- Added `test_parse_zcode_sqlite_cache_inclusive_normalization` and `test_parse_zcode_sqlite_cache_exclusive_preserved` (non-regression) to `zcode.rs`.
- Updated `test_parse_zcode_sqlite_model_usage`'s fixture to include `computed_total_tokens` and assert the normalized (not raw) input/output values.
- `cargo test --package tokscale-core sessions::zcode` — 13/13 pass.
- `cargo test --package tokscale-core` — no new failures vs. `main` (pre-existing 35 failures are unrelated path/env-var isolation issues on Windows, confirmed identical on `main` before this change).
- `cargo fmt` and `cargo clippy --package tokscale-core --lib` — clean.

## Related

[`Javis603/token-monitor` PR #68](https://github.com/Javis603/token-monitor/pull/68) hit the same cache-inclusive issue downstream. If this lands, that patch could likely be simplified or removed.

---
## Summary by cubic
Fixes double-counted ZCode tokens by subtracting cache and reasoning overlap from input/output so totals reflect real usage. Adds shape detection using reported totals and supports camelCase JSONL fields.

- **Bug Fixes**
  - Detect inclusive shape via computed totals; subtract `cache_read + cache_write` from input and `reasoning` from output.
  - `parse_zcode_sqlite()` reads `computed_total_tokens` and normalizes; probes the column directly to detect legacy schema and only subtracts unconditionally for true legacy DBs. Modern rows with NULL totals pass through unchanged.
  - `ZcodeUsage` adds camelCase aliases (`inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheCreationTokens`, `reasoningTokens`, `totalTokens`) and applies detection only when a total is present.
  - Tests cover inclusive normalization, exclusive preservation, legacy unconditional subtraction, and modern NULL-total pass-through; fixtures updated.

<sup>Written for commit daf12af8d3055a14c882367ee0442b4fbd6a8f0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/825?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/zcode.rs
+++ b/crates/tokscale-core/src/sessions/zcode.rs
@@ -350,12 +350,12 @@ pub fn parse_zcode_sqlite(db_path: &Path) -> Vec<UnifiedMessage> {
         ORDER BY COALESCE(mu.completed_at, mu.started_at, 0), mu.id
     "#;
 
-    let mut stmt = match conn
-        .prepare(modern_query)
-        .or_else(|_| conn.prepare(legacy_query))
-    {
-        Ok(stmt) => stmt,
-        Err(_) => return Vec::new(),
+    let (mut stmt, is_legacy_schema) = match conn.prepare(modern_query) {
+        Ok(stmt) => (stmt, false),
+        Err(_) => match conn.prepare(legacy_query) {
+            Ok(stmt) => (stmt, true),
+            Err(_) => return Vec::new(),
+        },
     };
 
     let rows = match stmt.query_map([], |row| {
@@ -418,18 +418,28 @@ pub fn parse_zcode_sqlite(db_path: &Path) -> Vec<UnifiedMessage> {
                 raw_reasoning,
                 Some(total),
             ),
-            // `computed_total_tokens` is absent only on pre-migration rows
-            // read through the legacy-schema fallback query. Every sampled
-            // row in a real ZCode database (all providers/models) is
-            // cache/reasoning-inclusive — see the fix PR description for the
-            // sampled evidence — so subtract unconditionally rather than
-            // passing the raw, overlap-laden columns straight through.
-            None => (
+            // When `computed_total_tokens` is NULL, distinguish two cases:
+            // 1. Legacy schema (column doesn't exist): unconditionally subtract,
+            //    since every sampled row in a real ZCode database is confirmed
+            //    cache/reasoning-inclusive.
+            // 2. Modern schema but this row's value is NULL: can't detect shape,
+            //    so pass through unchanged (the normalize function's default when
+            //    total is None). Subtracting unconditionally here would undercount
+            //    rows that are already cache-exclusive.
+            None if is_legacy_schema => (
                 subtract_overlap(
                     raw_input,
                     raw_cache_read.max(0).saturating_add(raw_cache_write.max(0)),
                 ),
                 subtract_overlap(raw_output, raw_reasoning),
+            ),
+            None => normalize_zcode_input_and_output(
+                raw_input,
+                raw_output,
+                raw_cache_read,
+                raw_cache_write,
+                raw_reasoning,
+                None,
             ),
         };
 
--- a/crates/tokscale-core/src/sessions/zcode.rs
+++ b/crates/tokscale-core/src/sessions/zcode.rs
@@ -43,36 +43,59 @@ struct ZcodeEntry {
 /// Token usage block — field names follow the Z.ai / GLM API convention.
 #[derive(Debug, Deserialize)]
 struct ZcodeUsage {
-    #[serde(alias = "input_tokens", alias = "prompt_tokens")]
+    #[serde(alias = "input_tokens", alias = "prompt_tokens", alias = "inputTokens")]
     input: Option<i64>,
-    #[serde(alias = "output_tokens", alias = "completion_tokens")]
+    #[serde(
+        alias = "output_tokens",
+        alias = "completion_tokens",
+        alias = "outputTokens"
+    )]
     output: Option<i64>,
-    #[serde(alias = "input_cache_read", alias = "cache_read_tokens")]
+    #[serde(
+        alias = "input_cache_read",
+        alias = "cache_read_tokens",
+        alias = "cacheReadTokens"
+    )]
     cache_read: Option<i64>,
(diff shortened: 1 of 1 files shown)


Ankairis:
Fixed in 32e73075. Now tracking whether the legacy query was actually used (via which `conn.prepare()` succeeded) rather than conflating it with NULL values under the modern schema. Rows with NULL `computed_total_tokens` in the modern schema now pass through unchanged instead of being unconditionally subtracted.

junhoyeo:
Thank you for this PR — for a first contribution this is exceptional work. The evidence-driven writeup (sampling 1300+ real rows across providers, cross-checking the same model via the claude client to isolate this as a ZCode-client data-shape issue, and following the existing gemini.rs normalization pattern) made this a pleasure to review, and your quick, correct response to the NULL-vs-legacy-schema review comment in 32e7307 was exactly right. We independently corroborated your findings against several third-party ZCode parsers (burnly, codeburn, vibe-usage), and they all agree with your inclusive-shape analysis, so the core fix lands as-is. We pushed a few small additions to your branch rather than asking for another round: two tests covering the branches your second commit added (a true legacy-schema table without the computed_total_tokens column, and a modern-schema row with NULL total asserting pass-through), plus a small hardening so is_legacy_schema is detected by probing the column directly instead of relying on the full modern SELECT (with its session JOIN) failing to prepare — that way an unrelated schema quirk can't misroute modern rows into unconditional subtraction. Your scope call on the JSONL directory/field mismatch was the right one; we'll track that discovery bug separately. Welcome aboard, and thanks again — fixes like this are exactly what keeps tokscale's accounting trustworthy.

junhoyeo:
Shipped in [v4.0.11](https://github.com/junhoyeo/tokscale/releases/tag/v4.0.11), included in the latest [v4.1.0](https://github.com/junhoyeo/tokscale/releases/tag/v4.1.0) 🎉 @Ankairis

Ankairis:
Yo thanks a ton for the awesome review! Super stoked that the fix held up against the cross-checks. And big thanks for adding those extra tests on my branch — I actually learned a thing or two from those 🔥 Already got my eye on a few other spots to poke around, so see you around the repo! 🚀🎉

PR #812, opened 2026-07-03
feat(tui): add monthly usage tab with daily drill-down

asf0:
Adds a new **Monthly** tab to the TUI that mirrors the Daily Usage table columns and lets users drill down into a month’s daily breakdown.

## What's Changed
- Added `MonthlyUsage` data model and `aggregate_monthly_from_daily()` helper.
- Added `Tab::Monthly` with full Daily-style columns: Month/Date, Turn, Msgs, Input, Output, Cache R, Cache W, Cache×, Total, Cost, Cost/1M.
- Added Enter/Esc keybindings to open and close a month’s daily breakdown inline.
- Updated footer count labels and help text for summary and detail modes.
- Updated cache refresh and data export to include monthly aggregation.
- Updated `tokscale monthly` to launch the TUI on the Monthly tab.
- Added unit tests for aggregation, sorting, detail filtering, keybindings, and full-column rendering.
## Screenshots
image
image
## Verification
- `cargo check -p tokscale-cli` passes.
- `cargo fmt -p tokscale-cli -- --check` passes.
- `cargo test -p tokscale-cli` passes except for one unrelated date-sensitive pre-existing test (`usage_reset_button_renders_when_credit_available`).

---
## Summary by cubic
Adds a Monthly tab to the TUI with inline daily drill-down, plus a fix to restore the monthly viewport after closing details and a small performance improvement in monthly detail filtering.

- **New Features**
  - Added `MonthlyUsage` and `aggregate_monthly_from_daily()`; monthly data computed on cache load/refresh and sortable by Date/Cost/Tokens (defaults to Date ↓).
  - New `Monthly` tab with Daily-style columns (adaptive on narrow screens) and Enter-to-open/Esc-to-close daily breakdown.
  - Tab cycle updated to include Monthly; `tokscale monthly` starts on the Monthly tab.
  - Footer counts and help text reflect summary/detail modes; export JSON now includes a `monthly` array.
  - Detail view exits if a month disappears after refresh and restores selection/scroll on close; tests cover aggregation, sorting, drill-down, keybindings, and rendering.

- **Performance**
  - Reduced allocations when filtering daily rows in the Monthly detail view.

<sup>Written for commit 2011b7fdd7cd7ced26fe6f13b30f2bd97f0fcbf5. Summary will update on new commits.</sup>

[...]

PR #830, opened 2026-07-05
test(tui): make usage_reset_button_renders_when_credit_available deterministic

### LINKED ISSUE #828: test(tui): make usage_reset_button_renders_when_credit_available deterministic
`tui::ui::usage::tests::usage_reset_button_renders_when_credit_available` is date-sensitive: it asserts a relative-date rendering (e.g. `expires Jul 12`), so whether it passes depends on the wall-clock date when the test runs. It currently fails on `main` on some dates and has been intermittently failing the Code Coverage / Test & Coverage checks on unrelated PRs (observed during reviews of #807, #812, and #825), forcing maintainers to mentally filter out a known-flaky failure.

Fix: make the test deterministic by either injecting a fixed clock into the code under test, or computing the expected string in the test from the same date math the widget uses, so the assertion holds on any date.

Mathews-Tom:
## Summary

Fixes #828. `tui::ui::usage::tests::usage_reset_button_renders_when_credit_available` hardcoded a fixed credit expiry (`2026-07-12T01:31:33Z`) and asserted a hardcoded rendered substring (`expires Jul 12`). `helpers::format_reset_time` derives that string from `Utc::now()` at test-run time, so the assertion silently depended on wall-clock time and the local/CI timezone.

## Root cause

Three independent ways this breaks, confirmed by simulating `format_reset_time`'s branch logic against the hardcoded date:

- Any local/CI timezone offset west of UTC by more than ~1.5h (US Pacific, US Eastern, etc.) renders the previous calendar day — `01:31 UTC` rolls back to `Jul 11` under a negative offset, since `chrono::Local` reads the OS/runner timezone. This fails on *any* date, not just near the deadline, which likely explains it showing up during review of #807, #812, and #825.
- Within 24h of the hardcoded expiry (from `2026-07-11T01:31:33Z` onward) it renders a relative countdown (`resets in Xh Ym`) instead of an absolute date.
- From the expiry onward — forever — it renders `resets now`, a permanent, unconditional failure once the calendar passes `2026-07-12T01:31:33Z`.

## Fix

- Added a `far_future_reset_timestamp()` test helper returning `(Utc::now() + Duration::days(10)).to_rfc3339()`, comfortably outside the 24h-relative window regardless of when the test runs.
- Threaded it through `output_with_reset_credits` as an explicit `expires_at: &str` parameter instead of a hardcoded literal, updating all 4 call sites.
- The flaky test now derives its expected substring via `helpers::format_reset_time(&expires_at).replace("resets", "expires")` — the same production code path the widget itself uses — instead of asserting against a hardcoded literal.

This mirrors the deterministic-test idiom already used elsewhere in the same file: `format_reset_time_with_now`'s own unit tests in `helpers.rs` call it with fixed instants for both `reset_at` and `now`, and the other reset-credit tests in `usage.rs` use unparseable placeholder strings like `"reset-a"` to sidestep date math via the `Err` fallback in `format_reset_time`.

## Testing

- `cargo test -p tokscale-cli` — full `tui::ui::usage::tests` module (40 tests) passes, including the fixed test.
- `TZ=America/Los_Angeles cargo test -p tokscale-cli usage_reset_button_renders_when_credit_available` — passes under the exact negative-UTC-offset condition that a Python simulation confirmed breaks the old hardcoded-date assertion (direct regression proof, not just logic review).
- `cargo fmt --all -- --check` and `cargo clippy --locked --workspace --all-features -- -D warnings` — clean.
- `cargo test --workspace --all-features` — 844+122+1091+10+4+3+3+4 passed, 0 failed.

## Note

I don't have write access to this repo, so I can't verify CI directly on this PR — happy to address any feedback.

---
## Summary by cubic
Make the `usage_reset_button_renders_when_credit_available` test deterministic by generating a future expiry and deriving the expected text from the same formatter used in the UI. This removes timezone and date-based flakiness.

- **Bug Fixes**
  - Added `far_future_reset_timestamp()` that returns `(Utc::now() + Duration::days(10)).to_rfc3339()` to force the absolute-date branch.
  - Passed `expires_at` into `output_with_reset_credits` and updated the call sites.
  - Derived the expected substring via `helpers::format_reset_time(expires_at).replace("resets", "expires")` instead of a hardcoded literal.
  - Test now passes regardless of local timezone and after the original date.

<sup>Written for commit 5d463607d55a607325463aed7c7fc78ea0b9880e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/830?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/tui/ui/usage.rs
+++ b/crates/tokscale-cli/src/tui/ui/usage.rs
@@ -2501,6 +2501,7 @@ mod tests {
     };
     use crate::tui::app::{Tab, TuiConfig};
     use crate::tui::data::UsageData;
+    use chrono::{Duration, Utc};
     use crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
     use ratatui::{backend::TestBackend, Terminal};
 
@@ -2534,10 +2535,22 @@ mod tests {
         output
     }
 
+    /// An RFC3339 instant far enough ahead of "now" that
+    /// `helpers::format_reset_time` always takes the absolute-date branch
+    /// (`resets <month> <day> <time>`), independent of wall-clock date or
+    /// local timezone. Using a hardcoded date here previously caused
+    /// `usage_reset_button_renders_when_credit_available` to fail whenever
+    /// tests ran within 24h of, on/after, or under a timezone that shifted,
+    /// the hardcoded expiry — see #828.
+    fn far_future_reset_timestamp() -> String {
+        (Utc::now() + Duration::days(10)).to_rfc3339()
+    }
+
     fn output_with_reset_credits(
         provider: &str,
         account: Option<UsageAccount>,
         available_count: u32,
+        expires_at: &str,
     ) -> UsageOutput {
         let mut output = output(provider, account);
         output.reset_credits = Some(UsageResetCredits {
@@ -2546,7 +2559,7 @@ mod tests {
                 id: Some("credit_1".to_string()),
                 status: Some("available".to_string()),
                 reset_type: Some("codex_rate_limits".to_string()),
-                expires_at: Some("2026-07-12T01:31:33Z".to_string()),
+                expires_at: Some(expires_at.to_string()),
                 title: Some("One free rate limit reset".to_string()),
                 description: None,
             }],
@@ -2670,6 +2683,7 @@ mod tests {
                 is_active: true,
             }),
             1,
+            &far_future_reset_timestamp(),
         );
         output.metrics.clear();
         app.subscription_usage = vec![output];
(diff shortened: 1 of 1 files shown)


junhoyeo:
Thank you @Mathews-Tom — second one merged today! This was exactly the fix #828 called for: the expected string now flows through the same `format_reset_time` path as the render (verified it mirrors the production transform at usage.rs:2459), so the assertion can never drift from the wall clock again. This flake was tripping Code Coverage on unrelated PRs, so this helps every contributor.

junhoyeo:
Shipped in [v4.0.11](https://github.com/junhoyeo/tokscale/releases/tag/v4.0.11), included in the latest [v4.1.0](https://github.com/junhoyeo/tokscale/releases/tag/v4.1.0) 🎉 @Mathews-Tom

PR #563, opened 2026-05-17
feat(models): add token timing and Claude Code stream dedupe

PekingSpades:
## Motivation

Tokscale already reports token volume and cost by model/provider, but those numbers do not show how fast each model actually responds. Comparing raw request duration is also misleading because requests vary a lot in token count.

This PR adds a normalized performance metric, `ms per 1K tokens`, so users can compare different providers and models on a common scale using the local usage data Tokscale already parses.

## What changed

- Adds `ModelPerformance` to model usage reports with:
  - `msPer1KTokens`
  - `totalDurationMs`
  - `timedTokens`
  - `sampleCount`
  - `tokenCoverage`
- Carries per-request `duration_ms` through unified parsed messages.
- Extracts request duration from supported local data sources:
  - Claude Code: user request timestamp to assistant usage timestamp, with streaming duplicate merge handling.
  - Codex: `turn_context` timestamp to `token_count` timestamp.
  - Copilot: chat span `startTime`/`endTime` or duration fields from telemetry records.
  - Kiro: prompt/request start timestamp to end/stream-end timestamp.
  - OpenCode: `time.completed - time.created`.
- Shows `ms/1K` in the `models` CLI table and includes `performance` in `models --json`.
- Persists performance data in TUI model cache data and bumps the cache schema so older caches refresh.

## Calculation

For each parsed message, Tokscale only records performance when both values are available and positive:

```text
messageTokens = input + output + cacheRead + cacheWrite + reasoning
durationMs = requestEndTime - requestStartTime
```

For each grouped model/provider/client entry:

```text
timedTokens += messageTokens
totalDurationMs += durationMs
sampleCount += 1
```

After aggregation:

```text
msPer1KTokens = totalDurationMs * 1000 / timedTokens
tokenCoverage = timedTokens / totalTokens
```

`tokenCoverage` indicates how much of the model's total token volume had usable timing data. If a model has tokens but no reliable duration samples, `msPer1KTokens` is `null` in JSON and rendered as an empty placeholder in tables.

Totals are computed from summed durations and summed timed tokens, not by averaging already-rounded per-model values. This keeps larger samples weighted correctly.

## Notes and limitations

- This measures observed local request wall time from client logs, not provider-side model inference latency.
- Missing or non-positive duration samples are ignored rather than treated as zero.
- The metric is best-effort because each client records timing data differently.
- Generated image/report rendering is intentionally unchanged in this PR.

## Testing

- `cargo test -p tokscale-core performance`
[...]

PR #850, opened 2026-07-09
feat(models): fold configurable model aliases into one canonical model

PekingSpades:
## Motivation

One physical model often arrives under several different name-strings depending on the supply channel — for example `claude-opus-4-8`, `claude-opus-4-8-cc` (a Claude Code channel), and `anthropic/claude-opus-4-8` (a namespaced provider) are all the same model. Because usage is grouped by the raw model-name string, that one model splits into multiple rows and its tokens, cost, and message counts are scattered across them. This adds an opt-in way to fold those variants into a single canonical model.

## What changed

A new `modelAliases` object in `settings.json` maps any alias to a canonical model name:

```json
{
  "modelAliases": {
    "claude-opus-4-8-cc": "claude-opus-4-8",
    "anthropic/claude-opus-4-8": "claude-opus-4-8"
  }
}
```

The fold is applied as the terminal step of `normalize_model_for_grouping`, the single function every grouping path already routes through, so it applies uniformly to the models report (table + `--json`), every `--group-by`, the monthly and hourly reports, the graph, per-model performance, and the TUI — in one place. The map is loaded once at startup into a process-global resolver, mirroring the existing `PRICING_SERVICE` pattern, and honors the global `--home` override exactly like scanner settings.

Keys and values are matched in the normalized model-name space (case-insensitive, dated-suffix- and `.`-vs-`-`-insensitive), resolution is single-hop, and empty/self-map entries are dropped. When no aliases are configured the behavior is byte-for-byte identical to today.

## Design notes

The fold runs strictly at grouping time, after per-message pricing has already been computed on the raw model id (`calculate_cost_with_provider`), and aggregation only sums the pre-computed cost — so folding can only relabel and merge already-costed buckets and can never change a cost total. It is independent of the pricing alias table (`pricing::aliases`) and orthogonal to `provider_identity`: it touches only the model dimension, and the existing provider-merge logic unions the providers of a folded model for free. The TUI light-cache schema version is bumped (9 → 10) because the normalized model string is persisted there, so a pre-feature cache refreshes once on upgrade.

## Scope and non-goals

Config-only: there is no built-in alias table, so nothing folds unless the user configures it, and distinct model families that happen to share a prefix (e.g. `gpt-5.1-codex` vs `gpt-5.2-codex`) are never merged by default. No new CLI flag, mirroring how provider canonicalization and scanner settings are config-only. No new dependencies.

The wiki `report` command intentionally continues to group on the raw `model_id` and is not folded: its entries are persisted append-only, so folding would leave a mix of raw and canonical names across sessions recorded before vs after aliases were configured. A `NOTE` in `commands/report.rs` records this and the one-line follow-up if it is ever wanted.

## Testing

- `cargo test -p tokscale-core model_alias` — resolver unit tests (folding, case/dotted-insensitive keys, single-hop, cap, lossy deserialize, flat-map round-trip).
- `cargo test -p tokscale-cli --test cli_tests alias` — end-to-end: variants fold into one row on `models --group-by model` and on `monthly`, an absent config is a no-op, and grand totals are unchanged by folding.
- `cargo test -p tokscale-cli --bin tokscale tui::settings` — old `settings.json` without `modelAliases` and a malformed `modelAliases` both load without wiping other settings.

---

cc @junhoyeo — same one-chokepoint approach as #563; would appreciate a review when you have a moment.

---
## Summary by cubic
Adds config-driven model alias folding so model variants merge into one canonical name in local reports and the TUI, without changing totals. Folding is presentation-only; submit/upload/export payloads keep raw model ids.

- New Features
  - New `settings.json` key `modelAliases` for `{ "alias": "canonical" }` mappings (case/date-insensitive; dot/dash-insensitive across providers).
  - Folding runs in `normalize_model_for_grouping` for local grouping (models, group-bys, monthly/hourly, TUI). Submit/export paths use `canonical_model_id` and never fold.
  - Single-hop resolution; empty/self maps dropped; cap of 4096 entries. Loaded once at startup, respects `--home`. Independent of pricing aliases and does not affect costs.

- Migration
  - Optional: add to `settings.json`, e.g. `"modelAliases": {"claude-opus-4-8-cc":"claude-opus-4-8","anthropic/claude-opus-4-8":"claude-opus-4-8"}`.
  - TUI cache schema bumped to 10; it refreshes once after upgrade.
  - The wiki `report` and the `graph`/submit payload continue to use raw `model_id` (no folding).

<sup>Written for commit ce16814a6b7741c1e5de64f8b604ec9545490e59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/850?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/report.rs
+++ b/crates/tokscale-cli/src/commands/report.rs
@@ -146,6 +146,15 @@ fn populate_wiki_from_sessions(db: &WikiDb, opts: &ReportOptions) -> Result<()> 
         agg.total_output = agg.total_output.saturating_add(msg.output);
         agg.total_cache_read = agg.total_cache_read.saturating_add(msg.cache_read);
         agg.total_cost += compute_msg_cost(msg, pricing.as_deref());
+        // NOTE: the wiki `report` view intentionally groups on the raw model_id
+        // and does not apply `modelAliases` folding (nor the grouping
+        // normalization every other report uses). Wiki entries are persisted
+        // append-only — previously-recorded sessions are not rewritten (see the
+        // `existing.contains` skip below) — so folding here would leave a mix of
+        // raw and canonical names across sessions recorded before vs after
+        // aliases were configured. To fold in a future change, wrap the key with
+        // `tokscale_core::normalize_model_for_grouping(&msg.model_id)` here and in
+        // the by-model/daily/session/JSON surfaces.
         *agg.models.entry(msg.model_id.clone()).or_insert(0) += 1;
         agg.message_count += msg.message_count;
     }
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -507,6 +507,11 @@ fn main() -> Result<()> {
     use std::io::IsTerminal;
 
     let cli = Cli::parse();
+    // Install user-configured model aliases once, before any report/graph/TUI
+    // path runs, so model-name variants fold consistently across every command.
+    // Honors the global `--home` override exactly like scanner settings; an
+    // empty or absent config is a strict no-op.
+    tokscale_core::model_alias::set_global(&tui::settings::load_model_aliases_for_home(&cli.home));
     let can_use_tui = std::io::stdin().is_terminal() && std::io::stdout().is_terminal();
 
     if cli.test_data {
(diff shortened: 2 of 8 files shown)


junhoyeo:
@PekingSpades Thanks a lot! Please check that cubic review, once thats good I'll merge

PR #860, opened 2026-07-12
feat(leaderboard): drop estimated time as a public metric

junhoyeo:
Stacked on #859 (base: `feat/compact-groups-leaderboard`). Continues the "leaderboard only shows numbers that are actually comparable" theme.

## Why

Estimated active time is not reliable as human coding time. Parallel agents, overlapping sessions, tool/runtime durations, device overlap, and client-specific parsing all inflate it far past wall-clock — a current-month leader showed **~19,163 hours** across a period with only **288 wall-clock hours**. A methodology tooltip under a number like that just advertises that we publish data we don't trust, so this removes it from every comparative surface instead.

## What changed (display + API only)

- **Sort dimension** — dropped `"time"` from `SortBy` / `LeaderboardSortBy` / `VALID_SORT_BY`. A stale `?sortBy=time` deep link now falls back to `tokens` rather than ranking on time.
- **Leaderboard payloads** — removed `totalActiveTimeMs` from `LeaderboardUser` and `stats`, the SQL aggregates in `getLeaderboard.ts` / `getGroupLeaderboard.ts`, the Time sort control, the desktop "Time" column, the mobile row's active-time line, and the metric strip's "Active time" tile.
- **Public profile** — removed the "Active time" insight from the profile Usage details panel and stripped `totalActiveTimeMs` from the `api/users/[username]` payload.

## What was deliberately kept

- **DB columns** (`total_active_time_ms`, `active_time_ms`, `session_count`, …) and **submit-path validation** — CLIs in the wild still POST these fields and must keep succeeding, and dropping columns would be a destructive migration on immutable-after-apply tables.
- **`/local` self-inspection view** — Active Time there is your own machine's data, not a cross-user comparison, so it stays. `GraphContainer`/`StatsPanel` retain the (now display-only) prop for that surface.
- **`sessionCount`** on profiles — a count, not a duration, so it's honest under the parallel-agent reality.

## Verification

- `vitest run` — **565/565** pass (leaderboard/profile/group test expectations updated for the leaner shape)
- `tsc --noEmit` — clean
- `eslint` — 0 errors (6 pre-existing warnings in untouched files)

## Risk

Narrowing the public leaderboard JSON could affect undocumented third-party consumers reading the removed `totalActiveTimeMs` field. Called out in the commit's `Not-tested` trailer.

## Follow-up idea (not in this PR)

If a third ranking dimension is wanted later, period-scoped **Active days** (from `daily_breakdown`) is bounded and immune to parallel-agent inflation — a much safer competition axis than time.

### DIFF
--- a/packages/frontend/__tests__/api/leaderboard.test.ts
+++ b/packages/frontend/__tests__/api/leaderboard.test.ts
@@ -31,7 +31,6 @@ describe("GET /api/leaderboard", () => {
           avatarUrl: null,
           totalTokens: 1200,
           totalCost: 12.5,
-          totalActiveTimeMs: 3600000,
         },
       ],
       pagination: {
@@ -45,7 +44,6 @@ describe("GET /api/leaderboard", () => {
       stats: {
         totalTokens: 1200,
         totalCost: 12.5,
-        totalActiveTimeMs: null,
         uniqueUsers: 1,
       },
       period: "all",
@@ -73,7 +71,6 @@ describe("GET /api/leaderboard", () => {
       "avatarUrl",
       "displayName",
       "rank",
-      "totalActiveTimeMs",
       "totalCost",
       "totalTokens",
       "userId",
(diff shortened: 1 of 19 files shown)


junhoyeo:
@codex review

Please review the current head after the stack rebase.

PR #869, opened 2026-07-12
fix(usage): derive Codex labels from window duration

astkaasa:
## Summary

- parse Codex `limit_window_seconds` instead of assuming the primary window is always 5 hours
- render weekly, day, hour, minute, and second window labels from the duration returned by the API
- preserve the legacy primary/secondary labels when older payloads omit the duration

## Root cause

Codex changed the general usage limit from a 5-hour window to a weekly window. Tokscale still labeled every `primary_window` as `5h`, so the percentage and reset timestamp were current while the displayed window name was stale.

The label now follows each window's own duration, so both the current weekly-only response and a future response containing both 5-hour and weekly windows render correctly regardless of primary/secondary ordering. The same behavior applies to additional model-specific limits such as Codex Spark.

## Validation

- `cargo test -p tokscale-cli commands::usage::codex::tests -- --test-threads=1`
- `cargo clippy --locked -p tokscale-cli --all-targets -- -D warnings`

---
## Summary by cubic
Fix Codex usage window labels by deriving them from `limit_window_seconds` instead of hardcoding 5h/weekly, so displayed limits match the API. Keeps legacy fallbacks for older payloads.

- **Bug Fixes**
  - Derive labels from `limit_window_seconds` and render Weekly/day/hour/minute/second.
  - Keep legacy "5h" (primary) and "Weekly" (secondary) when duration is missing.
  - Prefixes (e.g., "Spark week") now follow the window duration; tests added in `tokscale-cli`.

<sup>Written for commit 78fb6c5ae5a18359c9ad2925ee14a45cea7c58ec. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/869?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/usage/codex.rs
+++ b/crates/tokscale-cli/src/commands/usage/codex.rs
@@ -58,6 +58,7 @@ struct RateLimit {
 #[serde(rename_all = "snake_case")]
 struct Window {
     used_percent: Option<f64>,
+    limit_window_seconds: Option<i64>,
     #[serde(alias = "resets_at")]
     reset_at: Option<i64>,
 }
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
Thanks!

PR #880, opened 2026-07-13
fix(copilot): preserve trace-scoped parent hierarchy

### LINKED ISSUE #879: Copilot root-agent fallback loses attribute-less parent edges and crosses trace boundaries
## Summary

The root `invoke_agent` resolver added in #834 can still select the wrong trace fallback agent in two valid OpenTelemetry layouts: an intermediate span has no `attributes` object, or separate traces reuse the same `spanId`. Both cases can make an agentless Copilot usage record inherit a nested sub-agent instead of the root agent.

I reproduced both failures against `50ef057f5c59ea205410f96614fb93f460a47f58` (`v4.5.2-1-g50ef057f`). They were found while porting #834 into [TokenBar PR #44](https://github.com/Nanako0129/TokenBar/pull/44), where the two corrections and regressions are already running.

## Failure modes

| Case | Current behavior | Expected behavior |
|---|---|---|
| Attribute-less intermediate span | `resolve_trace_fallback_agents` reaches the `attributes` guard before recording the span's structural parent edge. The missing edge makes a nested `invoke_agent` appear to be a root candidate. | Parent links are collected from structural span fields even when the record has no attributes, so the nested invocation remains connected to the real root. |
| Reused span ID across traces | `parent_of` and `invoke_agent_span_ids` use bare `spanId` keys. A record from another trace can overwrite a parent edge or satisfy ancestor membership for the wrong trace. | Hierarchy identity is scoped by `(traceId, spanId)`, because OpenTelemetry span IDs are only unique within one trace. |

## Attribute-less intermediary reproduction

The relevant hierarchy is:

```text
invoke-root        agent = github.copilot.default
└── tool-task      no attributes
    └── invoke-sub agent = github.copilot.subagent
        └── chat-sub

chat-plain         no per-record agent
```

The intermediary still has `traceId`, `spanId`, and `parentSpanId`, but it has no `attributes` object. Because its `tool-task -> invoke-root` edge is currently skipped, `invoke-sub` is treated as a root invocation. When it appears first in export order, `chat-plain` inherits `github.copilot.subagent` instead of `github.copilot.default`. `chat-sub` should continue to keep its per-record sub-agent attribution.

## Cross-trace span ID reproduction

Two traces may legally reuse the same span ID:

```text
trace-a: invoke-root(spanId = root-shared) -> intermediary -> invoke-sub
trace-b: tool-span(spanId = root-shared) -> another parent
```

With a process-wide `HashMap<&str, &str>` and `HashSet<&str>`, trace B can replace trace A's `root-shared` parent entry or make an ancestor look like an invocation from the wrong trace. The root walk then classifies trace A's real root as nested and can fall back to its earlier nested agent. Each trace should resolve independently.

## Proposed correction

This does not require changing #834's fallback order or attribution policy. The narrow correction is:

1. Collect `spanId -> parentSpanId` structural edges immediately after reading `traceId`, before requiring `attributes`.
2. Store parent edges as `(traceId, spanId) -> parentSpanId`.
3. Store invocation membership as `(traceId, spanId)`.
4. Pass `traceId` into `is_root_invoke_agent` and keep every ancestor lookup within that trace.
5. Increment the Copilot parser cache version so entries created by the previous resolver are reparsed even when their source fingerprints are unchanged.

Per-record `gen_ai.agent.id` should continue to take precedence over the trace fallback, the first root invocation with an agent ID should remain the preferred fallback, and the existing cycle guard should remain unchanged.

## Regression coverage

A companion PR will include two hermetic parser regressions:

| Regression | Assertion |
|---|---|
| Nested invocation through an attribute-less intermediary | The nested chat retains `github.copilot.subagent`, while an agentless chat in the trace inherits `github.copilot.default`. |
| Two traces reuse the same span ID | Trace A resolves its root agent and trace B resolves its own agent without cross-trace hierarchy contamination. |
| Cached output from the previous Copilot parser version | The stale entry is rejected, the source is reparsed through the corrected resolver, and the corrected attribution survives save and reload. |

The existing single-invocation behavior should remain unchanged. The correction affects agent attribution only; it does not change token parsing, scanner discovery, cache architecture, or other Copilot data sources. It does require a Copilot parser-version increment because the affected attribution is part of the cached `UnifiedMessage` output.

Nanako0129:
## Summary

Closes #879.

This keeps the Copilot root-agent resolver from losing structural parent links on spans without `attributes` and from mixing hierarchies when separate OpenTelemetry traces reuse the same `spanId`. It also advances the Copilot parser cache version so previously cached mis-attribution is reparsed.

## Root cause

`resolve_trace_fallback_agents` previously collected `parentSpanId` only after requiring an `attributes` object. An attribute-less task or tool span could therefore remove the only edge connecting a nested `invoke_agent` to the real root. The hierarchy maps were also keyed by bare `spanId`, although OpenTelemetry only requires span IDs to be unique within a trace.

## Changes

| Change | Result |
|---|---|
| Collect structural parent edges before the attributes gate | Attribute-less intermediary spans remain part of the root walk. |
| Key parent edges and invocation membership by `(traceId, spanId)` | Separate traces cannot overwrite or satisfy each other's hierarchy entries. |
| Increment the Copilot parser cache version from 3 to 4 | Existing shards carrying the old fallback attribution are rejected and rebuilt. |
| Keep the existing root-first fallback, per-record precedence, and cycle guard | The correction does not change the intended #834 attribution policy. |

## Regression coverage

| Scenario | Expected result |
|---|---|
| Nested invocation behind an attribute-less intermediary | The nested chat keeps its sub-agent, while an agentless chat inherits the real root agent. |
| Two traces reuse the same span ID | Each trace resolves its own root independently. |
| Parser-version 3 shard with the old attribution | The stale entry is rejected, reparsed through the real Copilot parser, saved as version 4, and reloaded with the root attribution intact. |
| Existing single invocation | Behavior remains unchanged. |

## Verification

| Gate | Result |
|---|---|
| Pre-fix attribute-less intermediary regression | Failed as expected: selected `github.copilot.subagent` instead of `github.copilot.default`. |
| Pre-fix cross-trace span collision regression | Failed as expected: selected `github.copilot.subagent-a` instead of `github.copilot.root-a`. |
| Pre-fix parser-version 3 cache regression | Failed as expected: the unchanged-fingerprint stale entry was served. |
| Copilot hierarchy targeted tests | 5 passed. |
| Stale-v3 cache rebuild regression | Passed. |
| `cargo test -p tokscale-core` | 1,180 passed, 1 ignored. |
| `cargo test --workspace --all-features` | Passed. |
| `cargo fmt --all -- --check` | Passed. |
| `cargo clippy --locked --workspace --all-features -- -D warnings` | Passed with no lint findings. |
| `git diff --check origin/main...HEAD` | Passed. |

---
## Summary by cubic
Fixes Copilot agent attribution by preserving trace-scoped parent links and avoiding cross-trace span ID collisions. Also bumps the Copilot parser cache to v4 to invalidate stale v3 entries.

- **Bug Fixes**
  - Collect OTel parent edges before the attributes gate so attribute-less spans keep nested invoke_agent chains linked to the root.
  - Scope parent and invoke identities by (traceId, spanId) and traverse within the current trace to prevent hierarchy mixing when span IDs are reused.
  - Keep existing root-first fallback, per-record precedence, and cycle guard.
  - Bump the Copilot parser cache to v4; stale v3 shards are rejected and rebuilt.

<sup>Written for commit 20d9096a68a40d4a4e83581b0e0dd308aadc5ab7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/880?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -794,7 +794,7 @@ fn parser_version(client: ClientId) -> u32 {
         // so future changes have an obvious local version to increment.
         ClientId::Codex => 4,
         ClientId::Jcode => 4,
-        ClientId::Copilot => 3,
+        ClientId::Copilot => 4,
         // Pi subagent sessions now derive agent attribution from session_info
         // names; version-1 caches carry those messages without agent metadata.
         ClientId::Pi => 2,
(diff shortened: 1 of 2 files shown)


junhoyeo (approved):
Thanks!

PR #885, opened 2026-07-14
fix(devin-desktop): show pre-dedup message count and Windows paths in clients command

### LINKED ISSUE #872: Devin Desktop client reports zero usage on default devin-cloud agent (Windows)
## Summary

`tokscale --client devin-desktop` reports **0 messages / 0 tokens** on Windows even after days of active Devin Desktop usage. The ACP event log files exist and keep growing, but none of them ever contain a `usage_update` notification — only structural events (`tool_call`, `user_message_chunk`, `agent_message_chunk`, `session_info_update`, `config_option_update`, `current_mode_update`, `available_commands_update`).

For a Devin Desktop user on the default `devin-cloud` ACP agent (i.e. the vast majority of users — this is what "using Devin" means for most people), the `devin-desktop` client currently has no path to ever show real usage.

## Environment

- tokscale: `4.5.0`
- OS: Windows 11
- Devin Desktop: installed at `%LOCALAPPDATA%\Programs\Devin\Devin.exe`
- `devin.acp.enabledAgents` in `%APPDATA%\Devin\User\settings.json`: `{"devin-cloud": true}` (default)

## Steps to reproduce

1. Use Devin Desktop normally on Windows (default `devin-cloud` agent) for several days, including active multi-model/worktree sessions.
2. Run `tokscale clients` — `Devin Desktop` shows `sessions: ~/Library/Application Support/Devin/User/acp-events ✗` and `messages: 0`. (Note: the displayed path is the macOS path; on Windows the scanner does correctly also check `%APPDATA%\Devin\User\acp-events\`, which does exist and does contain files — see below.)
3. Run `tokscale models --client devin-desktop` — returns `Total: 0 messages, 0 tokens, $0.00`.

## Investigation

`%APPDATA%\Devin\User\acp-events\` contains multiple `.ndjson` files that are actively updated as Devin Desktop is used (11 files observed, ranging from a few KB to ~270 KB, spanning several weeks). Inspecting every line of every file for `sessionUpdate` values found only:

```
tool_call, tool_call_update, user_message_chunk, agent_message_chunk,
agent_thought_chunk, session_info_update, config_option_update,
current_mode_update, available_commands_update
```

**Not a single line across any file contains `usage_update`**, which is the only event type `parse_devin_desktop_ndjson_with_lookup` (in `crates/tokscale-core/src/sessions/devin.rs`) extracts token counts from (via `_meta["cognition.ai/inputTokens"]` etc., or the legacy `metadata.metrics` fallback).

The parser's own code comment acknowledges this is expected for many files:

```rust
// The Desktop app streams ACP events. Usage is not reliably present in
// the NDJSON itself; the authoritative usage lives in the CLI SQLite DB.
// We extract any embedded usage blocks we can find, but most files will
// yield no messages. ...
```

In practice, on this machine, **all** files yield no messages — because the ACP agent in use is `devin-cloud` (Devin's own hosted backend, and the default), which apparently never emits `usage_update` locally. Usage presumably stays server-side only.

I have been actively using Devin Desktop throughout this period (including multi-model/worktree sessions), so this isn't a case of the app being idle — the ACP event stream simply never records usage for the `devin-cloud` agent.

## Impact

For anyone using Devin Desktop the "normal" way (default `devin-cloud` agent, not manually switching to an ACP-connected third-party CLI agent), `tokscale --client devin-desktop` currently can't show any usage at all. This seems like a significant gap for the most common Devin Desktop usage pattern.

## Question for maintainers

- Is there a plan/roadmap for surfacing `devin-cloud` usage (e.g. via an account-level API), given it's the default agent and doesn't emit local ACP `usage_update` events?
- If `devin-desktop` support is currently scoped only to ACP-connected third-party agents (`claude-code`, `opencode`, `gemini`, etc. routed through the Cascade panel) rather than Devin's own cloud backend, it might be worth calling that out explicitly in the README so users on the default configuration don't expect to see data.

Happy to provide more log samples or test against other scenarios if helpful.

thedavidweng:
Closes #872.

The clients command showed 0 messages for Devin Desktop even when Cascade sessions were being parsed correctly, because the dedup filter suppressed messages matching CLI database session IDs.

Changes:
- Show pre-dedup message count for Devin Desktop in parse_local_clients
- Display devin-desktop additional scan paths (including Windows APPDATA) in clients command
- Make devin_desktop_additional_roots public for reuse in CLI display
- Add README note clarifying devin-cloud vs ACP-connected agent support

Test plan:
- cargo test -p tokscale-core: 1185 passed, 0 failed
- tokscale clients: Devin Desktop now shows messages: 1 (previously 0) and additional paths
- tokscale models --client devin-desktop: correctly reports Cascade session tokens

---
## Summary by cubic
Fixes Devin Desktop showing 0 messages in the `clients` command by using the pre-dedup count and listing Windows scan paths. Clarifies agent support in the README. Closes #872.

- **Bug Fixes**
  - Use pre-dedup Devin Desktop count in parse_local_clients so `clients` shows sessions even when they overlap with the CLI DB.
  - Display additional Devin Desktop scan roots in `clients`, including `%APPDATA%\Devin\User\acp-events\` on Windows.
  - Make `scanner::devin_desktop_additional_roots` public for CLI reuse.
  - README: clarify that `devin-cloud` doesn’t emit local `usage_update` events; ACP-connected agents (Cascade/Windsurf, `claude-code`, `opencode`) are supported.
  - Add regression test to pin pre-dedup Desktop count when a CLI session dedups the Desktop row.

<sup>Written for commit d897377df9647cad10e0dc98610789f7a1418008. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1351,6 +1351,8 @@ AI coding tools store their session data in cross-platform locations. Most tools
 | Devin Desktop | Linux: `~/.config/Devin/User/acp-events/`; macOS: `~/Library/Application Support/Devin/User/acp-events/` | `%APPDATA%\Devin\User\acp-events\` | Parses ACP usage events; the CLI database resolves matching session titles when present |
 | Synthetic | Re-attributed from other sources | Re-attributed from other sources | Detects `hf:` model prefix + `synthetic` provider |
 
+> **Devin Desktop agent support**: Local usage parsing works for ACP-connected agents (e.g. Cascade/Windsurf, claude-code, opencode) that emit `usage_update` events in the NDJSON stream. The default **devin-cloud** agent does not emit local `usage_update` events — its usage stays server-side and cannot be tracked by tokscale without an account-level API.
+
 > **Note**: On Windows, `~` expands to `%USERPROFILE%` (e.g., `C:\Users\YourName`). These tools intentionally use Unix-style paths (like `.local/share`) even on Windows for cross-platform consistency, rather than Windows-native paths like `%APPDATA%`.
 
 #### Windows-Specific Configuration
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -3878,6 +3878,20 @@ fn run_clients_command(json: bool, home_dir: Option<String>) -> Result<()> {
                         exists: path.exists(),
                     });
                 }
+                if client == ClientId::DevinDesktop {
+                    for root in tokscale_core::scanner::devin_desktop_additional_roots(
+                        &home_dir_str,
+                        use_env_roots,
+                    ) {
+                        let path_str = root.to_string_lossy().to_string();
+                        if !additional_paths.iter().any(|p| p.path == path_str) {
+                            additional_paths.push(AdditionalPath {
+                                path: path_str,
+                                exists: root.exists(),
+                            });
+                        }
+                    }
+                }
                 let legacy_paths = if client == ClientId::OpenClaw {
                     vec![
                         LegacyPath {
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -7770,6 +7770,92 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_local_clients_devin_nonzero_cli_usage_dedups_desktop_row_but_keeps_raw_count() {
+        let temp_dir = tempfile::TempDir::new().unwrap();
+        let db_path = temp_dir.path().join(".local/share/devin/cli/sessions.db");
+        std::fs::create_dir_all(db_path.parent().unwrap()).unwrap();
+        let conn = rusqlite::Connection::open(&db_path).unwrap();
+        conn.execute_batch(
+            "CREATE TABLE sessions (
+                 id TEXT PRIMARY KEY,
+                 working_directory TEXT NOT NULL,
+                 backend_type TEXT NOT NULL,
+                 model TEXT NOT NULL,
+                 title TEXT,
+                 agent_mode TEXT NOT NULL,
+                 created_at INTEGER NOT NULL,
+                 last_activity_at INTEGER NOT NULL
+             );
+             CREATE TABLE message_nodes (
+                 row_id INTEGER PRIMARY KEY AUTOINCREMENT,
+                 session_id TEXT NOT NULL,
+                 node_id INTEGER NOT NULL,
+                 parent_node_id INTEGER,
+                 chat_message TEXT NOT NULL,
+                 created_at INTEGER NOT NULL,
+                 metadata TEXT
+             );",
+        )
+        .unwrap();
+        conn.execute(
+            "INSERT INTO sessions (id, working_directory, backend_type, model, title, agent_mode, created_at, last_activity_at) VALUES ('cli-session', '/tmp/project', 'windsurf', 'gpt-5', 'Desktop task', 'accept-edits', 1, 1)",
+            [],
+        )
+        .unwrap();
+        // Unlike the zero-usage regression test above, this CLI row carries
+        // real attributable usage, so it must NOT be filtered by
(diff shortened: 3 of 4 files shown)


95leohuang:
Thank you for looking into this! I found another local data source that may help with the investigation.

On Windows, native Cascade trajectories are stored under the legacy Windsurf/Codeium path:

`%USERPROFILE%\.codeium\windsurf\cascade\*.pb`

Devin Desktop continues writing new files there. Each filename stem can be used as the Cascade ID.

I was able to retrieve the decoded trajectory by calling the local Devin/Windsurf language server RPC:

`/exa.language_server_pb.LanguageServerService/GetCascadeTrajectory`

The installed language server is located approximately at:

`%LOCALAPPDATA%\Programs\Devin\resources\app\extensions\windsurf\bin\language_server_windows_x64.exe`

The main usage data is available in:

`trajectory.generator_metadata[].chat_model.usage`

It contains fields such as:

- `model_uid`
- `billing_model_uid`
- `input_tokens`
- `output_tokens`
- `cache_read_tokens`
- `cache_write_tokens`

The corresponding timestamp and cost are available from:

- `chat_model.chat_start_metadata.created_at`
- `chat_model.model_cost`

Some additional model usage is available under:

`trajectory.steps[].metadata.model_usage`

Subtrajectories also need to be traversed recursively through `step.subtrajectory` and `step.subtrajectories[]`.

Using this source, I was able to obtain non-zero input, output, cache-read, and cache-write usage for native Cascade sessions. I hope this provides a useful lead for the implementation.

thedavidweng:
Thank you for the detailed investigation! This is a very promising lead.

I confirmed the .pb files exist on macOS too (~/.codeium/windsurf/cascade/*.pb), and the language server binary is available at /Applications/Devin.app/Contents/Resources/app/extensions/windsurf/bin/language_server_macos_arm.

However, the .pb files appear to be encrypted/obfuscated (no readable strings via strings(1)), so decoding them requires calling the language server RPC as you described. This would be a separate, more complex feature — it needs to spawn the platform-specific language server binary, connect via gRPC, and call GetCascadeTrajectory for each session.

The current PR (#885) addresses the immediate issue: Cascade sessions that emit usage_update events in the Devin Desktop NDJSON stream are now correctly counted. This works for the macOS Cascade setup I tested with.

The .pb trajectory approach would be valuable as a follow-up for cases where:
1. The NDJSON stream doesn't contain usage_update events (e.g. older sessions)
2. More granular per-step usage data is needed (trajectory.steps[].metadata.model_usage)

Would you be open to opening a separate issue for the .pb trajectory parser so it can be tracked independently?

95leohuang:
Thank you for confirming the same trajectory source on macOS!

Yes, I have opened #894 to track the native Cascade trajectory parser separately.

For clarification, the Windows approach is already working as a proof of concept rather than only being a possible lead. It successfully decoded all of my local Cascade trajectory files and extracted non-zero input, output, cache-read, and cache-write usage.

One implementation detail that may reduce the complexity: I did not need a full gRPC client. The language server exposes `GetCascadeTrajectory` as a unary Connect RPC over HTTP with a serialized protobuf request and response.

I included the working process, RPC headers, language-server arguments, protobuf fields, recursive subtrajectory handling, and dedup considerations in the new issue.

Thanks again for investigating this and for separating it from the NDJSON improvement in this PR.

junhoyeo (approved):
Reviewed: pre-dedup count logic and %APPDATA% path verified correct; regression test for the CLI-overlap dedup scenario added (revert-checked). Squash-merging.

PR #896, opened 2026-07-15
fix(codex): keep legacy replay turns behind child boundary

### LINKED ISSUE #895: Codex: nested child replay with legacy UUID v4 turn IDs bypasses the fork gate
## Summary

Tokscale 4.5.3 can count a nested Codex subagent's replayed ancestor history as live child usage when the replay begins with a legacy UUID v4 `turn_id`. This is a narrower follow-up to #679/#681: the current fork boundary handles UUID v7 ordering, but a UUID v4 replay turn bypasses that gate before the child's actual live turn starts.

## Reproduction shape

A depth-2 child JSONL begins with:

1. the child's UUID v7 `session_meta`;
2. replayed depth-1 and root `session_meta` records;
3. an ancestor `task_started` with a UUID v4 `turn_id` and a `started_at` timestamp earlier than the child session;
4. a matching UUID v4 `turn_context` followed by replayed cumulative `token_count` snapshots;
5. later, the child's real `task_started`/`turn_context`, whose timestamp is at or after the child session.

Expected: keep skipping through step 4 and count only the child's live usage from step 5 onward.

Actual: `forked_child_turn_starts_own_session` returns `true` when `codex_uuid_v7_order_key(turn_id)` is `None`, so the first UUID v4 ancestor `turn_context` opens the gate and the remaining replay is counted as child usage.

In an anonymized local corpus, three depth-2 child files were reported as 43,577 messages in aggregate even though only 83 post-spawn live token events remained after applying the task boundary.

## Proposed fix

For subagent forks, treat a non-v7 turn as child-local only when a qualifying `task_started` announced it. Filter `task_started` events by causal time: compare UUID v7 timestamps when available, otherwise compare `payload.started_at` against the child session's UUID v7 timestamp. Preserve the current fallback for human forks, which do not emit `task_started`.

I have a regression test and a small parser patch ready.

xingfanxia:
## What changed

- Parse `task_started.started_at` for Codex sessions.
- Admit a replay-window `task_started` only when its UUID v7 timestamp or legacy timestamp places it at/after the child session.
- Require non-v7 subagent turns to have a qualifying child-local `task_started` before they can end the replay skip.
- Add a depth-2 regression fixture with a replayed UUID v4 ancestor turn and a later live child turn.

## Why

Nested Codex child logs can replay ancestor `task_started`, `turn_context`, and cumulative `token_count` events before the child's own first turn. The existing gate returned `true` immediately when a replayed `turn_id` was not UUID v7, so legacy UUID v4 ancestor turns opened the gate and inflated the child's usage.

Human forks retain the existing fallback because they do not emit `task_started`. UUID v7 subagent behavior is unchanged except that replayed pre-child task starts are no longer added to the boundary set.

Fixes #895.

## Validation

- `cargo fmt --check`
- `cargo test -p tokscale-core sessions::codex::tests` (57 passed)
- `cargo test -p tokscale-core` (1186 passed, 1 ignored; integration tests passed)

---
## Summary by cubic
Fixes inflated token usage in nested Codex child sessions by keeping replayed legacy turns behind the child boundary and hardening `task_started` parsing. Non-integer `started_at` values no longer break parsing or open the gate.

- **Bug Fixes**
  - Parse and use `task_started.started_at` to order legacy UUID v4 turns; decode it leniently (int/float → i64; else None).
  - Admit replay-window `task_started` only if its v7 ms or `started_at` is at/after the child’s fork second; seconds-unit contract pinned.
  - For non‑v7 subagent turns, require a child‑local `task_started` to end the replay; keep the user‑fork fallback.
  - Ignore replayed pre‑child `task_started` for UUID v7 subagents.
  - Safer parsing (no panic on missing `turn_id`) and new tests: depth‑2 nested replay, seconds‑boundary, and malformed `started_at`.

<sup>Written for commit f6b4ab72b24868dcdd8c7c55e78cb57951ddfd2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/896?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/codex.rs
+++ b/crates/tokscale-core/src/sessions/codex.rs
@@ -41,6 +41,9 @@ pub struct CodexPayload {
     pub model_info: Option<CodexModelInfo>,
     pub info: Option<CodexInfo>,
     pub turn_id: Option<String>,
+    /// Unix timestamp (seconds) from `task_started` events. Legacy Codex turns
+    /// may use UUID v4 ids, so this is their only causal ordering signal.
+    pub started_at: Option<i64>,
     pub source: Option<Value>,
     /// Thread origin from session_meta. `"user"` marks a human-initiated fork
     /// (e.g. a VS Code "fork conversation"), which replays parent history but
@@ -337,9 +340,19 @@ fn parse_codex_reader<R: BufRead>(
                             && payload.payload_type.as_deref() == Some("task_started")
                         {
                             // The child's own turn is introduced by a
-                            // `task_started`; remember its turn_id so the gate can
-                            // recognize the child's own same-millisecond turn.
-                            if let Some(turn_id) = payload.turn_id.as_deref() {
+                            // `task_started`; remember it only when its id or
+                            // timestamp places it at/after the child session.
+                            // Nested child logs can replay ancestor task_started
+                            // events before the child's live turn.
+                            if forked_child_task_starts_own_session(
+                                &state,
+                                payload.turn_id.as_deref(),
+                                payload.started_at,
+                            ) {
+                                let turn_id = payload
+                                    .turn_id
+                                    .as_deref()
+                                    .expect("recognized child task_started events have a turn_id");
                                 state
                                     .forked_child_task_started_turn_ids
                                     .insert(turn_id.to_string());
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
Reviewed: replay-gate logic verified sound; started_at seconds-unit contract confirmed against codex-rs protocol source and hardened with lenient deserialization + regression tests. Squash-merging.

Issue #59, opened 2025-12-28
Remove duplicated features (fallback) in TypeScript and only keep Rust core


junhoyeo:
- Since we're packaging with napi/rs + bunx, actually no need to maintain both versions

PR #914, opened 2026-07-17
fix(cache): complete atomic shard writes on Windows

### LINKED ISSUE #913: Windows cache shard writes can fail with self-held or read-only file handles
## Problem

On Windows, `crates/tokscale-core/src/message_cache.rs::write_shard_with_limit` has two caller-owned handle states that can make an otherwise successful cache shard write return an error:

1. The temporary `BufWriter<File>` remains alive when `fs_atomic::replace_file` calls `MoveFileExW`. The caller therefore attempts replacement while it still owns the temporary file handle.
2. After replacement, the final shard is reopened with `File::open` and passed to `sync_all()`. That is a read-only handle, while Windows `FlushFileBuffers` requires a handle opened with write access.

The current ordering is effectively:

```rust
writer.flush()?;
writer.get_ref().sync_all()?;
replace_file(&tmp_path, final_path)?;
File::open(final_path)?.sync_all()?;
```

The expected behavior is that a fully serialized and synced temporary shard can be atomically installed and durably synced without a handle owned by the cache writer causing `Access is denied` or another Windows I/O failure.

## Relationship to #893 and #906

Issue #893 reported repeated Windows cache shard save failures and explicitly suspected that the program still held a cache-file reference. PR #906 added bounded retries for transient antivirus, indexer, and cloud-sync contention in `MoveFileExW`, and prevented the warning from corrupting the TUI. It did not release the cache writer's own temporary handle before replacement or change the access mode used for the final durability sync.

Those retry and TUI changes remain useful and are outside this issue. This report covers the deterministic handle ownership and access mode at the shard writer call site.

## Downstream evidence

TokenBar's Windows downstream encountered and fixed the same handle-ordering failure in its monolithic source-message cache. The fix was validated on Windows through the downstream Rust suite, hostile-environment suite, x64 release build, profile-manifest check, and FFI smoke, then recovered byte-for-byte into [TokenBar PR #59](https://github.com/Nanako0129/TokenBar/pull/59). The full evidence chain is documented in the [Windows downstream handoff](https://github.com/Nanako0129/TokenBar/issues/45#issuecomment-5002865759), with the vendor fix at `aec5bd88` and validated descendants `6ada038` and `d40fc54`.

Current upstream uses shard files rather than TokenBar's monolithic cache, so the downstream patch cannot be cherry-picked wholesale. The handle-ordering invariant maps directly to `write_shard_with_limit`; cache format, shard identity, retry policy, parser versions, and TUI warning behavior do not need to change.

## Proposed fix

- Flush and sync the temporary writer, then explicitly drop it before `replace_file`.
- Reopen the final shard with `OpenOptions` read and write access before `sync_all()`.
- Add a direct shard write/read regression.

## Verification boundary

The regression and full workspace gates can establish platform-neutral non-regression. The existing `Build Native` workflow compiles Windows x64 and arm64 binaries but does not execute Windows tests. Unless a Windows runner checks out the proposed branch and runs the regression, upstream Windows runtime verification should remain attributed to the downstream evidence above rather than inferred from macOS/Linux tests or compile-only CI.

Nanako0129:
Fixes #913

## Summary

- Release the temporary cache shard writer before atomic replacement so the cache does not keep its own source handle open across `MoveFileExW`.
- Reopen the installed shard with read and write access before the final durability sync, matching the access requirements of Windows `FlushFileBuffers`.
- Add a direct shard write/read regression without changing cache format, parser versions, shard identity, retry policy, or TUI warning behavior.

## Current-upstream adaptation

| Validated downstream behavior | Current upstream location | Adaptation |
|---|---|---|
| Flush and sync the monolithic temporary cache, then drop its writer before replacement | `write_shard_with_limit` temporary envelope writer | Apply the same handle ordering to each shard writer |
| Reopen the installed cache read/write before `sync_all()` | Final shard durability sync | Use `OpenOptions::new().read(true).write(true)` |
| Monolithic cache layout and schema | Sharded envelopes and per-parser versions | Deliberately excluded; no format or identity change |
| Atomic replacement implementation | Shared `fs_atomic::replace_file` with #906 retry behavior | Deliberately unchanged |

The source behavior was recovered into [TokenBar PR #59](https://github.com/Nanako0129/TokenBar/pull/59). The Windows validation ancestry and full downstream evidence are recorded in the [handoff comment](https://github.com/Nanako0129/TokenBar/issues/45#issuecomment-5002865759).

## Verification

- `cargo test -p tokscale-core test_write_shard_round_trips_after_atomic_replace -- --test-threads=1`
- `cargo test -p tokscale-core test_save_if_dirty_marks_cache_clean -- --test-threads=1`
- `cargo build --release -p tokscale-cli`
- `git diff --check`

## Windows evidence boundary

The downstream handle-ordering fix was executed and validated on Windows. This PR adapts that invariant to the current shard writer. Upstream `Build Native` checks Windows x64 and arm64 compilation only; it does not execute this regression on Windows. Unless an additional Windows runner runs the test, this PR does not claim independent upstream Windows runtime verification.

---
## Summary by cubic
Fixes atomic shard writes on Windows by closing the temp writer before replacement and reopening the final shard for a proper flush. Prevents handle conflicts and ensures cache durability after writes.

- **Bug Fixes**
  - Drop the temporary shard writer before atomic replace to avoid carrying an open source handle into `MoveFileExW`.
  - Reopen the installed shard with read+write before the final `sync_all()` to satisfy Windows `FlushFileBuffers`.
  - Add a regression test to verify shard write/read round-trip after replace.
  - No changes to cache format, shard identity, parser versions, retry policy, or TUI behavior.

<sup>Written for commit e15e66bd72052cc3c7b4d2782bd844d7f7c6b308. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/914?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1356,8 +1356,10 @@ fn write_shard_with_limit(
             .map_err(std::io::Error::other)?;
         writer.flush()?;
         writer.get_ref().sync_all()?;
+        drop(writer);
         crate::fs_atomic::replace_file(&tmp_path, final_path)?;
-        File::open(final_path)?.sync_all()?;
+        let final_file = OpenOptions::new().read(true).write(true).open(final_path)?;
+        final_file.sync_all()?;
         Ok(())
     })();
 
@@ -2531,6 +2533,29 @@ mod tests {
         std::fs::write(file.path(), rewritten).unwrap();
 
         assert!(!codex_prefix_matches(file.path(), &incremental_cache));
+    }
+
+    #[test]
+    fn test_write_shard_round_trips_after_atomic_replace() {
+        let source = write_temp_file(b"{}\n");
+        let identity = CacheIdentity::for_client(ClientId::Claude);
+        let entry = test_entry(identity, source.path(), "session-1");
+        let shard_dir = TempDir::new().unwrap();
+        let shard_path = shard_dir.path().join("shard.bin");
+
+        write_shard_with_limit(
+            &shard_path,
+            identity,
+            std::slice::from_ref(&entry),
+            MAX_CACHE_SHARD_BYTES,
+        )
+        .unwrap();
+
+        assert!(matches!(
+            read_shard(&shard_path, identity),
+            ShardReadStatus::Loaded(entries)
+                if entries.len() == 1 && entries[0].messages[0].session_id == "session-1"
+        ));
     }
 
     #[test]


junhoyeo (approved):
Thanks a lot!

Issue #909, opened 2026-07-17
Multi-device submissions drop same-client usage on overlapping dates after #886


BenLampson:
## Summary

PR #886 reintroduced data loss for users who actively use the same client on multiple devices. When two distinct devices submit usage for the same date and client (for example, Codex on both a desktop and another workstation), the server keeps only the larger per-client snapshot instead of summing the independent device usage. The surviving daily row is then re-attributed to whichever device submitted last.

This regresses the multi-device aggregation behavior previously addressed by #52, #248, #517, and #560.

## Production reproduction

1. Sign in to the same Tokscale account on two machines with distinct generated device IDs.
2. Generate independent Codex usage on both machines on the same date.
3. Run `tokscale submit` on Machine A.
4. Run `tokscale submit` on Machine B.
5. Inspect the profile day breakdown and the per-device API responses.

Observed on July 16, 2026:

| Source | Tokens | Cost |
| --- | ---: | ---: |
| Machine A local report | 829,971,497 | $616.8952 |
| Machine B local report | approximately 140.6M | $115.83 |
| Expected account total | approximately 970.6M | $732.73 |
| Actual server/profile total | 829,971,497 | $616.8952 |

Both devices were registered with different device keys and both had recent successful submission timestamps. However, the per-device API attributed Machine A's 829,971,497-token row to the later-submitting Machine B, while Machine A had no row for that date. Machine B's independent approximately 140.6M-token usage was not present in the account total.

## Expected behavior

Distinct devices should own independent daily rows. Re-submitting from the same device should be idempotent for that device, while usage from different devices should be additive even when both devices use the same client and model on the same date.

For example, if Device A submits 100 Codex tokens and Device B submits 40 Codex tokens for the same date, the account total should be 140, with the device detail endpoints reporting 100 and 40 respectively.

## Actual behavior

The account retains one row for the date. When the same client key appears in both device payloads, `mergeClientBreakdownsWithRegressionGuard` treats the two snapshots as overlapping history and keeps the larger value. The update also changes `submitted_device_id` to the latest submitting device, so device attribution flips even though the retained usage originated on the other machine.

## Root cause

#886 intentionally restored uniqueness on `(submission_id, date)`, replacing the prior `(submission_id, submitted_device_id, date)` constraint. It fetches existing days across the whole account, merges colliding data by client key using newest-wins-with-regression-guard semantics, and reassigns the merged row to the current device.

[...]

PR #918, opened 2026-07-18
fix(submit): make daily_breakdown upsert conflict-target-agnostic (migration bridge)

junhoyeo:
## Why

**Phase 1 of an expand/contract rollout** for the `daily_breakdown` unique-key change in #910.

#910 swaps the table's unique key from `(submission_id, date)` to `(submission_id, submitted_device_id, date)` via migration 0017. But production applies migrations **before** `next build` (`vercel.json` buildCommand → `migrate-prod.ts`), and the **old** deployment keeps serving for the entire build. So the moment 0017 drops the `(submission_id, date)` constraint, the currently-deployed route's `ON CONFLICT (submission_id, date) DO UPDATE` throws Postgres **42P10** ("no unique or exclusion constraint matching the ON CONFLICT specification") on every insert-containing submit — for the whole build window, and indefinitely if that build later fails.

A Codex review of #910 (high effort) flagged this as a real availability + client-side-delivery-gap risk, not merely a few stray 500s. This PR removes the coupling so #910 can deploy cleanly.

## What

Replace the qualified conflict target on the `daily_breakdown` INSERT with an **unqualified `ON CONFLICT DO NOTHING`**, which matches whichever unique constraint currently exists — so this code runs correctly against **both** the pre-0017 and post-0017 schemas.

`DO NOTHING` (vs the old `DO UPDATE`) is behavior-preserving:
- `existingDaysMap` already routes every date that has a row for the submission to the `toUpdate` path — an `UPDATE ... WHERE d.id = batch.id`, itself constraint-agnostic.
- So `toInsert` holds only genuinely-new dates; the conflict clause is purely a **concurrent-insert race net** (another submit inserting the same key between the `SELECT ... FOR UPDATE` and this INSERT).
- On that rare race, skipping is safe because **STEP 3d recomputes the submission's denormalized totals by summing the surviving daily rows**, so the winning row is still counted.

No migration, no schema change, no other `daily_breakdown` upsert exists (verified). The legacy-adoption path already uses a plain UPDATE.

## Rollout order

1. **Merge + deploy THIS PR** → prod now runs conflict-target-agnostic code.
2. Then merge #910 (migration 0017 + device-scoped merge). Its build window is served by this code, which tolerates both schemas → **no submit 500s, no delivery gap, no transient #909 regression**.

## Tests

Existing 13 submit-auth tests pass unchanged. Added an assertion locking in the bridge invariant: the INSERT must emit an unqualified `ON CONFLICT DO NOTHING` and must **not** name `(submission_id, date)`. `tsc --noEmit` clean.

---
## Summary by cubic
Makes the `daily_breakdown` insert conflict clause schema-agnostic by using `ON CONFLICT DO NOTHING`, so submits work during the unique-key migration and avoid Postgres 42P10 errors. Also clarifies race-handling comments to reflect the actual protection path.

- **Bug Fixes**
  - Replace `ON CONFLICT (submission_id, date) DO UPDATE` with `ON CONFLICT DO NOTHING` in `packages/frontend/src/app/api/submit/route.ts` to work with both `(submission_id, date)` and `(submission_id, submitted_device_id, date)`.
  - Add a test in `packages/frontend/__tests__/api/submitAuth.test.ts` asserting the unqualified conflict clause, and update comments to attribute race handling to the UPDATE’s `NOT EXISTS` check plus savepoint (not the insert).

- **Migration**
  - Deploy this before the key change to `(submission_id, submitted_device_id, date)` (in #910 / migration 0017) to prevent submit failures during the build window.

<sup>Written for commit 221f6a4aa720eebb6f6ee5b9e494e3543b324c5b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/918?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/__tests__/api/submitAuth.test.ts
+++ b/packages/frontend/__tests__/api/submitAuth.test.ts
@@ -506,7 +506,8 @@ describe("POST /api/submit auth path", () => {
       displayName: "Test device",
     }));
     expect(tx.execute).toHaveBeenCalledTimes(1);
-    expect(flattenSqlChunks(tx.execute.mock.calls[0][0])).toEqual(
+    const insertChunks = flattenSqlChunks(tx.execute.mock.calls[0][0]);
+    expect(insertChunks).toEqual(
       expect.arrayContaining([
         expect.stringContaining("INSERT INTO daily_breakdown"),
         "submission-1",
@@ -514,6 +515,20 @@ describe("POST /api/submit auth path", () => {
         "2026-04-30",
       ]),
     );
+    // Expand/contract bridge invariant: the daily_breakdown INSERT must use an
+    // UNQUALIFIED `ON CONFLICT DO NOTHING` (no named constraint) so it survives
+    // the migration that swaps the table's unique key from (submission_id, date)
+    // to (submission_id, submitted_device_id, date). Naming either column set
+    // would throw 42P10 against the other schema during the deploy window.
+    expect(insertChunks).toEqual(
+      expect.arrayContaining([expect.stringContaining("ON CONFLICT DO NOTHING")]),
+    );
+    expect(
+      insertChunks.some(
+        (chunk) =>
+          typeof chunk === "string" && chunk.includes("ON CONFLICT (submission_id, date)"),
+      ),
+    ).toBe(false);
     expect(submissionUpdateValues).toEqual(
       expect.objectContaining({
         mcpServers: ["github", "slack"],
--- a/packages/frontend/src/app/api/submit/route.ts
+++ b/packages/frontend/src/app/api/submit/route.ts
@@ -370,11 +370,11 @@ export async function POST(request: Request) {
         // Race note: two concurrent submits from the same user can both reach
         // this branch before either has committed. The second UPDATE will try
         // to re-stamp submitted_device_id on rows the first already claimed,
-        // which can violate the (submission_id, submitted_device_id, date)
-        // unique constraint. The ON CONFLICT DO NOTHING below makes the UPDATE
-        // skip conflicting rows rather than throw, and the outer try/catch falls
-        // through to the normal insert path if a unique violation still escapes
-        // (e.g. via a concurrent INSERT racing the UPDATE window).
+        // which can violate a daily_breakdown unique constraint. The UPDATE's
+        // own NOT EXISTS dup predicate below skips rows that would collide, and
+        // the savepoint + outer try/catch fall through to the normal insert
+        // path if a unique violation still escapes (e.g. via a concurrent
+        // INSERT racing the UPDATE window).
         try {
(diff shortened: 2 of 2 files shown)


junhoyeo:
## Codex adversarial review — converged to APPROVE (2 rounds)

Ran two rounds of Codex adversarial review (gpt-5.6-terra, high). Final verdict: **APPROVE as a correct Phase-1 compatibility bridge**, with one operational condition.

**Confirmed:**
- `ON CONFLICT DO NOTHING` (unqualified) eliminates the 42P10 outage and matches whichever unique constraint exists, so it's correct against both the current `(submission_id, date)` key and #910's `(submission_id, submitted_device_id, date)` key.
- Behavior-equivalent to the old `DO UPDATE` against the **current** schema — duplicate input dates are rejected upstream (`validation/submission.ts`), existing dates route to the id-keyed UPDATE, so the conflict clause is purely a concurrent-insert race net; on that race, STEP 3d recomputes totals from surviving rows.
- The `#910-build-fails` state is safe: the bridge structurally can't create multi-device rows (every existing date routes to update-by-id), so it just retains current one-row-per-date behavior indefinitely — no new corruption.
- Correct baseline confirmed: judged against **current prod** behavior (which it preserves), the bridge introduces no new data loss. The same-client cross-device merge is pre-existing #909 behavior that #910 fixes — not introduced here.

**Fixed during review** (`221f6a4a`): a misattributed legacy-adoption comment that credited the daily-insert's `ON CONFLICT` for protection actually provided by the UPDATE's `NOT EXISTS` predicate + savepoint.

**Operational condition (the one real residual):** there is no request-drain/version fence in the platform. A bridge request admitted just before #910 promotes could execute its DB transaction *after* #910's device-aware code has written multi-device rows — a **seconds-wide cutover overlap** where the bridge's whole-submission map could re-attribute/overwrite another device's row or hit a clean 23505 rollback. This is inherent to the promotion cutover (not introduced by the bridge — the bridge shrinks the dangerous window from *the entire build* to *the promotion instant*).

### Required rollout policy
1. Promote **#918 to production before #910's migration runs.**
2. Do **not** roll back #910's application code while retaining migration 0017 (rollback past a non-additive migration is inherently unsafe).
3. Accept the seconds-wide promotion-cutover overlap (clean-failure-dominated, self-healing on retry), or deploy #910 at a low-traffic moment to minimize it.

CI: all 7 checks green including cubic.

PR #920, opened 2026-07-18
feat(opencode): parse OpenCode v2 session_message schema

saiqulhaq:
## Problem

OpenCode v2 (`opencode-next.db`) reports **zero** usage in tokscale. The v2 storage layout moves per-message data out of the `message` table (empty in v2) and into a new `session_message` table:

- The role lives in a `type` column instead of a JSON `$.role`.
- Model and provider are **nested** under `$.model` (`model.id` / `model.providerID`) rather than the top-level `$.modelID` / `$.providerID`.
- `tokens`, `cost`, and `time` JSON shapes are otherwise identical to v1.

The `opencode-next.db` file is already discovered (its name matches the `opencode-<channel>.db` rule), but `parse_opencode_sqlite` only queried the v1 `message` table, so nothing was parsed.

## Change

Extend the SQLite parser to read both schemas:

- `OpenCodeMessage.role` is now optional, and a nested `model { id, providerID }` is added. Model/provider resolve from the top-level v1 fields **or** the nested v2 object (top-level wins when both exist).
- The per-row processing is refactored into a shared accumulator; the new v2 `session_message` query and the existing v1 `message` query both feed through it, so v1 and v2 databases both work and any cross-table overlap deduplicates via the existing fingerprint logic.
- Workspace directory continues to come from `session.directory`.

## Tests

New unit tests cover nested-model resolution, top-level precedence, the `type='assistant'` filter (skipping user/synthetic/tokenless rows), workspace-from-`session.directory`, row-id dedup fallback, negative-token clamping, and forked-history dedup.

- `cargo clippy --locked -p tokscale-core --lib -- -D warnings` — clean
- `cargo test -p tokscale-core --lib` — 1226 passed, 0 failed

Verified against a real `opencode-next.db`: OpenCode usage went from `0` to ~10k parsed assistant messages across all models, with other clients unaffected.

---
## Summary by cubic
Adds OpenCode v2 SQLite support by parsing `session_message` with nested model/provider fields, fixing zero usage in `opencode-next.db`. Improves dedup to avoid merging distinct messages while preserving v1 behavior and keeping JSON file parsing strict.

- **New Features**
  - Parse v2 `session_message` (type='assistant') with nested `$.model.id`/`$.model.providerID`; resolve model/provider from top-level v1 fields when present.
  - Shared accumulator merges v1/v2 by fingerprint; workspace from `session.directory`; filter to assistant rows with tokens; clamp negative tokens.

- **Bug Fixes**
  - Require explicit `role: "assistant"` in JSON file parsing; missing-role acceptance stays scoped to the v2 SQLite path.
  - Prevent merging distinct messages that collide on fingerprint by tracking embedded `$.id` per entry; fork copies still collapse and dedup key promotion is preserved.

<sup>Written for commit 1b779740f3a6b14568323379999e3906aab25964. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/920?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/opencode.rs
+++ b/crates/tokscale-core/src/sessions/opencode.rs
@@ -141,7 +141,11 @@ struct OpenCodeSqliteFingerprint {
 
 #[derive(Debug, Clone)]
 struct OpenCodeSqliteDedupState {
-    has_embedded_message_id: bool,
+    /// The entry's embedded (`$.id`) message id, if any. Two rows that share
+    /// every fingerprint field but carry *different* embedded ids are distinct
+    /// messages, not fork copies, and must not be merged. A fork copies the id,
+    /// so equal ids (or an id absent on either side) still merge.
+    message_id: Option<String>,
     has_workspace_conflict: bool,
 }
 
@@ -276,11 +280,13 @@ type OpenCodeSqliteRow = (String, String, String, Option<String>);
 /// Accumulates parsed assistant messages across OpenCode's v1 (`message`) and
 /// v2 (`session_message`) tables, applying fingerprint-based deduplication so
 /// forked-history copies — and any overlap between the two tables — collapse
-/// into a single entry.
+/// into a single entry. A fingerprint maps to a *list* of entries, one per
+/// distinct embedded message id, so two genuinely different messages that
+/// happen to collide on every fingerprint field are kept apart.
 #[derive(Default)]
 struct OpenCodeSqliteAccumulator {
     messages: Vec<UnifiedMessage>,
-    fingerprint_indices: HashMap<OpenCodeSqliteFingerprint, usize>,
+    fingerprint_indices: HashMap<OpenCodeSqliteFingerprint, Vec<usize>>,
     dedup_states: Vec<OpenCodeSqliteDedupState>,
 }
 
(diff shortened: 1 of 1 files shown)


junhoyeo:
Reviewed this — clean, well-structured v2 support (shared accumulator, v1/v2 fingerprint dedup, thorough tests). Pushed one maintainer commit (7f27cb0c) for cubic's P2 finding, which is valid:

Making `OpenCodeMessage.role` optional for the v2 `session_message` path also loosened `parse_opencode_file` — `is_assistant()` treats a missing role as assistant, so a role-less or malformed OpenCode **JSON file** (previously skipped because the required `role` failed to deserialize) would now count as assistant usage. `parse_opencode_file` is a live path (called from the JSON-file scan in lib.rs), so this mattered.

Fix keeps the file path strict (`role` must be explicitly `"assistant"`); missing-role acceptance stays scoped to the v2 SQLite query, whose SQL already filters `type = 'assistant'`. Added a regression test covering a role-less and a user-role file. Verified locally: `cargo test -p tokscale-core --lib opencode` (89 passed), `cargo clippy --locked -p tokscale-core --lib -- -D warnings`, and `cargo fmt --all -- --check` all clean.

The rest of the change looks correct — the borrow-ordering adaptations (`resolve_*` before moving fields, `ref`/`clone` in the accumulator) and the v2-then-v1 query ordering with prepare-failure-as-no-op are all sound.

junhoyeo:
Pushed a second maintainer commit (1b779740) addressing a defect an independent Codex validation flagged in the dedup logic (pre-existing behavior this refactor surfaced, not a regression this PR introduces):

The SQLite dedup keyed only on a content fingerprint (timestamp, model, provider, clamped tokens, cost, agent) and tracked embedded message ids as a boolean, so two genuinely distinct assistant messages that collided on every fingerprint field would be merged into one — undercounting usage. The false-merge is astronomically unlikely for real distinct turns, but differing `$.id`s are a definitive distinctness signal the code ignored.

Fix: track the actual embedded `$.id` per accumulated entry, and map each fingerprint to a *list* of entries (one per distinct message identity). A row merges only when it is not a definitively-different message — equal ids, or an id absent on either side, still collapse (fork copies share the id); two differing ids are kept apart. Fork-history dedup and the row-id → embedded-id `dedup_key` promotion are preserved.

Added a regression test (two rows identical on every fingerprint field but with different `$.id` stay separate, while a same-id fork still collapses). Verified locally: `cargo test -p tokscale-core --lib opencode` (90 passed), clippy `-D warnings`, and `cargo fmt --all -- --check` all clean.

Codex's verdict on the rest (v1/v2 role handling, model/provider resolution, the borrow refactor, SQL/workspace) was that they're correct.

PR #925, opened 2026-07-20
fix(kimi): preserve larger saturated updates

junhoyeo:
## Summary
- Rank duplicate legacy Kimi `StatusUpdate` records with an exact `i128` token total.
- Keep the existing saturating total for reporting paths.
- Add a regression for a later smaller same-ID update that previously replaced a larger extreme usage record.

## Root cause
`TokenBreakdown::total()` saturates at `i64::MAX`. After #923 retained extreme Kimi buckets, different actual totals could tie during deduplication and the later timestamp could replace the larger breakdown.

## Validation
- `cargo test --locked -p tokscale-core test_parse_kimi_deduplication_preserves_larger_extreme_usage`
- `cargo test --locked --release -p tokscale-core test_parse_kimi_deduplication_preserves_larger_extreme_usage`
- `git diff --check`

Follow-up to #923.

---
## Summary by cubic
Fixes Kimi StatusUpdate dedup so the larger extreme-usage record is kept when totals saturate. Prevents a later smaller update with the same ID from replacing the real larger one.

- Bug Fixes
  - Rank duplicate Kimi `StatusUpdate` records with an exact i128 sum of the five token buckets to avoid `i64::MAX` ties.
  - Keep the saturating total only for reporting paths; existing progress ordering stays the same.
  - Added a regression test to ensure larger extreme-usage updates aren’t replaced by later smaller ones.

<sup>Written for commit 860b42d3ed5f3160883b45d3ccb52d9e7dd799e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/925?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/kimi.rs
+++ b/crates/tokscale-core/src/sessions/kimi.rs
@@ -321,11 +321,23 @@ pub fn parse_kimi_file(path: &Path) -> Vec<UnifiedMessage> {
 }
 
 fn should_replace_status_update(existing: &UnifiedMessage, candidate: &UnifiedMessage) -> bool {
-    let existing_total = existing.tokens.total();
-    let candidate_total = candidate.tokens.total();
+    let existing_total = exact_token_total(&existing.tokens);
+    let candidate_total = exact_token_total(&candidate.tokens);
 
     candidate_total > existing_total
         || (candidate_total == existing_total && candidate.timestamp >= existing.timestamp)
+}
+
+/// Compare Kimi progress updates without collapsing distinct large totals.
+///
+/// `TokenBreakdown::total` saturates for reporting, while five i64 buckets fit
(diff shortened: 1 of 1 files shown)


junhoyeo:
Closing as superseded by #927, merged as `1c74ba4b`. That PR implements the same exact `i128` token-total ordering and includes a stronger regression test. This draft now conflicts with `main` and has no remaining unique functional change.

PR #936, opened 2026-07-22
fix(pricing): apply 272k rates to full request

astkaasa:
## Summary

- Apply OpenAI's >272K full-request pricing only to the documented GPT-5.4, GPT-5.4 Pro, GPT-5.5, GPT-5.6, GPT-5.6 Sol, GPT-5.6 Terra, and GPT-5.6 Luna families, including dated snapshots.
- Trigger the high rates from the request's combined input context (uncached input + cache read + cache write), rather than evaluating token buckets independently.
- Preserve the generic progressive tier calculation for Sakana Fugu, Azure, custom pricing, and every other model.
- Keep cache-write pricing aligned with the model's documented high/base input-price ratio when the full-request rule is active.
- For unprefixed model IDs used by Codex and OpenCode with `provider=openai`, always prefer a LiteLLM OpenAI price row with complete input, output, and applicable cache-read 272K rates for the documented model families. OpenRouter price rows remain outside the OpenAI-specific full-request rule.

## Root cause

Codex and OpenCode record models such as `gpt-5.6-sol` separately from `provider=openai`. Provider-aware lookup matched OpenRouter's `openai/gpt-5.6-sol` row before LiteLLM's unprefixed row. OpenRouter's published pricing metadata exposes fixed prompt/completion/cache rates but no 272K threshold fields, so the full-request rule could not activate even though LiteLLM carried the required tier data.

The selection now follows the actual provider rather than the relative completeness of two independent pricing catalogs: when the provider is OpenAI and LiteLLM has the required tier data for a documented model family, LiteLLM is selected even if an OpenRouter row later gains tier fields. Traffic explicitly attributed to OpenRouter does not inherit OpenAI's full-request pricing semantics.

OpenAI documents the full-request behavior for [GPT-5.4](https://developers.openai.com/api/docs/models/gpt-5.4), [GPT-5.5](https://developers.openai.com/api/docs/models/gpt-5.5), and the [GPT-5.6 family](https://developers.openai.com/api/docs/models/gpt-5.6-sol). OpenRouter's [pricing schema](https://openrouter.ai/docs/guides/overview/models) does not expose context-threshold price fields.

## Tests

- Exact 272K boundary stays on normal rates.
- Combined input above 272K applies the high rates to the full request.
- Output volume alone does not trigger the full-request rule.
- Provider-aware lookup for `gpt-5.6-sol + provider=openai` selects tier-complete LiteLLM data when OpenRouter has only base rates.
- Provider-aware lookup keeps selecting LiteLLM for `provider=openai` even if an OpenRouter row gains tier fields.
- The LiteLLM preference requires valid base and above-272K input/output rates, plus an above-272K cache-read rate whenever a base cache-read rate exists.
- Non-OpenAI provider hints do not trigger the LiteLLM preference.
- Empty and `unknown` provider hints behave like an absent hint for both lookup and cost calculation.
- Model, source, and provider scope exclusions remain progressive.
- Existing generic 272K progressive-tier tests remain unchanged.

`cargo test -p tokscale-core`
`cargo clippy -p tokscale-core --all-targets -- -D warnings`
`cargo fmt --all -- --check`

### DIFF
--- a/crates/tokscale-core/src/pricing/lookup.rs
+++ b/crates/tokscale-core/src/pricing/lookup.rs
@@ -480,6 +480,11 @@ impl PricingLookup {
             }
         }
 
+        let exact_litellm = self.exact_match_litellm(model_id);
+        if should_prefer_openai_tiered_litellm(model_id, provider_id, exact_litellm.as_ref()) {
+            return exact_litellm;
+        }
+
         if let Some(result) = choose_best_source_result(
             self.exact_match_litellm_for_provider(model_id, provider_id),
             self.exact_match_openrouter_for_provider(model_id, provider_id),
@@ -488,7 +493,7 @@ impl PricingLookup {
             return Some(result);
         }
 
-        if let Some(result) = self.exact_match_litellm(model_id) {
+        if let Some(result) = exact_litellm {
             return Some(result);
         }
         // An unscoped OpenRouter FULL-KEY match is the id's own canonical key,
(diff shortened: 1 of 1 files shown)


junhoyeo (approved):
Thanks!

PR #937, opened 2026-07-22
fix(jcode): avoid double-counting cached input tokens

yousiki:
## Summary

- normalize Jcode usage so OpenAI/OpenRouter cached input is not counted again as paid input
- preserve Anthropic-style split accounting using message-local cache-creation presence, including an explicit zero value
- invalidate cached Jcode parse results by bumping the source parser version

## Problem

Jcode stores provider-reported `input_tokens` and cache buckets on assistant messages. OpenAI-style providers report `cache_read_input_tokens` as a subset of `input_tokens`, while Anthropic-style providers report separate buckets. The parser previously copied both fields directly into Tokscale, so cached input was included in the regular input bucket and then charged again through the cache-read bucket.

The normalization intentionally uses message-local usage shape rather than session-level provider/model metadata because Jcode sessions can switch routes and the top-level metadata may describe a later turn.

## Validation

- Jcode parser regression test run three consecutive times
- frozen local-history comparison: input `492,295,743` → `36,180,031`, cache read remained `456,115,712`, output/message totals were unchanged, and estimated cost changed from `$2,670.69` to `$456.24`

---
## Summary by cubic
Fixes `Jcode` token accounting in `tokscale-core` to stop double-charging cached input for OpenAI/OpenRouter routes, while keeping Anthropic-style split accounting. Also bumps the `Jcode` parser version to invalidate stale caches.

- **Bug Fixes**
  - For OpenAI/OpenRouter-style reports, subtract `cache_read_input_tokens` from `input_tokens` before pricing/aggregation.
  - Preserve Anthropic-style split when `cache_creation_input_tokens` is present (including zero).
  - Use message-local usage shape to handle mid-session route switches.

- **Migration**
  - `Jcode` parser version bumped to v6; existing cached parse results are invalidated automatically.

<sup>Written for commit 4b60cc136de7106d4b99261be5de098faff1a9d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -805,7 +805,9 @@ fn parser_version(client: ClientId) -> u32 {
         // v4->v5: jcode's assistant-message timestamp is now back-calculated
         // to the turn start (timestamp - tool_duration_ms) instead of using
         // the recorded (end-anchored) timestamp directly. Follow-up to #890.
-        ClientId::Jcode => 5,
+        // v5->v6: OpenAI-style Jcode usage now removes cache-read overlap from
+        // input_tokens before pricing and aggregation.
+        ClientId::Jcode => 6,
         ClientId::Copilot => 5,
         // Pi subagent sessions now derive agent attribution from session_info
         // names; version-1 caches carry those messages without agent metadata.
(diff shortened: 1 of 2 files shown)


junhoyeo (approved):
Thanks!

Issue #334, opened 2026-03-16
Regression: `tokscale submit` still fails with future-date validation on @latest


ShivamB25:
## Description
I'm still hitting the same validation failure that was fixed in #318/#319, but with `bunx tokscale@latest submit`.

## Command
```bash
bunx tokscale@latest submit
```

## Actual Output
```text
Error: Validation failed
  - Date range extends into the future: 2026-03-17
  - Future date found in contributions: 2026-03-17
```

## Why this looks like a regression
- #318 reported this exact failure mode and was closed by #319 (1-day timezone buffer).
- I'm seeing the same error pattern again while using `@latest`.

## Environment
- Timezone: Asia/Calcutta (UTC+5:30)
- Local date when running: 2026-03-17
- Command: `bunx tokscale@latest submit`
- `npm view tokscale version`: `2.0.12`

## Notes
When this fails, it usually means one of these is happening:
1. A validation path is still comparing against strict UTC `today` without the intended buffer, or
2. The API deployment handling submit validation is behind the package release that included #319.

## Request
Could you check whether the submit validation path currently deployed still includes the #319 buffer logic in `packages/frontend/src/lib/validation/submission.ts`, and whether there's any alternate validation path for contributions/dateRange that bypasses it?

Happy to test a canary build or run with extra debug output if useful.

Issue #318, opened 2026-03-12
[Bug] submit rejects today's date in UTC+ timezones ("extends into the future")


cuipengfei:
## Description

\`tokscale submit\` rejects submissions containing today's date when run from UTC+ timezones (e.g. CST/UTC+8) during early local hours.

## Steps to Reproduce

1. Be in a UTC+ timezone (e.g. CST, UTC+8)
2. Run \`tokscale submit\` when local time is past midnight but UTC date hasn't rolled over yet (e.g. local 01:46 CST = 17:46 UTC previous day)

## Expected Behavior

Today's date in the user's local timezone should be accepted.

## Actual Behavior

```
Error: Validation failed
  - Date range extends into the future: 2026-03-13
  - Future date found in contributions: 2026-03-13
```

## Root Cause

Timezone mismatch between CLI and server validation:

- **CLI** (\`crates/tokscale-core/src/sessions/mod.rs\`) generates dates using **local timezone**:
  ```rust
  /// Convert Unix milliseconds to a local YYYY-MM-DD date string.
  fn timestamp_to_date(timestamp_ms: i64) -> String {
      timestamp_to_date_with_timezone(timestamp_ms, &chrono::Local)
  }
  ```

- **Server** (\`packages/frontend/src/lib/validation/submission.ts\`) validates against **UTC**:
  ```typescript
  // Step 2: No future dates (using UTC since DailyContribution.date is in UTC)
  const todayStr = new Date().toISOString().split("T")[0];
  ```
  The comment claims "DailyContribution.date is in UTC" but the CLI actually uses \`chrono::Local\`.

For CST (UTC+8), this creates an 8-hour daily window (00:00–08:00 local) where today's date is rejected.

## Impact

All users in UTC+ timezones are affected. The rejection window scales with timezone offset (up to 14 hours for UTC+14).

## Suggested Fix

Either:
1. **Server-side** (recommended): Add timezone buffer to the future-date check (e.g. allow up to UTC+14 offset), or accept dates ≤ tomorrow UTC
2. **CLI-side**: Use \`chrono::Utc\` instead of \`chrono::Local\` in \`timestamp_to_date\` — but this would make TUI display dates differ from submitted dates

## Environment

- tokscale v2.0.10
- OS: Linux (WSL2)
- Timezone: CST (UTC+8)

---

*This issue was filed by [Omo](https://github.com/code-yeongyu/oh-my-openagent) (AI agent) on behalf of the user after automated root cause analysis.*

[...]

PR #952, opened 2026-07-23
feat(frontend): add display timezone preference for profile charts

junhoyeo:
## Problem

The CLI buckets daily contributions by the user's **local** timezone, so a KST user's early-morning usage is submitted under a date that is still "tomorrow" in UTC. The web dashboard computes profile chart ranges against UTC "today" (`toUtcDateString` in `src/lib/publicProfileData.ts`), so freshly submitted local-today data is clipped from the contribution graph until UTC midnight rolls over. Related local-timezone date context: #318, #334 (CLI-side submit validation, already fixed).

## What changed

Display-only, client-side fix — no DB or API changes:

- **`src/lib/useSettings.ts`** — new `timezone` setting in `localStorage["tokscale-settings"]`: `"auto"` (default) or an IANA zone, validated with `Intl.DateTimeFormat` in the parser (invalid values fall back to `"auto"`).
- **`src/lib/timezone.ts`** (new) — resolves the effective timezone (`"auto"` → browser zone), computes "today" as `YYYY-MM-DD` in that zone via `Intl.DateTimeFormat` parts (never `toISOString()`, which is UTC), extends a range end to local today without ever shortening it, and anchors calendar dates to a zone for label formatting.
- **`/settings` page** — new "Timezone" section with an "Auto (browser)" option plus a modest grouped list of common IANA zones (~45, not the full database).
- **Profile page** (`ProfilePageClient.tsx`) — after mount, the rolling chart range end becomes `max(server end, today-in-effective-tz)`, so local-today contributions render immediately on the lifetime view. The extension is mount-gated so the first client render matches SSR, and limited to `period=all` because week/month contributions are filtered by the UTC range server-side.
- **`ProfileContributionGraph.tsx`** — the hardcoded `timeZone: "UTC"` formatters are now cached per effective timezone; calendar-date labels (month markers, cell titles, tooltips, range summary, day breakdown) are formatted at the date's midnight *in the effective zone*, so labels always name the intended calendar day (no off-by-one for negative offsets).
- **`ProfileOverview.tsx`** — "Joined" and "Updated" dates use the effective timezone after hydration (previously viewer-default / hardcoded UTC).

## Deliberately NOT changed

- **Leaderboard stays UTC** — `src/lib/leaderboard/` untouched, as is embed SVG rendering (`src/lib/embed/`).
- **No DB migrations, no API changes** — the preference lives in localStorage only.
- **Week/month profile periods** — the server filters those contributions by the UTC range (`dailyBreakdownFilter`), so a client-side extension there could only render an empty trailing day; left as-is.

## Verification

- `tsc --noEmit` passes; `eslint` reports 0 errors (15 pre-existing warnings, identical on `main`).
- `vitest run`: 663 tests pass, including 9 new ones in `__tests__/lib/timezone.test.ts` covering zone validation, today-in-zone, range extension, and calendar anchoring (incl. `Pacific/Kiritimati` +14 edge).
- `next build` succeeds.
- Manual check: with `now = 2026-07-23T16:00Z`, a Seoul viewer's range end extends `2026-07-23` → `2026-07-24` while a UTC viewer's stays put, and `2026-07-24` still labels as "Fri, Jul 24" in `America/New_York`.

---
## Summary by cubic
Adds a display timezone preference for profile timestamps and anchors all profile chart windows on the newest submitted date so local‑today contributions render consistently for everyone. This is display-only; charts and totals are driven by data on the server, not the viewer’s clock.

- **New Features**
  - New `timezone` in `useSettings` (`"auto"` or IANA zone) with a selector on `/settings`; stored in `localStorage` and validated via `lib/timezone`.
  - Profile overview timestamps (“Updated”, “Joined”) render in the effective timezone after hydration; SSR uses UTC to avoid hydration mismatches.

- **Bug Fixes**
  - Server now ends lifetime, 7d, and 30d windows at the latest submitted date (`MAX(submissions.dateEnd)`), falling back to UTC today; period queries remove the upper `lte` bound and trim results to the anchored window; intensity scaling uses only visible days.
  - Removed the no-op calendar timezone path; contribution graph labels and buckets remain stable and viewer‑timezone independent.
  - No API or DB changes; leaderboard and embeds remain UTC.

<sup>Written for commit 405cf0b6a5f320a605dc17a99937c0f926afe34e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/952?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/__tests__/api/usersProfile.test.ts
+++ b/packages/frontend/__tests__/api/usersProfile.test.ts
@@ -8,6 +8,7 @@ import {
   vi,
 } from "vitest";
 
+import { createContributionCalendar } from "../../src/components/profile/ProfileContributionGraph";
 import { expectNoNarrowedCostCast } from "../support/costCastWidths";
 
 const mockState = vi.hoisted(() => {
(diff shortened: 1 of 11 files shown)


junhoyeo:
Adversarial review (codex plus an independent pass, with exhaustive sweeps rather than argument). **Verdict: needs rework, not patching.** CI being green is not in question — none of this is a build or lint issue.

## The finding that changes the shape of the PR

**The range extension is keyed to the *viewer's* timezone, but day attribution is keyed to the *submitter's*.** `ProfilePageClient.tsx:118-128` + `timezone.ts:149-157`.

- **Fails to fire when it should.** Owner in `Asia/Seoul`, viewer in `America/Los_Angeles`, now `2026-07-23T16:00Z`. The owner's CLI submitted `date: "2026-07-24"`. The viewer's local today is `2026-07-23`, so no extension happens and the owner's newest day stays clipped. *The PR does not fix the reported bug for that viewer.*
- **Fires when it shouldn't.** Owner in LA with no `07-24` row, viewer in Seoul, same instant → an empty trailing column appended to a stranger's chart.

The correct anchor needs no timezone at all, and **it is already in the payload and discarded**: `publicProfileData.ts:154` computes `latestDate: MAX(submissions.dateEnd)` and ships it as `dateRange` at `:620`; `getProfileChartRange` (`ProfilePageClient.tsx:75-80`) throws it away for `period=all` by returning `chartRange ?? apiRange`. Anchoring the display end to the data's own max date is viewer-independent, submitter-correct, and fixes both directions.

That makes the viewer-timezone preference the wrong axis for this problem.

## Blocking

**B1 — a token total moves when the user changes a setting labelled "display".** `ProfilePageClient.tsx:339-340` → `ProfileUsageChart.tsx:914-915, 968-971`. "Range total tokens" is computed over `[rangeStart, rangeEnd]`, and `rangeEnd` is now timezone-extended. With a `2026-07-24` row present (accepted — `validation/submission.ts:342-350` allows UTC today + 2 days, and the `period=all` query has no date filter), switching Seoul → Los Angeles drops the headline number by that row's tokens, and switching back restores it. On a product whose currency is token totals, that is a correctness defect.

**B2 — the graph's active-day count contradicts ProfileOverview on the same screen.** `ProfileContributionGraph.tsx:707` vs `ProfileOverview.tsx:394-399` fed by `publicProfileData.ts:616`. `stats.activeDays` is always server-scoped to the UTC range; the graph now counts over the extended range. ProfileOverview renders above the tab bar and the graph is the default tab, so both are visible at once. On `main` they agree.

## Should-fix

**S1 — roughly 150 of the 548 added lines are a provable no-op.** `timezone.ts:161-228` and the `timeZone` threading through `ProfileContributionGraph`. Rendering the deleted formatters against the new path across **1,832,094 label renderings** (3 formats × 418 zones × every day 2024→2027) produced **zero differences**. It is true by construction: `calendarInstantInTimeZone` anchors to local midnight of the same Y-M-D, so a zone formatter necessarily prints the Y-M-D the UTC formatter printed. The PR body's "no off-by-one for negative offsets" describes a bug that could not occur — the replaced formatters were already pinned to `timeZone: "UTC"` with exact UTC-midnight inputs from `parseUtcDate`.

**S2 — in the only two inputs where the new path differs, it is wrong.** Sweeping 2000–2030 × 418 zones found exactly two divergences, both regressions:

```
Pacific/Apia     2011-12-30: OLD="Friday, December 30, 2011"  NEW="Saturday, December 31, 2011"
Pacific/Fakaofo  2011-12-30: OLD="Friday, December 30, 2011"  NEW="Saturday, December 31, 2011"
```

The 2011 dateline move skipped that civil date. Reachable via a backfilled 2011 profile viewed from Apia, since `createContributionRangeOptions` emits a year option for every year present. Absurdly narrow — but it means S1's machinery is a no-op everywhere it is right and wrong everywhere it differs.

**S4 — settings copy.** `SettingsClient.tsx:962-966` says the setting is "used to display dates on profile pages, including the contribution graph." Given S1 it has *zero* effect on any graph date label; its only graph effects are whether the newest column carries data and the two range-scoped totals riding on it. The copy should say day attribution comes from the submitting machine, and the name wants coordinating with #960 Layer 3's `meta.timezone` before it ships — otherwise `/settings` ends up with a "Timezone" control that does not affect attribution sitting beside a resync control that does.

## Credit where due: the timezone library itself is correct

Sweeping all 418 runtime IANA zones × every day 2000-01-01→2030-12-31 (~4.7M cases) asserting round-trip identity found only the two nonexistent civil dates above. Verified clean: skipped midnight (`America/Santiago`), ambiguous midnight (`Asia/Beirut`), midnight-shifting DST (`Africa/Cairo`), 30-minute DST (`Australia/Lord_Howe`), `Pacific/Chatham +12:45`, `Antarctica/Troll`. It also survives an engine that ignores `hourCycle: "h23"` (pre-2020 Safari), because the second pass's date-key check rescues it. This holds structurally, not by luck.

## Recommendation

Anchor the display range to `dateRange.end` from the payload and drop the viewer-timezone dependency from the range logic entirely. That fixes the actual reported bug in both directions, removes B1 and B2 at the root rather than patching two symptoms, and makes S1/S2 moot by deleting the machinery.

If a display-timezone preference is still wanted afterwards, it should land separately, scoped to things it genuinely affects, with copy that says what it does not do.

Relates to #960 — the confusion this PR risks is precisely the display-vs-scan timezone distinction that issue turns on.

junhoyeo:
Pushed four commits (`b3bffc0b..405cf0b6`) taking the review's root-cause route.

## The fix went server-side, not client-side

The review recommended anchoring in `getProfileChartRange` (`ProfilePageClient.tsx:75-80`) by preferring `dateRange.end`. **That is insufficient, and the implementation proved it rather than assuming it.** The server computes `stats.activeDays` from `scopedContributions`, filtered at `publicProfileData.ts:479` against a `chartRange` that still ended at UTC today. A client-only anchor makes the graph count `2026-07-24` while ProfileOverview does not — so **B2 would be live in exactly the scenario being fixed**.

Demonstrated, not argued: with the server fix stashed and the range assertions removed, the calendar assertion failed with `expected '2026-07-23' to be '2026-07-24'`.

So `getRollingProfileDateRange` now takes the newest submitted date and ends the window at `latest > utcToday ? latest : utcToday`, with `chartRange` moved after the stats query so it can see `MAX(submissions.dateEnd)`. The viewer's clock is no longer an input anywhere in the range path. Fixing it there makes the client-side anchor a provable no-op, so `getProfileChartRange` needed no change at all.

**B1 and B2 dissolve rather than get patched.** `ProfileUsageChart` was not touched — its `rangeEnd` input simply stopped varying. And because `chartRange` is resolved before `scopedContributions` filters on it, the graph and ProfileOverview are scoped to the same window by construction.

## The no-op machinery is gone

S1 was verified before anything was deleted — the 1,832,094-comparison sweep reproduced exactly (418 zones × 3 formats × 2024–2027, **zero differences**), and the 2011 dateline window reproduced S2's four divergences (`Pacific/Apia` and `Pacific/Fakaofo`, `2011-12-30`, both formats). Then 125 lines came out of `timezone.ts`, and `ProfileContributionGraph.tsx` is now **byte-identical to `origin/main`** — `git diff origin/main HEAD -- <that file>` prints nothing.

## Answering the question I flagged

After the anchor moved and the dead code went, does the preference still do anything? **Yes, but much less than the copy claimed** — it affects the "Updated" and "Joined" timestamps, which are absolute instants, and nothing else. The settings copy now names exactly those two, and states plainly that it does not move usage between days, that the scanning machine's clock decides day attribution, and that the graph, daily totals and leaderboard are unaffected. `timezone.ts`, `ProfileOverview` and `useSettings` docs all now say "absolute instants only, never calendar buckets", pointing at #960 for the scan-timezone side.

That distinction is the thing #960 turns on, so it was worth getting into the copy rather than leaving it implied.

## Verification

- `npx vitest run` — **666 passed / 72 files** (I re-ran this on the pushed tree, not just in review)
- `tsc --noEmit` exit 0; `npm run lint` 0 errors (15 pre-existing warnings, all in untouched files)
- Each commit verified in isolation, not only at the tip
- Rebase probe onto current `main` (which now carries #955): 5 commits, zero conflicts, 663 tests green. Probe branch deleted.
- Hardening: `parseDateKey()` added so a regex-matching but out-of-range anchor (`2026-13-45`) falls back to UTC today instead of feeding `NaN` into `toUtcDateString` and 500ing the profile route

Issue #964, opened 2026-07-26
Sessions tab wide layout is over-subscribed: needs 161 columns, activates at 80


junhoyeo:
Split out of #956, which exposed this rather than caused it. Same framing as #960: measure the actual constraint, then fix the defect, instead of picking a threshold that papers over it.

## The measurement

`crates/tokscale-cli/src/tui/ui/sessions.rs`. The wide layout activates at `!app.is_narrow()`, i.e. **80 columns**. Its constraints, at natural width:

```
Session 20 (Min) + Client 12 + Turn 5 + Msgs 5
  + Input 10 + Output 10 + Cache R 10 + Cache W 10 + Cache× 8
  + Total 10 + Cost 10 + Cost/1M 10 + Duration 9 + Last Active 17   =  146
  + 13 inter-column separators                                      =  159
  + 2 block borders                                                 =  161
```

| layout | cells | terminal columns needed |
|---|---|---|
| wide, turn data, no Model | 159 | **161** |
| wide, turn data, with Model (18) | 178 | **180** |
| wide, no turn data, no Model | 154 | **156** |
| **actually activates at** | | **80** |

So for every width from 80 to 160 the ratatui solver shrinks every column to fit. Headers degrade to stubs and the sort indicator gets clipped off whichever column carries it. This has been true since long before #956 — it is simply invisible because nobody counted.

## Why this matters now

#956 adds a Model column and gates it on the budget above, so it appears only at 180+. That gate is *correct* — it is derived rather than hand-picked, and every sort indicator is legible whenever Model shows. But its premise is "all ten metric columns render at natural width", a state the table almost never reaches. The practical result is a feature that ships and effectively never renders.

The gate is not the problem. It is a correct response to a broken budget.

## What a fix looks like

The mechanism already exists — `has_turn_data` and `show_model` both conditionally include a column across the header vector, the row cells, the constraint vector, and the sort indices. Generalising that to a priority-ordered set, and admitting columns only while the budget allows, would make **every** threshold a consequence of measured widths rather than a constant:

- at 80: the columns that actually fit, rendered legibly
- at 120: more
- at 180: everything, Model included

That also deletes `MODEL_COLUMN_MIN_CHARS` and `wide_layout_fixed_width` as special cases — Model stops being a gated exception and becomes one entry in the priority order.

[...]

PR #965, opened 2026-07-26
docs(tui): propose a width-budgeted column set for the Sessions tab

junhoyeo:
Design proposal for #964. **Documentation only — no behaviour change.** `docs/sessions-column-budget.md`.

## Why this is worse than #964 says

I filed #964 describing header labels degrading to stubs. That undersold it. Driving the existing `render_body` test helper at several widths shows the **magnitude suffix is silently clipped from values**:

```
=== width 80 ===
|│Session              Cli Turn Msg Inpu Out Cach Cac Cach Tot Cost Cos Dura Las│|
|│abc-123              Ope 2    4   876K 234 0    0   0.0x 1.1 $12. $11 1h 0 202│|

=== width 161 ===
|│Session              Client       Turn  Msgs  Input      Output     Cache R    …│|
|│abc-123              OpenCode     2     4     876K       234K       0          …│|
```

| cell | true value | rendered at 80 |
|---|---|---|
| Output | 234K | `234` |
| Total | 1.1M | `1.1` |
| Cost | $12.35 | `$12.` |

`234` and `1.1` are plausible-looking numbers that are wrong by 1000×, with nothing to signal truncation. ratatui clips cell content without an ellipsis. That is a correctness problem, not a cosmetic one, and it is the strongest argument for fixing the budget rather than living with it.

The same probe independently confirms the arithmetic in #964: at **161** every column renders in full without Model, and at **180** Model appears.

## What the document proposes

One `SessionColumn` enum whose `header` / `natural` / `constraint` / `sort_field` / `available` / `cell` methods are exhaustive matches with **no `_` arm**, so adding a column fails to compile until every sync point exists. Columns are admitted in priority order while the width budget allows, with related columns grouped all-or-nothing (`Input`+`Output`, `Cache R`+`Cache W`) because `Input` without `Output` reads as a total.

Two corrections to my framing in #964 that came out of drafting it:

- There are **five** sync points, not four. `wide_layout_fixed_width` holds a second copy of the constraint sum and gets deleted too. Sort indices become `chosen.iter().position(…)`, which cannot be wrong by construction.
- #964 asks which single column goes first at 100 columns. The honest answer is *all of them, plus Input/Output, plus Model* — the core six already need 69 cells and `Last Active` alone adds 18. A hundred columns holds seven columns, not thirteen.

Under the recommended order, Model lands at **106** columns instead of 180.

## What it asks you to decide

The document is explicit that these are decisions, not implementation details:

1. **Priority order** — recommended `Last Active → Model → Input+Output → Cache R+Cache W → Duration → Cache× → Cost/1M`, ranked by navigational vs. diagnostic value.
2. **Slack distribution** — which column absorbs surplus, and therefore which one sawtooths at the thresholds.
3. **`has_model_data`** — add an availability gate symmetric with `has_turn_data`, or let Model show em-dashes? Not independent of (1).
4. **The 120-column regression** — trading six stub columns for eight complete ones makes something visibly *worse* for some users. This is the honest cost and the reason the priority order is a decision rather than a detail.

## Process notes

Drafted, adversarially critiqued, and revised; the critique pass is what produced the "arrays are not compiler-checked" section, which closes a real gap in the first draft's central no-drift claim.

I verified every number in the document myself against `sessions.rs` and against the rendered output above rather than taking the draft's word for it — #964 exists because I previously proposed a threshold without doing the arithmetic and was off by ~60 columns, so a design document repeating an unverified number would be worse than none.

No code changes, so nothing here can regress the narrow or very-narrow layouts. Those are called out in the document as invariants any implementation must preserve.

---
## Summary by cubic
Design doc proposing a width‑budgeted, priority‑based column set for the Sessions tab to prevent clipped values and keep headers/arrows legible across 80–160 columns. Docs only; adds `docs/sessions-column-budget.md` for #964 with a compiler‑checked `SessionColumn` design, priority admission + slack rules that show Model at ~106 cols, and a corrected full‑table width of 183 (Client widened to 15).

- **Migration**
  - Use a single `SessionColumn` descriptor to derive headers, natural widths, availability, sort field, and cells; derive constraints via `WideLayout::constraint(&WideCtx)` so resolved widths flow to both constraints and cell formatters (prevents Model-width divergence).
  - Admit grouped columns by priority; stop at the first non‑fitting group; let Session absorb slack to 40, then Model to 36; recommended order: Last Active → Model → Input+Output → Cache R+Cache W → Duration → Cache× → Cost/1M.
  - Remove `wide_layout_fixed_width`, manual sort indices, and gate helpers; compute sort arrows by column identity.
  - Add tests: full width sweep for label/arrow legibility and monotonic header sets; literal threshold table; `ALL`/`WIDE_ORDER`/`WIDE_PRIORITY` permutation check; per‑formatter fit checks; add a compact branch for `format_cost_per_million` to avoid overflow.

<sup>Written for commit 7221fefb1f9200cac7885aa27396de25dba2a4ae. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/965?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/sessions-column-budget.md
+++ b/docs/sessions-column-budget.md
@@ -261,40 +261,48 @@ constraint is therefore *derived*, never written:
 
 ```rust
 // `natural()` is the ADMISSION width — what a column must be given to be let
-// in at all. `allocated()` is what it actually receives after slack
-// distribution, and it is the only width the renderer may use. For every fixed
-// column the two are equal; for Session and Model they diverge, which is the
-// entire point of slack distribution.
-fn allocated(c: SessionColumn, layout: &WideLayout) -> u16 {
-    match c {
-        SessionColumn::Session => layout.session_width,
-        SessionColumn::Model => layout.model_width,
-        other => other.natural(&layout.ctx),
-    }
-}
-
-fn constraint(c: SessionColumn, layout: &WideLayout) -> Constraint {
-    if c.grows() {
-        Constraint::Min(allocated(c, layout))
-    } else {
-        Constraint::Length(allocated(c, layout))
+// in at all. What it actually RECEIVES after slack distribution is the only
+// width the renderer may use. For every fixed column the two are equal; for
+// Session and Model they diverge, which is the entire point of slack
+// distribution, so the resolved layout has to be in scope here.
+impl WideLayout {
+    fn constraint(&self, c: SessionColumn, ctx: &WideCtx) -> Constraint {
+        let width = match c {
+            // The allocated width, not the natural one — this is the whole point.
+            SessionColumn::Model => self.model_width,
+            other => other.natural(ctx),
+        };
+        if c.grows() { Constraint::Min(width) } else { Constraint::Length(width) }
     }
 }
 ```
 
-**Both `constraint()` and `cell()` take `&WideLayout`, and neither may take
-`&WideCtx`.** An earlier revision of this document derived `constraint` from
-`ctx` while `cell` took `layout`, which is a real bug that two independent
-reviewers caught: `Model.grows()` is `false`, so its constraint came out as
+`Session` needs no arm: it is the sole `Min` column, so ratatui hands it every
+cell slack distribution did not give `Model`. `self.session_width` exists for
+`cell()` to truncate the title against, and must equal what the solver actually
+allocates or the title is clipped with no ellipsis.
+
+**`constraint()` must have the resolved layout in scope, not only `WideCtx`.**
+An earlier revision of this document derived `constraint` from `ctx` alone
+while `cell` took `layout`, which is a real bug two independent reviewers
+caught: `Model.grows()` is `false`, so its constraint came out as
 `Length(natural(Model))` — always 18 — while `cell()` formatted content to
 `layout.model_width`, up to 36. ratatui allocates what the constraint asks for,
 so every cell wider than 18 was silently clipped and the whole `model_extra`
-computation was dead. `Session` survived only because it is `Constraint::Min`
-and absorbs surplus regardless.
+computation was dead code. `Session` survived only because it is
+`Constraint::Min` and absorbs surplus regardless.
+
+The first attempt at this correction was itself wrong, and a reviewer caught
+that too: it introduced a free function reading `layout.ctx`, a field
+`WideLayout` does not have, so it would not compile. Making `constraint` a
+method on `WideLayout` that *also* takes `&WideCtx` is what the implementation
+in #966 does, and it is the right shape — the layout supplies the widths
+admission computed, `ctx` supplies the fixed naturals, and neither is
+reachable without the other.
 
 The lesson generalises: any width that admission *computes* must flow to the
-constraint, not just to the formatter. Threading one `&WideLayout` through both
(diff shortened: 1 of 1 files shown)


junhoyeo:
Both findings are real and both are fixed in `0002fa1e`. Codex and cubic landed on each of them independently, which is a good signal — they are the two places where the document argued for a property it did not actually have.

## 1. The constraint/layout split (codex P1, cubic P2)

Confirmed by reading the document's own derivations against each other:

```rust
let widths = chosen.iter().map(|c| constraint(*c, &ctx));            // &WideCtx
let cells  = chosen.iter().map(|c| c.cell(session, app, &layout));   // &WideLayout
```

and the document explicitly keeps `model_width` off `WideCtx` ("both are outputs of admission, which is why neither appears in `WideCtx`"). Since `Model.grows()` is `false`, its constraint resolved to `Length(natural(Model))` — always 18 — while `cell()` formatted to `layout.model_width`, up to 36. ratatui allocates what the constraint asks for, so everything past 18 was clipped and the entire `model_extra` computation was dead. `Session` masked the bug because `Constraint::Min` absorbs surplus whether or not the constraint names it.

Fixed by making both derivations take `&WideLayout` through a shared `allocated()`:

```rust
fn allocated(c: SessionColumn, layout: &WideLayout) -> u16 {
    match c {
        SessionColumn::Session => layout.session_width,
        SessionColumn::Model   => layout.model_width,
        other                  => other.natural(&layout.ctx),
    }
}
```

so a width that admission *computes* cannot reach the formatter without also reaching the constraint. Worth stating plainly: this was the exact class of budget-vs-layout divergence the design exists to eliminate, sitting inside the design. There is now a `Directive:` trailer on the commit saying neither function may ever take `&WideCtx` again.

## 2. `format_cost_per_million` overflows its own column (codex P2, cubic P2)

Also confirmed. `widgets.rs:54-60` is a plain `format!("${:.2}", per_m)` with no compact branch, so `format_cost_per_million(100.0, 100)` produces `$1000000.00` — 11 cells clipped into 10 **at 180 columns, where every column is admitted at natural width**.

cubic is right that the document contradicted itself: its non-goals already noted the formatter is unbounded, while the testing section claimed truncation was "prevented structurally". That claim is true of column *squeeze* and false of a formatter outgrowing the width it was sized for — which is the same plausible-looking-wrong-number failure this whole document is about.

The testing section now requires two things rather than waving it off: a per-formatter assertion that each column's output fits its `natural()` width under adversarial inputs, and a compact branch for `format_cost_per_million`. Recommended over widening Cost/1M to 11, which would raise every threshold that admits it; the thresholds here are already tight and `$1.00M/M` reads fine. Added to the decisions list as item 5 — the fix is required, only its form is a choice.

Both were caught before any of this shipped as code, which is what the document was for.

Issue #970, opened 2026-07-28
fix(autosubmit): verify macOS launchd activation before reporting success


Mathews-Tom:
## Problem

On macOS, `tokscale autosubmit enable` can print a `launchctl load` diagnostic such as `Load failed: 5: Input/output error` and then unconditionally report `Autosubmit enabled`. A local reproduction on `tokscale 4.0.11` found that the generated `ai.tokscale.autosubmit` LaunchAgent was nevertheless valid and loaded in `gui/<uid>`; the scheduler integration cannot distinguish that state from an actual failed installation.

`autosubmit.rs` currently:

- generates legacy `launchctl load`/`unload` commands instead of domain-explicit `bootstrap gui/<uid>`/`bootout`;
- calls `Command::status()`, which discards stdout/stderr and accepts any zero exit status without verifying the service target via `launchctl print gui/<uid>/ai.tokscale.autosubmit`;
- writes the LaunchAgent with `RunAtLoad=true` before persisting the enabled autosubmit settings, so the immediate run can observe the old disabled configuration and exit without a submission; and
- embeds `current_exe()` in a durable plist. With `bunx`, that can be a workspace or temporary `node_modules` path which disappears after the command completes or the installation changes.

Do not direct users to run `launchctl bootstrap` as root. This is a per-user LaunchAgent and belongs in the logged-in user domain.

## Expected behavior

`autosubmit enable` must persist its settings before activation, install the service into `gui/<uid>` with the modern launchctl interface, and only report success after it verifies that `launchctl print gui/<uid>/ai.tokscale.autosubmit` succeeds. Activation failures must return the captured diagnostic and leave no enabled configuration that claims a scheduler is installed.

The long-lived scheduler must not retain a transient executable path from `bunx` or a local workspace. Resolve a stable installed launcher, or reject activation with an actionable persistent-install requirement.

## Reproduction evidence

```text
bunx tokscale autosubmit enable
Load failed: 5: Input/output error
Try running `launchctl bootstrap` as root for richer errors.
Autosubmit enabled: every 1440 minutes via launchd.
```

After this command, the generated plist passed `plutil -lint`, and `launchctl print gui/501/ai.tokscale.autosubmit` showed a loaded agent with `run interval = 86400 seconds`, `runs = 1`, and `last exit code = 0`. `autosubmit status --json` still showed `lastRunAtMs: null`, consistent with the `RunAtLoad` ordering race.

## Acceptance criteria

[...]

PR #973, opened 2026-07-29
fix(autosubmit): report when the scheduled copy is a stale build

### LINKED ISSUE #972: autosubmit: the managed executable copy does not follow upgrades
Follow-up to #971 (merged as `3ce29f97`). Strictly additive — this is not a regression, and #971 was right to land as-is.

## What happens

#971 copies the running binary into the config root and renders every scheduler backend against that copy (`crates/tokscale-cli/src/commands/autosubmit.rs:241`). The copy is written only inside the enable path.

`AutosubmitSettings.managed_executable` (`crates/tokscale-cli/src/tui/settings.rs:75`) records a **path**, not a version. Nothing compares it against the running binary, and neither `status()` nor `status_output()` mentions it.

So after `npm update -g tokscale` — or brew, or replacing the binary directly — the scheduled job keeps running the previously copied build, indefinitely and silently, until the user happens to re-run `autosubmit enable`.

The failure mode is that it is invisible, more than that it happens. A user who upgrades specifically to pick up a parser fix has no signal that their scheduled submissions are still using the old parser. `main` currently carries `parser_version` bumps for Kimi (#967) and Hermes (#961), so the next release is exactly such a case.

Note this is *not* introduced by #971: `main` already rendered the spec against `current_exe()` before it. The difference is that the old path pointed at the installed binary, which an upgrade replaces in place; the new one points at a copy, which it does not.

## Why the fix is cheaper than it looks

**On Unix, refreshing requires no scheduler interaction at all.** The plist and crontab point at a fixed path, so atomically replacing the file at that path is sufficient — the next scheduled invocation runs the new binary with no `bootout`/`bootstrap` cycle and no window in which the user is unscheduled. `tokscale_core::fs_atomic::replace_file` already does this and #971 already uses it. Replacing a running binary's inode is safe: an in-flight scheduled run keeps executing the old inode and completes normally.

**Windows is the exception and sets the scope.** Running executables are locked, which is why #971 already writes versioned filenames (`versioned_windows_managed_executable_path` at `:1298`, `is_windows_managed_executable_name` at `:1367`). A refresh there must write a new versioned file, re-render the scheduler spec to point at it, and reap the old one — a genuine scheduler transaction with #971's rollback concerns. The implementation should not pretend the two platforms share a code path.

## Proposed change

1. **Record the version.** Add `managed_executable_version: Option<String>` to `AutosubmitSettings`. Every field there is `#[serde(default)]`, so existing configs deserialize as `None`, which reads correctly as "unknown — refresh at the next opportunity". Set it beside `managed_executable` at `:241`, clear it beside `:376`.
2. **Compare on `differs`, not `newer`.** No semver parsing, and a deliberate downgrade should propagate — the user pinned that version on purpose.
3. **Refresh only within `autosubmit` subcommands.** A plain `tokscale report` must not write into the autosubmit directory. This also bounds the clobbering risk from (2) to contexts where the user is explicitly acting on autosubmit.
4. **Surface drift in both status paths.** `status()` and `status_output()` are separate human and JSON surfaces; drift must appear in both or scripted checks miss it.
5. **Warn from `tokscale submit`.** Warn only, never write — it is the interactive twin of the scheduled job and where a user would notice.

## Edge cases that need tests

- **Self-comparison.** Running the managed copy directly makes `current_exe()` canonicalize to `managed_executable`, so versions always match. Detect and skip, rather than reporting a spurious "up to date".
- **A refresh must not touch launchd on Unix**, so a concurrent scheduled run cannot be de-scheduled by one.
- **`None` version on an existing config** refreshes once and then settles — not on every invocation.
- **Windows versioned refresh** leaves exactly one managed executable, a spec pointing at it, and no orphans.

## Deliberately rejected

- **Self-refresh from the scheduled run.** The copy replacing itself is dangerous on Unix and impossible on Windows, and that process has no reliable handle on the installed binary — which is the whole reason #970 asked for a copy.
- **Hashing instead of versions.** Catches same-version rebuilds that only developers hit, and costs a full read of a large binary on every autosubmit command.
- **Hardlinking instead of copying.** Package managers replace rather than modify in place, so the link breaks on upgrade anyway — all the fragility, none of the benefit.

junhoyeo:
Closes #972. Follow-up to #971 (`3ce29f97`), and strictly additive — #971 was right to land as-is.

## The problem

#971 renders every scheduler backend against a copy of the binary under the config root, and only `enable` writes that copy. Replacing the installed binary — `npm update -g tokscale`, brew, dropping one in — does not touch it, so the scheduled job keeps submitting with the old build until the user happens to re-run `enable`.

**The failure is that it is invisible, more than that it happens.** Settings recorded the copy's path but not its version, and neither status surface mentioned it. Someone upgrading specifically to pick up a parser fix had no signal that their scheduled submissions still used the old parser — and `main` currently carries `parser_version` bumps for Kimi (#967) and Hermes (#961), so the next release is exactly that case.

Not a regression from #971: `main` already rendered against `current_exe()`. The difference is that the old path was the installed binary, which an upgrade replaces, and the new one is a copy, which it does not.

## What this does

Records the version beside the path and reports drift in both status surfaces.

```
$ tokscale autosubmit status
Autosubmit is enabled.
  Interval: 60 minutes
  Scheduler: launchd
  Clients: default submit clients
  Scheduled binary: 4.6.1 (this build is 4.7.0)
    The scheduler runs its own copy, which upgrades do not replace. Re-run `tokscale autosubmit enable` to refresh it.

$ tokscale autosubmit status --json
{"managedExecutable": "…/autosubmit/tokscale", "managedExecutableVersion": "4.6.1", "managedExecutableStale": true}
```

Both, because scripted health checks read the boolean and would miss a human-only warning.

## Detection only, deliberately

Refreshing stays manual. On Unix it would be an atomic replace at a fixed path — no scheduler interaction at all, since the plist already points there. On **Windows** the copy is versioned to dodge the executable lock (`versioned_windows_managed_executable_path`), so a refresh means writing a new file, re-rendering the spec, and reaping the old one — a scheduler transaction carrying #971's rollback concerns. Bundling those two would hide the asymmetry. #972 has the design if you want it next.

## Decisions worth reviewing

- **Any difference counts, including a downgrade.** No semver parsing, and the question is what the scheduler is *running*, not which build is newer — someone who pinned an older version should see that reflected.
- **A missing version reads as drift, not as current.** It predates the field, so the build is genuinely unknown; assuming current would under-report. Resolves on the next `enable`.
- **Running the managed copy reports no drift.** It would compare against itself and always match. A binary cannot observe its own staleness, so this returns "no drift" rather than a false all-clear.
- **Version, not hash.** Hashing catches only same-version rebuilds that developers produce, at the cost of reading the whole binary on every status call.

## A latent test race this surfaced

Adding two config-directory tests made the autosubmit suite fail **differently on every run**. `EnvVarGuard` had no lock, so every test redirecting `TOKSCALE_CONFIG_DIR` raced every other — one saves settings while another's guard restores the variable underneath it. `main` passed only through scheduling luck; I confirmed it is clean there across three runs, and that my branch passed 40/40 single-threaded while failing varied subsets in parallel.

Fixed by holding a mutex inside the guard, so every existing caller is serialised without touching a single test body. No test takes two guards at once, so it cannot deadlock, and poisoning is recovered because `Drop` restores the variable regardless. Five consecutive parallel runs now pass 40/40.

## Verification

- `cargo test -p tokscale-cli` — 964 bin + 133 integration, 0 failed (8 new)
- `cargo clippy -p tokscale-cli --all-targets -- -D warnings` exit 0, `cargo fmt --check` clean
- **End-to-end against the built binary**, not just unit tests: `enable` records `4.7.0`, an edited `4.6.1` surfaces in both status paths with `managedExecutableStale: true`

`--all-targets` because the usual invocation omits it and therefore never lints `#[cfg(test)]` code — that gap let a real violation through on #966.

---
## Summary by cubic
Detects and reports when autosubmit’s scheduled binary is stale, and now prints a safe `enable` command that preserves your config. The scheduled build version is stamped only after the scheduler installs, and status/JSON clearly surface drift.

- **Bug Fixes**
  - `autosubmit status` shows scheduled vs current and prints an exact refresh command that keeps interval, clients, date filters, and scheduler. `--json` adds `managedExecutableStale` plus `managedExecutable` and `managedExecutableVersion`.
  - Version is recorded only after scheduler installation; any version difference (including downgrades) or a missing version counts as drift; running the managed copy reports no drift.
  - `autosubmit disable` clears the recorded version; if the post-install save fails, we warn and status reports drift until the next `enable`.
  - Serialized `TOKSCALE_CONFIG_DIR` tests using `serial_test`’s `#[serial]` across modules; removed duplicate `#[serial]` attributes to avoid double-lock deadlocks.

- **Migration**
  - If status reports a stale scheduled binary, run the printed `tokscale autosubmit enable ...` command to refresh the managed copy.

<sup>Written for commit 26005e9a85cc56737663527ced1ede04f05e3b02. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/973?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/autosubmit.rs
+++ b/crates/tokscale-cli/src/commands/autosubmit.rs
@@ -249,9 +249,14 @@ where
     };
     let mut next_autosubmit = next_autosubmit;
     next_autosubmit.managed_executable = Some(exe.to_string_lossy().into_owned());
-    // Stamped from the build doing the copying, not read back off the file, so
-    // it describes what the scheduler will actually run.
-    next_autosubmit.managed_executable_version = Some(RUNNING_VERSION.to_string());
+    // The version is deliberately NOT stamped here. This save happens before the
+    // scheduler is installed, and on Windows `exe` is a freshly versioned path
+    // while the task still points at the previous one. Recording the version now
+    // would mean that a re-enable killed between this save and the install below
+    // leaves settings claiming a build the scheduler is not running, and
+    // `managed_executable_is_stale` would report clean at exactly the moment it
+    // is wrong. It is stamped after installation succeeds instead; until then
+    // `None` reads as "unknown", which reports drift.
 
     let mut next_settings = previous_settings.clone();
     next_settings.autosubmit = next_autosubmit;
@@ -314,6 +319,20 @@ where
                 &mut uninstaller,
             );
         }
+    }
+
+    // Second phase of the version stamp: only now is the scheduler actually
+    // pointing at `exe`, so only now does recording its build describe reality.
+    //
+    // A failure here is not worth unwinding a correctly installed scheduler. The
+    // version stays `None`, status reports drift, and re-running `enable` clears
+    // it — the conservative direction, and self-healing.
+    next_settings.autosubmit.managed_executable_version = Some(RUNNING_VERSION.to_string());
+    if let Err(error) = next_settings.save() {
+        eprintln!(
+            "Warning: autosubmit is enabled, but the scheduled build could not be recorded: {error}\n         \
+             `tokscale autosubmit status` will report it as out of date until you re-run `enable`."
+        );
     }
 
     println!(
@@ -2456,6 +2475,50 @@ mod tests {
         assert!(
             output.managed_executable_stale,
             "scripted checks read the boolean rather than diffing versions themselves"
+        );
+    }
+
+    #[test]
+    fn enable_withholds_the_version_until_the_scheduler_is_installed() {
+        use std::cell::Cell;
+
+        // On Windows a re-enable writes a freshly versioned copy while the task
+        // still points at the previous one, so a process killed between the
+        // first settings save and the scheduler install would leave settings
+        // claiming a build the scheduler is not running -- and the stale check
+        // would report clean at exactly the moment it is wrong. Recording the
+        // version only after installation means an interrupted enable leaves
+        // `None`, which reads as unknown and reports drift.
+        let temp = TempDir::new().unwrap();
+        let _guard = EnvVarGuard::set("TOKSCALE_CONFIG_DIR", temp.path());
+        let observed = Cell::new(Some(String::new()));
+
+        enable_with_scheduler_operations(
+            enable_args(SchedulerKind::Cron),
+            |_, _, _| {
+                observed.set(
+                    crate::tui::settings::Settings::load()
+                        .autosubmit
+                        .managed_executable_version,
+                );
+                Ok(())
+            },
+            |_, _, _| Ok(()),
+        )
+        .unwrap();
+
+        assert_eq!(
+            observed.take(),
+            None,
+            "the version must not be persisted before the scheduler points at the new copy"
+        );
+        assert_eq!(
+            crate::tui::settings::Settings::load()
+                .autosubmit
+                .managed_executable_version
+                .as_deref(),
+            Some(RUNNING_VERSION),
+            "and must be persisted once installation succeeds"
         );
     }
 
--- a/crates/tokscale-cli/src/commands/autosubmit.rs
+++ b/crates/tokscale-cli/src/commands/autosubmit.rs
@@ -15,6 +15,9 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 const JOB_ID: &str = "ai.tokscale.autosubmit";
+/// Version of the build answering the current call, stamped into settings when
+/// the managed copy is written so later runs can tell whether it has drifted.
+const RUNNING_VERSION: &str = env!("CARGO_PKG_VERSION");
 const CRON_MARKER_BEGIN: &str = "# BEGIN TOKSCALE AUTOSUBMIT";
 const CRON_MARKER_END: &str = "# END TOKSCALE AUTOSUBMIT";
 const SKIP_SCHEDULER_ENV: &str = "TOKSCALE_AUTOSUBMIT_SKIP_SCHEDULER";
@@ -170,6 +173,12 @@ struct StatusOutput {
     yesterday: bool,
     week: bool,
     month: bool,
+    managed_executable: Option<String>,
+    managed_executable_version: Option<String>,
+    /// True when the scheduled copy came from a different build than the one
+    /// answering this call. Scripted health checks read this rather than
+    /// diffing the two version strings themselves.
+    managed_executable_stale: bool,
     last_run_at_ms: Option<i64>,
     last_error: Option<String>,
 }
@@ -213,6 +222,7 @@ where
         last_run_at_ms: previous_settings.autosubmit.last_run_at_ms,
         last_error: None,
         managed_executable: None,
+        managed_executable_version: None,
     };
     let source_exe =
         std::env::current_exe().context("Could not resolve current tokscale executable")?;
@@ -239,6 +249,9 @@ where
     };
     let mut next_autosubmit = next_autosubmit;
     next_autosubmit.managed_executable = Some(exe.to_string_lossy().into_owned());
+    // Stamped from the build doing the copying, not read back off the file, so
+    // it describes what the scheduler will actually run.
+    next_autosubmit.managed_executable_version = Some(RUNNING_VERSION.to_string());
 
     let mut next_settings = previous_settings.clone();
     next_settings.autosubmit = next_autosubmit;
@@ -333,6 +346,17 @@ pub fn status(json: bool) -> Result<()> {
         } else {
             println!("  Clients: default submit clients");
         }
+        if managed_executable_is_stale(&autosubmit) {
+            let scheduled = autosubmit
+                .managed_executable_version
+                .as_deref()
+                .unwrap_or("unknown");
+            println!(
+                "  Scheduled binary: {scheduled} (this build is {RUNNING_VERSION})\n    \
+                 The scheduler runs its own copy, which upgrades do not replace. \
+                 Re-run `tokscale autosubmit enable` to refresh it."
+            );
+        }
     } else {
         println!("Autosubmit is disabled.");
     }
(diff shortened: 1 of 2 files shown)


junhoyeo:
`/codex:adversarial-review` found a real defect, and it is one this PR introduced rather than an older one. Fixed in `77c1e66e`.

## The finding

Codex traced the enable sequence: the settings save is at `autosubmit.rs:258`, the scheduler install at `:301`. The first commit stamped the version at `:254` — **before** the scheduler actually points at the new copy.

On Windows a re-enable writes a freshly *versioned* managed copy while the task still points at the previous one. A process killed between those two steps leaves settings claiming a build the scheduler is not running, and `managed_executable_is_stale` compares that claim against the running build, matches, and reports **clean at exactly the moment it is wrong**.

I confirmed both the mechanism and codex's scoping. It is Windows-only: on Unix the managed path is stable and its contents were already replaced before the save, so the plist points at the same path holding the new binary and the two agree.

What makes it worth fixing rather than filing is that it is a flaw in the signal this PR adds. Before this PR there was no clean/stale claim, so nobody could be misled by one. Shipping a check that is confidently wrong in a narrow window is worse than shipping no check.

## The fix

The version is now stamped in a second save, after installation succeeds.

No new state was needed: `None` already reads as "unknown", which reports drift. So an interrupted enable now errs toward warning instead of reassuring, using semantics the first commit had already chosen. Codex proposed a general pending/verified provenance mechanism; this is the narrower version that gets the same guarantee out of the existing field, without a second field to keep in sync.

A failure of that post-install save deliberately does **not** unwind a correctly installed scheduler. The version stays `None`, status reports drift, and re-running `enable` clears it — unwinding a working scheduler over a provenance stamp would be worse than the problem it solves.

The constraint from #970 is preserved: the first save still happens before installation, so the `RunAtLoad` run sees enabled settings.

## Verification

Pinned with a test using the same installer seam as the existing ordering test — it loads settings *inside* the installer callback and asserts the version is still `None` there, then `Some` afterwards.

Mutation-checked: restoring the single-phase stamp fails it with

```
assertion `left == right` failed: the version must not be persisted before the scheduler points at the new copy
  left: Some("4.7.0")
 right: None
```

`cargo test -p tokscale-cli` — 965 bin + 133 integration, 0 failed. `cargo clippy --all-targets -- -D warnings` exit 0, `cargo fmt --check` exit 0.

junhoyeo:
Both findings were valid, and both were flaws in this PR rather than pre-existing. Fixed in `5e58af0e`.

## The refresh advice was actively harmful

This is the one worth dwelling on. My status message said *"Re-run `tokscale autosubmit enable` to refresh it"*. But `enable` rebuilds `next_autosubmit` entirely from its arguments — only `last_run_at_ms` survives (`autosubmit.rs:210-226`) — and `--interval` defaults to `24h`.

So anyone running `--interval 30m --client codex,claude` who followed my advice would have had their interval **multiplied by 48** and their client filter silently dropped. The remediation caused a worse problem than the drift it was remedying, and it appeared in status output, where nobody reads a footnote before pasting.

It now reconstructs the configuration already in settings:

```
  Scheduled binary: 4.6.1 (this build is 4.7.0)
    The scheduler runs its own copy, which upgrades do not replace. Refresh it with:
      tokscale autosubmit enable --interval 30m --client codex,claude --scheduler launchd
```

Verified end-to-end against the built binary with exactly that non-default setup, plus two unit tests — one pinning a full configuration round-trip, one pinning that unset options stay off the command line.

There is a `Directive:` trailer on the commit noting that any future flag added to `enable` must also be added to `enable_command_for`, or the advice starts silently dropping it again.

## The env serialization was the wrong shape

Also correct, and I could check it rather than take it on faith: **`serial_test = "3.0"` is already a dev-dependency** (`crates/tokscale-cli/Cargo.toml:76`), and `device.rs:186`, `paths.rs:45` and `auth.rs:614` already use `#[serial]` — *and those files redirect the same `TOKSCALE_CONFIG_DIR`*.

So my module-local mutex serialized autosubmit tests against each other and against nothing else. It looked like a fix while coordinating none of the actual cross-module contention. Replaced with `#[serial]` on all 18 env-mutating tests, matching the crate's existing convention, with the reasoning recorded on `EnvVarGuard` so the mutex does not get reinvented.

Three consecutive parallel runs clean.

## Verification

`cargo test -p tokscale-cli` — 967 bin + 133 integration, 0 failed. `cargo clippy --all-targets -- -D warnings` exit 0, `cargo fmt --check` exit 0.

junhoyeo:
Correct, and mine again. Fixed in `26005e9a`.

The previous commit marked env-mutating tests by checking whether `#[serial]` already followed `#[test]`. This module was already using the fully-qualified `#[serial_test::serial]`, which that check did not recognize — so it stacked a bare attribute on top of 15 tests that were already serialized.

It did not hang in practice; the suite and CI both passed. But your reasoning holds regardless: two acquisitions of the same non-reentrant lock is a deadlock waiting on scheduling, and the duplication buys nothing either way.

Fixed by following *this module's* spelling rather than importing the one from `device.rs` and `paths.rs` — that keeps the change to the 3 tests this PR actually introduced instead of rewriting 15 that were already correct.

```
bare #[serial]:            0
#[serial_test::serial]:   18   (15 pre-existing + the 3 added here)
adjacent duplicates:       0
```

Three consecutive parallel runs under a 120s timeout confirm it is not hanging. The guard's doc comment named the wrong spelling and now names the right one, with an explicit line about not stacking both — since that is exactly the trap.

`cargo test -p tokscale-cli` — 967 bin + 133 integration, 0 failed. `cargo clippy --all-targets -- -D warnings` exit 0, `cargo fmt --check` exit 0. Net −11 lines.

---

Worth stating plainly: this is the fourth round on this PR, and rounds 3 and 4 were both *fixes* that introduced new defects — a mutex that coordinated none of the actual contention, then a conversion that double-annotated. Both times I verified my change worked (tests green) rather than that it did what I claimed. The reviews have been carrying real weight here.

PR #975, opened 2026-07-29
fix(9router): read usageHistory table, handle missing requestDetails

n1klaus:
The 9router bridge script now handles both the legacy `requestDetails` table and the new `usageHistory` table added by the v0.5.30 schema migration, with crash resilience when either table is missing.

### Changes

- **Crash resilience**: Each table query is wrapped in `try/except sqlite3.OperationalError` — missing tables are silently skipped. Non-table errors (corrupt DB, locked DB, column mismatch) propagate instead of being silently swallowed.
- **`usageHistory` support**: New `convert_usage_history_row` parses the nested `tokens` JSON column and implements a five-state cost policy: positive cost → provider-reported, zero/missing + free model → `{total: 0.0}`, zero/missing + paid model → cost omitted from output.
- **ID namespace prefixes**: Output IDs are prefixed `rd-` for requestDetails rows and `uh-` for usageHistory rows, preventing cross-table dedup collisions when both tables contain the same raw integer ID.
- **Code quality improvements**: Extracted `_query_and_convert` helper to eliminate duplicate query/convert loops; extracted `_date_str_from_ms` shared timestamp formatter; renamed `convert_row_to_entry` to `convert_request_details_row` for naming symmetry with `convert_usage_history_row`.
- **Test coverage**: Added 16 new tests (60 total) covering cost policy transitions, error propagation paths, cross-table dedup, token parsing robustness, string scalar coercion, and test helper hardening (`CREATE TABLE IF NOT EXISTS`).

### Documentation

The bridge script's docstring has been updated to document the cost policy behavior. No changes to `docs/9router-bridge.md` — the documented contract is unchanged.

Closes the crash-on-missing-table regression introduced by the v0.5.30 migration.

---
## Summary by cubic
Updates the 9router tokscale bridge to read from the new `usageHistory` table and continue gracefully when `requestDetails` is missing after v0.5.30. It also namespaces IDs (`rd-`/`uh-`) and hardens cost parsing (including non-finite values) so tokscale pricing stays correct.

- **New Features**
  - Read `usageHistory` with token JSON parsing and cost rules: positive `cost` passes through; zero/missing → free models embed `{total: 0.0}`, paid models omit cost.
  - Prefix output IDs as `rd-` (requestDetails) and `uh-` (usageHistory) to prevent cross-table dedup collisions.

- **Bug Fixes**
  - Per-table query resilience: only “no such table” is skipped; corrupt DBs and column mismatches re-raise.
  - Normalize `usageHistory.cost` when non-numeric or non-finite (inf/nan) to follow the missing-cost policy (free → `{total: 0.0}`, paid → omit) and avoid crashes/invalid JSON.

<sup>Written for commit f4f4f9c9d004e180593d2f97cbcbd4bada4744fa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/975?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ benchmarks/synthetic-data/
 
 target/
 *.node
+__pycache__/
 
 # Debug logs
 *.log
@@ -34,7 +35,9 @@ self-host/
 .next/
 .vercel
 .env*.local
-.claude/skills/
+.claude/
 .omp/
 .omx/
 .codegraph/
+.agents/
+skills-lock.json
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,7 @@ self-host/
 .next/
 .vercel
 .env*.local
+.claude/skills/
+.omp/
+.omx/
+.codegraph/
--- a/scripts/9router_tokscale_bridge_gjc.py
+++ b/scripts/9router_tokscale_bridge_gjc.py
@@ -35,6 +35,7 @@ See docs/9router-bridge.md for full documentation.
 """
 
 import json
+import math
 import os
 import tempfile
 import sqlite3
@@ -298,6 +299,8 @@ def convert_usage_history_row(row, stats: dict | None = None) -> dict | None:
     if cost is not None:
         try:
             cost = float(cost)
+            if not math.isfinite(cost):
+                cost = None
         except (ValueError, TypeError):
             cost = None
 
--- a/scripts/9router_tokscale_bridge_gjc.py
+++ b/scripts/9router_tokscale_bridge_gjc.py
@@ -17,6 +17,14 @@ CostSource::ProviderReported, which prevents tokscale from repricing via its
 pricing database. Omitting the cost field lets tokscale reprice from
 tokens + pricing data.
 
+This applies to rows from the `requestDetails` table, where costs must be
+inferred from tokens.  Rows from the `usageHistory` table have a dedicated
+`cost` REAL column with provider-reported values; positive costs from this
+column ARE passed through as authoritative.  Zero or missing costs from
+`usageHistory` follow the same default policy (omit for paid models, $0.00
+for free-tier).
+
+
 For FREE-tier models (ids ending in "-free" or ":free"), the bridge embeds
 "cost": {"total": 0.0} on purpose: tokscale's pricing lookup strips the
 "-free" suffix, so an omitted cost would reprice e.g. kimi-k2.5-free at the
@@ -37,6 +45,7 @@ from urllib.parse import quote
 ROUTER_DB = Path.home() / ".9router" / "db" / "data.sqlite"
 BRIDGE_DIR = Path.home() / ".local" / "share" / "9router-tokscale" / "sessions"
 
+
 def discover_router_dbs() -> list[Path]:
     """Discover the current 9Router DB and all backup DBs.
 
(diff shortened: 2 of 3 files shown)


junhoyeo (commented):
Reviewed properly rather than on CI-green. The substance is solid — two things I'd change, neither about the bridge logic.

## What I checked hardest, and it holds

**The `rd-`/`uh-` prefixes are safe.** This was my main concern, because changing an identifier that feeds a dedup key is how you turn an undercount into a permanent double-count. Traced it end to end:

- `gjc.rs:230` does consume it — the dedup key is `<session id>:<message id>` whenever `entry.id` is non-empty, so the prefix genuinely changes gjc's keys.
- But the bridge writes one file per date (`9router-{date}.jsonl`) atomically via temp + rename, so each run **overwrites** rather than accumulates. Only one version of a given date's entries exists on disk at a time.
- IDs never reach the server; the submit payload carries per-day per-client totals, not message ids.

So the prefix fixes the stated cross-table collision without a migration hazard. Worth stating in the PR body, because "we changed the id format" reads alarming without that reasoning.

**Error discrimination is right, and it's the part people usually get wrong.** Only `"no such table"` is swallowed; corrupt DB, locked DB and column mismatch all propagate. A blanket `except sqlite3.OperationalError: pass` would have turned a corrupt database into a silent zero, and that is not what this does.

**The cost policy is more careful than the description claims.** `math.isfinite` rejects NaN/inf and the `cost > 0.0` test means a negative cost falls through to the free/paid policy instead of being passed off as provider-reported. Both degrade toward "let tokscale reprice", which is the safe direction. The docstring's explanation of why free-tier needs an explicit `0.0` — because the pricing lookup strips the `-free` suffix and would reprice at the paid rate — is the kind of reasoning that stops someone "simplifying" it later.

`python3 -m pytest scripts/test_9router_tokscale_bridge_gjc.py` → **62 passed**.

## Please split the `.gitignore` change out

`__pycache__/` is defensible — this PR adds Python tests. But `.claude/skills/`, `.omp/`, `.codegraph/`, `.agents/` and `skills-lock.json` are agent-tooling artifacts with no relationship to a 9router bridge fix.

They're probably all correct additions, which is exactly why they should land on their own: buried in a bridge PR they get no review of their own merits, and `git log` for `.gitignore` stops explaining itself. Happy to take them as a separate one-liner if you'd rather not carry it.

Also drops a blank line above the `.claude/worktrees/` comment block — unintended churn from the merge, worth restoring.

## Nit: say why zero, when it's zero

If both tables are absent, the run ends at `Files: 0, Messages: 0`, which is the same output as "you genuinely have no usage". A user hitting this right after a v0.5.30 upgrade — precisely the scenario this PR exists for — can't tell those apart.

Not the silent-success trap in full, since the counts *are* printed. But one line when neither table was found ("neither `requestDetails` nor `usageHistory` present — is this a 9router DB?") would turn a confusing zero into a diagnosis.

Nothing here blocks merging; the second point is a nit and the first is a request to move code, not change it.

junhoyeo (approved):
Approved. Reviewed in detail above — the ID prefixing is safe (traced through gjc's dedup key and the atomic per-date overwrite), error discrimination correctly propagates non-missing-table failures, and the cost policy degrades safely on NaN/inf/negative. 62 tests pass.

The .gitignore split is a request, not a blocker — happy to take those entries as a separate PR.

PR #985, opened 2026-07-30
feat(report): add MiniMax summarizer backend via OpenAI-compatible API

octo-patch:
Reason: The report summarizer had no MiniMax inference backend, so MiniMax models could not classify sessions through the OpenAI-compatible API.

## What changed
- Add a `minimax` summarizer backend to `tokscale report`, selectable with `--summarizer minimax`, alongside the existing summarizer backends.
- The backend classifies sessions with a MiniMax text model over the OpenAI-compatible `/chat/completions` endpoint, reusing the same system prompt and JSON contract as the other LLM backends.
- Credentials are read from `MINIMAX_API_KEY`, falling back to `MINIMAX_API_TOKEN`, matching the existing MiniMax usage-tracking module.
- `MINIMAX_API_REGION` selects the global (`https://api.minimax.io/v1`) or mainland-China (`https://api.minimaxi.com/v1`) endpoint. `MINIMAX_MODEL` overrides the default model (`MiniMax-M3`); the smaller `MiniMax-M2.7` can be selected the same way.
- The request runs on a current-thread Tokio runtime so it fits the otherwise synchronous summarizer pipeline. Any failure (missing key, network error, unparseable response) is logged and degrades to an empty batch, so the report still renders.
- Update the `--summarizer` help text and the unknown-backend error message to list `minimax`.

## Tests
- Added unit tests covering region-to-base-URL resolution (global default and CN aliases) and OpenAI-compatible response parsing (`choices[0].message.content`, plus malformed input).

## Checks
- `cargo check -p tokscale-cli --tests`
- `cargo test -p tokscale-cli --bin tokscale -- minimax parse_openai` (9 passed, 0 failed)
- `cargo fmt --check`

---
## Summary by cubic
Adds a MiniMax summarizer backend to `tokscale report` using the OpenAI-compatible `/chat/completions` API, so you can classify sessions with MiniMax models while keeping the same prompt and JSON output as other backends.

- **New Features**
  - New `minimax` backend, selectable with `--summarizer minimax`; CLI help and unknown-backend error updated.
  - Uses the same system prompt and JSON contract; failures log and return an empty batch so reports still render.
  - Docs: README examples now list `minimax` across all locales.

- **Migration**
  - Set `MINIMAX_API_KEY` or `MINIMAX_API_TOKEN`.
  - Optional: set `MINIMAX_API_REGION=cn` (aliases: `cn_zh`, `china`, `zh`) and/or `MINIMAX_MODEL` to override the default `MiniMax-M3`.
  - Run: `tokscale report --summarizer minimax`.

<sup>Written for commit 6f0b5d76c5d27959aaf30ae8c3598f675308c5d4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -699,10 +699,11 @@ tokscale report --week
 # Claude Code をサマライザーバックエンドとして使用
 tokscale report --week --summarizer claude
 
-# Codex、Gemini、Kiro を使用
+# Codex、Gemini、Kiro、MiniMax を使用
 tokscale report --summarizer codex
 tokscale report --summarizer gemini
 tokscale report --summarizer kiro
+tokscale report --summarizer minimax
 
 # LLM 要約をスキップ（生データのみ表示）
 tokscale report --no-summarize
--- a/README.ko.md
+++ b/README.ko.md
@@ -696,10 +696,11 @@ tokscale report --week
 # Claude Code를 요약 백엔드로 사용
 tokscale report --week --summarizer claude
 
-# Codex, Gemini 또는 Kiro 사용
+# Codex, Gemini, Kiro 또는 MiniMax 사용
 tokscale report --summarizer codex
 tokscale report --summarizer gemini
 tokscale report --summarizer kiro
+tokscale report --summarizer minimax
 
 # LLM 요약 건너뛰기 (원본 데이터만 표시)
 tokscale report --no-summarize
--- a/README.md
+++ b/README.md
@@ -703,10 +703,11 @@ tokscale report --week
 # Use Claude Code as the summarizer backend
 tokscale report --week --summarizer claude
 
-# Use Codex, Gemini, or Kiro
+# Use Codex, Gemini, Kiro, or MiniMax
 tokscale report --summarizer codex
 tokscale report --summarizer gemini
 tokscale report --summarizer kiro
+tokscale report --summarizer minimax
 
 # Skip LLM summarization (show raw data only)
 tokscale report --no-summarize
--- a/README.zh-cn.md
+++ b/README.zh-cn.md
@@ -700,10 +700,11 @@ tokscale report --week
 # 使用 Claude Code 作为摘要器后端
 tokscale report --week --summarizer claude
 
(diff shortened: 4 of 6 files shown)


junhoyeo (approved):
Reviewed this properly rather than on green CI — cubic finding nothing is not the same as the code being right, and a new backend handling API credentials is worth reading.

It holds up.

## Credentials

The key comes from the environment and goes out as a `Bearer` header, never in a URL — so it cannot leak into a `reqwest` error, which carries the URL but not headers. Neither `eprintln!` path interpolates it. The `MINIMAX_API_KEY` → `MINIMAX_API_TOKEN` fallback matches the existing usage-tracking module rather than inventing a third convention, and the missing-key error names both.

## Region handling

`minimax_base_url` defaults to global and accepts `cn`, `cn_zh`, `china`, `zh`, with trimming and case folding. Pulling that out as a pure function is what makes it testable, and the two tests pin exactly the branch that would otherwise be discovered by someone in the wrong region getting auth failures against the wrong host.

## Verified locally

`cargo test -p tokscale-cli minimax` — 7 passed, including both new ones. Wiring is complete: dispatch arm, and `--summarizer` help updated to `apple-fm, claude, codex, gemini, kiro, minimax`.

## One gap, which I filled rather than bouncing back

The help text listed `minimax` but the README examples did not, so the backend was discoverable only by running `--help`. Pushed `6f0b5d76` adding it to all four locales, matching how every other backend is listed. Nothing else changed — shout if you would rather own that commit yourself and I will drop it.

Thanks for the reuse of the shared system prompt and JSON contract instead of a parallel implementation; that is the part most likely to drift later, and it did not.

PR #998, opened 2026-08-01
feat(clients): add standalone Cline CLI and Kimchi sessions

jR4dh3y:
## Summary

- Add standalone Cline CLI session parsing from `~/.cline/data/sessions/` (or `$CLINE_DIR/data/sessions/`) while preserving the existing VS Code Cline parser.
- Add first-class Kimchi Coding support from Pi-compatible JSONL sessions in `~/.config/kimchi/harness/sessions/` (or `$KIMCHI_CODING_AGENT_DIR/sessions/`).
- Preserve model, provider, workspace, session, input/output/cache token, and cost data; Kimchi remains distinct from Pi despite the shared session format.
- Register both clients across scanning, CLI filters, TUI navigation, wrapped output, frontend metadata, and documentation.

## Validation

- `git diff --check`
- `cargo test -p tokscale-core` — 1,329 passed, 1 ignored
- `cargo test -p tokscale-cli` — 972 unit tests and 133 integration tests passed
- Live scan: `cargo run -p tokscale-cli -- --client kimchi --light --no-spinner` successfully read the local Kimchi session store.

The frontend typecheck could not run in this checkout because the `tsc` binary is not installed.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -69,6 +69,7 @@
 | <img width="48px" src=".github/assets/client-droid.png" alt="Droid" /> | [Droid (Factory Droid)](https://factory.ai/) | `~/.factory/sessions/` |
 | <img width="48px" src=".github/assets/client-pi.png" alt="Pi" /> | [Pi](https://github.com/badlogic/pi-mono) | `~/.pi/agent/sessions/` and `~/.omp/agent/sessions/` ([Oh My Pi](https://github.com/can1357/oh-my-pi)) |
 | <img width="48px" src=".github/assets/client-senpi.png" alt="Senpi" /> | [Senpi (OmO Native)](https://github.com/code-yeongyu/senpi) | `~/.senpi/agent/sessions/` (`SENPI_CODING_AGENT_DIR` でオーバーライド可能) |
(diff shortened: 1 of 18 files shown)


junhoyeo:
@jR4dh3y Thanks for the Cline CLI and Kimchi work. Kimchi's path, parser reuse, cache, dedup, and frontend registration look coherent, but the Cline side has two correctness blockers before merge:

- Cline persists tool results as `role: "user"` content blocks. The parser currently drops `content`, treats every user-role record as a fresh human interaction, and marks the next assistant record as `is_turn_start`. A normal tool loop therefore inflates turn counts. Please deserialize content and start a turn only for a genuine human-prompt record; add a prompt → assistant/tool-use → user/tool-result → assistant fixture expecting exactly one turn.
- Discovery checks only `CLINE_DATA_DIR`. Please match Cline's upstream precedence: `CLINE_SESSION_DATA_DIR` → `CLINE_DATA_DIR/sessions` → `CLINE_DIR/data/sessions` → `~/.cline/data/sessions`, with isolation/precedence tests and coverage for `use_env_roots=false`.

Please also apply the generated Cline logical dedup keys in both cached and local parse paths, and mirror the new Cline/Kimchi support entries into `README.ko.md`, `README.ja.md`, and `README.zh-cn.md` as required by the repository checklist.

Issue #849, opened 2026-07-09
Grok Build is reporting incorrect values


entrptaher:
On a new `npx tokscale` command, it lists only the input values and completely misses the output, cache read and other values even though they are present in the files inside ~/.grok directory.
 
<img width="2001" height="637" alt="Image" src="https://github.com/user-attachments/assets/fae16a99-f2f7-44bf-9ba7-b56ed12aa243" />

junhoyeo:
Thanks — confirmed and fixed. The screenshot's split metrics are from `~/.grok/logs/unified.jsonl`, whereas Tokscale previously only consumed cumulative totals from legacy session updates.

I independently verified the unified-log schema from public runtime captures, so a contributor fixture was not needed: cached prompt tokens are included in prompt tokens, and reasoning tokens are included in completion tokens. Tokscale now reads the unified log, records non-cached input, cache reads, visible output, and reasoning as non-overlapping buckets, and uses it instead of legacy data only for sessions the unified log covers. The parser also keeps a session-specific model selection ahead of later global catalog refreshes, avoiding a possible model/pricing misattribution.

Fixed in junhoyeo/tokscale@ed7986425efb8bad8df62e16269e30c15a82801a. The regression coverage includes the reported breakdown, duplicate/invalid rows, legacy overlap, scanner discovery, cache-path parsing, and the model-precedence case. It is queued for the next release.

Astro-Han:
Still seeing this on tokscale 4.5.3 (current npm latest).

Grok Build rows still land as input-only: Output / Cache R / Cache W are 0 for k3, grok-4.5, glm-5.2, etc. Claude / Codex / Pi on the same day look normal.

I do have `~/.grok/logs/unified.jsonl` (a few MB, actively written). Session dirs also have the richer path:

`_x.ai/session/update` → `params.update.usage.modelUsage.<model>`

with `inputTokens`, `outputTokens`, `totalTokens`, `cachedReadTokens`, `reasoningTokens`. On k3, `cachedReadTokens` is usually ~98% of input, so zeroing cache blows up cost if you price it.

Rough check for one day of k3: tokscale reports ~2.3M input / $7 with a custom-pricing override. Summing final `modelUsage` per session is more like ~1e8 tokens (some resume double-count risk, but not a 40× gap). Looks like the old positive-`totalTokens`-delta-as-input path more than a full split.

[...]

PR #1000, opened 2026-08-01
fix(grok): parse unified usage with scoped attribution

jR4dh3y:
## Summary

Grok's unified log now carries role-specific lifecycle events for parent and child model selection. This PR scopes that attribution to the PID generation and exact child session, while keeping the unified token breakdown fix for the original issue.

The earlier parser could reuse a model after a process restarted with the same PID, or let a child inference inherit a parent or unrelated session model. Missing, malformed, or conflicting child evidence now stays `grok-unknown` instead of guessing.

## Changes

- Recognize current parent, spawn, completion, failure, and `AuthManager::new` lifecycle records.
- Scope parent model authority to `(pid, process generation)` and child authority to `(pid, process generation, subagent session)`.
- Support pidless session model changes and recover consistent model/workspace metadata from legacy rows without double-counting them.
- Preserve non-overlapping input, output, cache-read, and reasoning buckets; clamp malformed overlap values and keep tool-loop rows at zero user-message count.
- Bump the Grok parser cache identity because the same source can now produce a different model ID.

The attribution rules are adapted from the merged [Nanako0129/tokscale-core#2](https://github.com/Nanako0129/tokscale-core/pull/2) implementation for this repository's parser and cache layout.

## Verification

- `cargo test -p tokscale-core` — 1333 passed, 1 ignored.
- `cargo test -p tokscale-cli` — 972 passed, 1 ignored; 133 integration tests passed.
- `git diff --check` passed.
- A local cold and warm Grok scan returned the same totals: input 15,346,334; output 922,871; cache-read 170,854,784; reasoning 1,344,668.

Refs #849

---
## Summary by cubic
Scopes Grok’s unified model attribution by PID generation and exact child session, prefers per‑inference usage from `~/.grok/logs/unified.jsonl` and legacy per‑turn usage rows, and applies pricing after unified selection, addressing #849. Conflicting child evidence stays `grok-unknown` and unpriced; unified coverage and costs are stable across runs.

- **New Features**
  - Parse unified inference events and legacy per‑turn usage; split overlapping buckets (cache read, reasoning), dedupe by event ID or full row, and keep only post‑usage fallback deltas.
  - Prefer unified rows; recover model/workspace from consistent legacy rows; selection is order‑independent and prevents double counting.
  - Discover `logs/unified.jsonl` for the primary Grok home and `scanner.extraScanPaths.grok` roots (home or `sessions` shapes); keep recursive `updates.jsonl` and ignore nested archives.

- **Bug Fixes**
  - Scope model attribution to `(pid, generation)` and `(pid, generation, subagent session)`; persist `model_attribution_conflicted` so conflicted rows stay `grok-unknown` and unpriced.
  - Bump Grok parser cache to v6 and fingerprint the full sessions metadata tree; bump cache shard format to v4; directory-aware fingerprints; parse a stable unified snapshot; reprice after unified selection.

<sup>Written for commit 1e6f74010bcc7f6fb6be41f78d3a6fa3c8ddd8df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1000?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/aggregator.rs
+++ b/crates/tokscale-core/src/aggregator.rs
@@ -782,6 +782,7 @@ mod tests {
             dedup_key: None,
             session_title: None,
             is_turn_start: false,
+            model_attribution_conflicted: false,
         }
     }
 
@@ -1397,6 +1398,7 @@ mod tests {
             dedup_key: None,
             session_title: None,
             is_turn_start: false,
+            model_attribution_conflicted: false,
             duration_ms: None,
         }
     }
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -4049,6 +4049,7 @@ pub fn parsed_to_unified(msg: &ParsedMessage, cost: f64) -> UnifiedMessage {
         dedup_key: None,
         session_title: None,
         is_turn_start: false,
+        model_attribution_conflicted: false,
     }
 }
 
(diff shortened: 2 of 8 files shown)


junhoyeo:
@jR4dh3y The scoped attribution implementation looks correct; the full core suite, clippy, formatting, and targeted Grok tests pass.

I recommend one contained fix in this PR: make configured Grok roots receive the same dual-source discovery as the primary Grok home. Currently [`unified.jsonl` is added only for the resolved primary home](https://github.com/junhoyeo/tokscale/blob/95f44b170cbda4861eaf902d2e5934455336b1ed/crates/tokscale-core/src/scanner.rs#L1164), while `scanner.extraScanPaths.grok` continues to use only the registered `updates.jsonl` pattern. Please centralize Grok-home discovery so a supported alternate root contributes both sources, and add a scanner regression test. If alternate roots are intentionally legacy-only, that limitation should be explicit in the settings documentation.

The global unified-log cache also scales with the complete log and historical session tree. I do not think this PR should be held for an incremental-cache redesign: preserving attribution state and handling log truncation/rotation makes that a separate performance change. A benchmark plus a follow-up issue for offset/state-based caching is the right scope.

jR4dh3y:
Implemented the review fixes in `25431ec2` (`fix(grok): keep unified coverage and pricing consistent`).

- Reprice Grok messages after unified/legacy attribution so recovered model IDs have correct costs on cold and warm cache paths.
- Treat Grok input/output usage buckets as inclusive even when `totalTokens` is omitted, preventing cache-read and reasoning double counting.
- Fingerprint the complete unified-log session metadata tree, including workspace/session directories and metadata siblings, and bump the Grok parser cache version.
- Deduplicate unified inference rows by stable event IDs when present, otherwise by the normalized source row, so exact duplicates collapse without dropping distinct repeated events.
- Replace legacy activity only when it is proven to overlap unified activity; preserve older or otherwise unmatched rows in partially migrated sessions.

Validation completed: `cargo fmt --all -- --check`, `cargo clippy -p tokscale-core --all-targets -- -D warnings`, and `cargo test -p tokscale-core` (1,338 passed, 1 ignored).

jR4dh3y:
Fixed the new order-dependence finding in `95f44b17` (`fix(grok): make legacy selection order-independent`). Exact activity matches now consume only exact activity coverage, while fallback timestamp coverage is consumed only by fallback rows, so the selector produces the same result regardless of legacy row order. Added `selector_is_order_invariant_for_activity_and_fallback_rows` to cover the regression. Validation passes: `cargo fmt --all -- --check`, `cargo clippy -p tokscale-core --all-targets -- -D warnings`, and `cargo test -p tokscale-core` (1,339 passed, 1 ignored).

jR4dh3y:
Both recommendations addressed.

**Dual-source discovery for configured Grok roots** — fixed in `ac9a51e1` (`fix(grok): discover unified log from alternate scan roots`). Grok root discovery is now centralized in a single `push_grok_dual_source_scan_tasks` helper that registers the legacy `updates.jsonl` scan on the original root (preserving `extraScanPaths` recursive-root semantics — the configured path is never replaced by a `sessions/` subdirectory) and derives the sibling `<home>/logs/unified.jsonl` task from it. The helper is applied uniformly to the resolved primary home, every `scanner.extraScanPaths.grok` entry, and `TOKSCALE_EXTRA_DIRS` grok entries, so an alternate root now contributes both sources. The home is derived robustly whether the configured root points at the home itself (`~/.grok`) or its `sessions` subdirectory (the shape the primary resolution returns).

Added two scanner regression tests: `test_grok_extra_scan_path_discovers_both_sources` (asserts the alternate root yields both `updates.jsonl` and `logs/unified.jsonl`, and retains a nested `updates.jsonl` outside `sessions/` to prove recursion is preserved) and `test_grok_extra_scan_path_sessions_shape_discovers_unified_log` (covers the `sessions/`-dir configuration shape). Full core suite (1374 tests), clippy, and `fmt --check` all pass, and the branch is merged with current `main`. Since alternate roots are now fully dual-source (not legacy-only), no settings-documentation limitation was needed.

**Unified-log cache scope** — agreed; kept out of this PR. Filed the follow-up as https://github.com/junhoyeo/tokscale/issues/1018, with a benchmark plus offset/state-based caching and acceptance criteria covering truncation/rotation detection, append-consistency, attribution-state preservation, and the sessions-metadata fingerprint dependency.

jR4dh3y:
Addressed the three new scanner findings in `380e247d` (`fix(grok): handle nested alternate scan roots`).

- `grok_home_from_scan_root` now searches ancestors for the nearest `sessions` component using case-insensitive matching, so roots ending in `Sessions` and roots below `sessions/<workspace>/<session>` derive the correct Grok home and sibling `logs/unified.jsonl`.
- Both Grok scanner regression tests now capture and remove `TOKSCALE_EXTRA_DIRS` as well as `GROK_HOME`, keeping their exact file-count assertions hermetic.
- Added `test_grok_extra_scan_path_nested_session_shape_discovers_unified_log`, covering a mixed-case `Sessions` directory with an alternate root below it.

Validation: the three targeted scanner tests pass, `cargo fmt -p tokscale-core -- --check` passes, the full core suite passes (1375 passed, 1 ignored), and `cargo clippy -p tokscale-core --all-targets -- -D warnings` passes.

Issue #1027, opened 2026-08-04
cargo test -p tokscale-core --lib fails to compile on main: build_graph_from_messages call sites missing the bucket timezone argument


MoerAI:
## Summary

`cargo test -p tokscale-core --lib` does not compile on current `main` (`56b2008c`). Two tests added by #1026 call `build_graph_from_messages` with four arguments, but #1016 had already added a fifth parameter. Both PRs merged cleanly as text, so the conflict only surfaces at compile time, and it blocks the entire `tokscale-core` unit test suite.

## Repro

```console
$ git rev-parse HEAD
56b2008c92649ddd6a7fc5af525e6b0f318b36c7

$ cargo test -p tokscale-core --lib --no-run
error[E0061]: this function takes 5 arguments but 4 arguments were supplied
    --> crates/tokscale-core/src/lib.rs:7955:21
     |
7955 |           let graph = build_graph_from_messages(
     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^-
     |  argument #5 of type `&bucket_tz::BucketTimezone` is missing

error[E0061]: this function takes 5 arguments but 4 arguments were supplied
    --> crates/tokscale-core/src/lib.rs:7995:21
     |
7995 |           let error = build_graph_from_messages(
     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^-
     |  argument #5 of type `&bucket_tz::BucketTimezone` is missing

error: could not compile `tokscale-core` (lib test) due to 2 previous errors
```

## Cause

`49dbf525` ("fix(cli): pin the timezone usage days are bucketed into", #1016) added `bucket_timezone: &bucket_tz::BucketTimezone` as the fifth parameter of `build_graph_from_messages` and updated every call site that existed on its branch. `56b2008c` ("fix(pricing): resolve gemini-default routing label for submission", #1026) added two new tests — `submission_excludes_unpriced_generic_gemini_default_but_keeps_priceable_usage` and `submission_still_rejects_unpriced_concrete_models` — that were written against the four-argument signature. Neither branch touched the same lines, so git merged both without a textual conflict.

The pre-existing test `graph_pricing_policy_is_strict_only_for_submission` at `lib.rs:7897` already passes `&crate::bucket_tz::BucketTimezone::Local` as the fifth argument, so it shows the intended shape for the two broken call sites.

## Impact

[...]

Issue #1021, opened 2026-08-03
submit fails on github-copilot/oswe-vscode-prime, nvidia/minimaxai/minimax-m2.7, gemini-3-pro-preview, zai/glm-4.5-flash — pricing is unavailable


sigco3111:
## Summary

`tokscale submit` (v4.9.0) rejects the entire batch with `pricing is unavailable for submitted token usage` when local sessions include these model IDs:

- `github-copilot/oswe-vscode-prime`
- `nvidia/minimaxai/minimax-m2.7`
- (cache-dependent) `google/gemini-3-pro-preview`, `zai/glm-4.5-flash`

I checked the relevant code paths and want to confirm whether the right fix is upstream or local before filing a PR. Findings below — please correct anything I got wrong.

## Repro

```sh
bunx tokscale@latest --version  # 4.9.0
bunx tokscale@latest submit
```

Output:

```
Error: pricing is unavailable for submitted token usage:
github-copilot/oswe-vscode-prime,
google/gemini-3-pro-preview (×17),
nvidia/minimaxai/minimax-m2.7 (×N),
zai/glm-4.5-flash
```

## Findings from `crates/tokscale-core/`

1. The error originates at `src/lib.rs:2955` (no pricing) and `src/lib.rs:2988` (unpriced rows). `validate_priced_messages` (line 2950) rejects the whole batch when any token-bearing row has neither `has_authoritative_cost()` nor a `covers_usage_with_provider` hit (line 2969).

2. `covers_usage_with_provider` (pricing/mod.rs:394) → `lookup_with_source_and_provider` (pricing/lookup.rs:321) which **does** have sophisticated prefix stripping, but the actionable whitelist (`PROVIDER_PREFIXES` at lookup.rs:6-18 and `RESELLER_PROVIDER_PREFIXES` at lookup.rs:37-48) does not include:
   - `github-copilot/` (note the hyphen vs underscore)
   - `github_copilot/` **is** in `EXCLUDED_LITELLM_PREFIXES` (pricing/mod.rs:26) which prevents LiteLLM hits from being considered at all — comment at mod.rs:22-26 explains the rationale (subscription pricing, "$0.00 meaningless for per-token").
   - `nvidia/` and `nvidia_nim/` — NIM catalog keys in LiteLLM use `nvidia_nim/nvidia/...`; user-visible IDs from OpenCode show up as `nvidia/...`.
[...]

Issue #1014, opened 2026-08-03
Windows suite: 33 pre-existing failures, six root-cause groups


junhoyeo:
Split out of #997. #1012 adds `windows-latest` to `test_coverage.yml` and `--no-fail-fast` to the test step. Running the suite on Windows for the first time produced **34 failures**, not the 12 #997 reported — because cargo used to stop at the first failing binary, so **nine of the ten test binaries had never run at all.**

#1012 fixes a number of these and leaves the windows leg `continue-on-error: true`. This issue tracks the remaining **33**, so the leg can become a hard gate.

**On counting.** The baseline run cited below is on `db6014b2`, this PR's *first* commit, which already contains the `$HOME` resolver work — so its 34 is a post-partial-fix number, not a clean pre-fix baseline. windows-latest has never run without any fixes, because the commit that added the matrix also added them, so the PR's total effect is not measurable from CI. What is measurable: `db6014b2` = 2461 passed / 34 failed, `1468a332` = 2462 passed / 33 failed, `e50869eb` = 2468 passed / 33 failed (the +6 are that commit's own new tests; the failing set is byte-identical to `1468a332`).

## Baseline

Run: https://github.com/junhoyeo/tokscale/actions/runs/30773847485 (job `Code Coverage (windows-latest)`)

| binary | passed | failed |
|---|---|---|
| `tokscale-cli` (bin) | 975 | 1 → **0 after #1012** |
| `cli_tests` (integration) | 120 | **13** |
| `tokscale-core` (lib) | 1331 | **20** |
| other 7 binaries | 35 | 0 |

The 7 quiet binaries had also never run before; they pass.

**The enumeration below now reconciles to 33.** The first revision of this issue listed only 31: the Group 2 and Group 3 headers undercounted their own bullet lists (5 vs 6, 4 vs 5), and two failures were missing entirely — `test_submit_default_graph_includes_antigravity_cache_rows` (added to Group 1) and `test_scan_all_clients_kimi_code_home_override_is_not_trimmed` (new Group 7). 14 + 6 + 5 + 3 + 2 + 2 + 1 = 33.

## Group 1 — config dir is not redirectable on Windows (14 tests)

[...]

PR #1028, opened 2026-08-04
fix(pricing): keep published cache rates so submit stops rejecting Codex usage

### LINKED ISSUE #1013: `tokscale submit` aborts when usage contains unpriced or free model IDs
## Description

`tokscale submit` aborts the entire submission when locally scanned usage contains model IDs for which pricing is unavailable.

In this case, Cursor usage sync completes successfully and the local session scan then fails because the history includes free, automatic, or provider-specific model IDs. Since some of these models are free or do not have a directly resolvable public price, it would be useful if submission could continue while treating them as zero-cost or unknown-cost usage.

## Steps to reproduce

1. On Windows, have local Cursor/OpenCode/Codex session data containing one or more unpriced model IDs.
2. Run:

```powershell
bunx tokscale submit
```

3. Observe that the command stops during the local session scan.

## Actual behavior

```text
Tokscale - Submit Usage Data

Syncing Cursor usage data...
Cursor: 6581 usage events synced
Error: pricing is unavailable for submitted token usage: opencode/qwen3.6-plus-free, openai/codex-auto-review, cursor/auto, cursor/composer-1, anthropic/claude-4-sonnet-thinking
```

The actual error repeats each model ID once per affected usage event. The captured run contained:

- `cursor/auto`: 4,122 occurrences
- `openai/codex-auto-review`: 320
- `anthropic/claude-4-sonnet-thinking`: 174
- `opencode/qwen3.6-plus-free`: 24
- `cursor/composer-1`: 14

This also produces a very large error message.

## Expected behavior

The submission should not fail entirely because some usage has no available pricing. Possible handling:

- submit token counts with cost marked as unknown/null;
- treat explicitly free models such as `qwen3.6-plus-free` as zero-cost;
- skip only the affected cost calculation and print a deduplicated warning with counts; or
- offer a flag to allow submission with unknown pricing.

At minimum, the error should deduplicate model IDs and report occurrence counts.

## Environment

- OS: Windows 11
- Shell: PowerShell
- Runner: `bunx tokscale submit`
- Date reproduced: 2026-08-03

MoerAI:
Fixes #1013. Also covers the `gemini-3-pro-preview` half of #1021.

## What happens today

`tokscale submit` aborts before any network call as soon as local history contains a model whose cached tokens have no published rate:

```console
$ tokscale submit
  Tokscale - Submit Usage Data
Error: pricing is unavailable for submitted token usage: openai/gpt-5.2-codex, openai/gpt-5.2-codex, openai/gpt-5.2-codex, ... [632 entries]
```

Deduplicated, that run was `openai/gpt-5.2-codex` x414, `openai/gpt-5.1-codex-max` x180, `openai/gpt-5.1-codex` x29, `openai/gpt-5.1-codex-mini` x5, `google/gemini-3-pro-preview` x3. #1006 scoped strict per-bucket validation to the submit path in 4.9.0, which is why 4.8.0 still works on identical data.

## Root cause

`ModelPricing::covers_usage` requires a rate for every populated token bucket, and Codex sessions always carry cached-prefix tokens. Two separate paths hand it a row with no cache-read rate.

**OpenRouter drops cache rates it already fetched.** Author pricing is read from `/models/{id}/endpoints` by matching `provider_name` against the model id's author, and when no such endpoint exists the row falls back to the `/models` list price, which hardcodes both cache fields to `None`. OpenRouter serves all four failing Codex models only from an `Azure` endpoint, so the `OpenAI` lookup misses and the `input_cache_read` price each of them publishes is thrown away:

```console
$ curl -s https://openrouter.ai/api/v1/models/openai/gpt-5.2-codex/endpoints | jq '.data.endpoints[] | {provider_name, cache: .pricing.input_cache_read}'
{ "provider_name": "Azure", "cache": "0.000000175" }
```

**A provider hint can resolve onto a row that has no cache rates when the canonical row does.** Under a `google` hint, `gemini-3-pro-preview` resolved to `gmi/google/gemini-3-pro-preview` (no cache-read rate) even though the bare `gemini-3-pro-preview` row publishes one. Coverage was judged against the hinted row alone.

Both paths also under-billed reports, because `compute_cost` treats a missing rate as zero. This is why the failure looks intermittent across machines and platforms in #1013 and #1021: it depends entirely on which rows the pricing caches happen to hold. `gpt-5.2-codex` was present in LiteLLM with a cache-read rate when I started investigating and had been removed upstream an hour later, which flipped the resolution to the cache-less OpenRouter row.

## Changes

`fix(openrouter): keep cache pricing when a model has no author endpoint` — the author's own endpoint still wins, so a model whose author does serve it keeps that price rather than a reseller's markup. Only when the author endpoint is absent does the lookup adopt an endpoint quoting the same base prompt and completion price as the list row already being used, which keeps the cache rates published alongside a price that was going to be accepted regardless. An endpoint quoting a different base price is a different deal and is never adopted.

`fix(pricing): fill missing bucket rates from the canonical row when a provider hint omits them` — resolution borrows rates for the buckets the hinted row cannot price from the canonical unhinted row. Rates already present are never overwritten, so a reseller keeps its own markup instead of silently repricing to the author's cheaper rate — `azure_ai/grok-code-fast-1` at $3.50/$17.50 must not collapse onto `xai/grok-code-fast-1` at $0.20/$1.50. Each bucket is filled together with its long-context tiers so a tiered walk never mixes rates from two rows, and the hinted row's source and matched key are preserved because `compute_cost_for_lookup` branches on both for OpenAI's full-request 272k tiering. If the filled row still cannot cover the usage, the hinted row is returned unchanged and the usage stays unpriced.

`fix(submit): deduplicate model ids in the unpriced-usage error` — ids are reported once each, in first-seen order, with an occurrence count when they repeat, as requested in #1013.

`fix(test): pass the bucket timezone to build_graph_from_messages in submission tests` — `cargo test -p tokscale-core --lib` does not compile on `main`, so none of the regression tests above could run without this. Filed separately as #1027; happy to split it out if you would rather land it on its own.

## One existing expectation changed

`incomplete_unhinted_result_does_not_replace_provider_pricing` now expects `3.0` instead of `1.0`. It pairs a provider row that prices only input with a canonical row that prices only output, and its `1.0` expectation encoded the output bucket costing nothing. The hazard it guards against — the canonical row replacing the provider row wholesale and pricing input at zero — would total `2.0` and is still rejected. The provider row keeps its own input rate and only borrows the output rate it does not publish.

## Verification

Before and after on the same machine and the same local history, Windows 11 / PowerShell:

```console
$ tokscale submit --dry-run          # before
Error: pricing is unavailable for submitted token usage: openai/gpt-5.2-codex, ... [632 entries]
$ echo $LASTEXITCODE
1

$ tokscale submit --dry-run          # after
  Data to submit:
    Date range: 2025-11-27 to 2026-08-04
    Active days: 148
    Total tokens: 16,723,103,394
    Total cost: $15407.96
    Clients: claude, codex, gemini, gjc, opencode, senpi
    Models: 32 models
  Dry run - not submitting data.
$ echo $LASTEXITCODE
0
```

The cached tokens are now priced instead of billed at zero. `gpt-5.2-codex` goes from `$5.7180` to `$9.2843`, which is exactly its 20,379,136 cached tokens at the $0.175/M rate OpenRouter publishes, and `tokscale pricing gpt-5.2-codex` now reports `Cache Read: $0.17 / 1M tokens`.

- `cargo test -p tokscale-core --lib` — 1386 passed, 20 failed. The same 20 fail on a clean worktree of `56b2008c` with only the compile fix applied, so there are no new failures; they are the Windows `HOME`/path group tracked in #1014.
- `cargo test -p tokscale-cli` and `cargo test -p tokscale-cli --test cli_tests` — failure sets identical to that same baseline (1 unit, 23 integration), all in the same Windows path group.
- `cargo clippy -p tokscale-core --all-targets -- -D warnings` and `cargo fmt --all -- --check` clean. `cargo clippy -p tokscale-cli --all-targets` already fails on `main` for unrelated reasons (unused imports and Windows-inactive helper functions); this PR touches no CLI sources.
- 8 new tests cover the OpenRouter endpoint selection, the bucket fill and its markup guard, the genuinely-unpriced rejection, custom-pricing precedence, and the error dedup.

## Not addressed

A model with no published rate anywhere for a bucket it actually uses is still rejected, which is deliberate: submitting it would bill real usage at zero. `opencode/minimax-m2.5-free` from the #1013 report is an example — models.dev gives it input, output and cache-read rates but no cache-write rate. The broader options discussed in #1013 and #1021 (a `--prune-unpriced` flag, zero-cost custom pricing entries, subscription-priced providers) are policy decisions I did not want to make unilaterally.

---
## Summary by cubic
Stops `tokscale submit` from aborting when cached tokens are present by preserving OpenRouter cache prices and borrowing only missing bucket rates when base prices match. Fixes #1013 and covers the `gemini-3-pro-preview` half of #1021.

- **Bug Fixes**
  - OpenRouter: keep cache pricing when the model has no author endpoint by adopting an endpoint that matches the listed base price; the author endpoint still wins; prefer the endpoint that publishes the most cache rates.
  - Provider hints: borrow only missing bucket rates from the canonical row when both quote the same base rates and share at least one bucket; never overwrite existing rates (protects reseller markups); keep long‑context tiers and source; if still incomplete, leave unpriced.
  - Custom pricing: custom rows decide coverage before any provider resolution or borrowing.
  - Submission UX: de‑duplicate unpriced model ids and show counts; simplify the excluded‑usage message.
  - Billing correctness: cached tokens are now priced instead of billed at zero in reports.

<sup>Written for commit 12761daf134cf882a1db269f27c5766bfd11aa96. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1028?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -5488,9 +5488,11 @@ fn report_unpriced_submission_exclusions(
     use colored::Colorize;
 
     for row in excluded {
-        let remaining_usage_message = has_remaining_usage
-            .then_some(" Remaining priced usage will be submitted.")
-            .unwrap_or_default();
+        let remaining_usage_message = if has_remaining_usage {
+            " Remaining priced usage will be submitted."
+        } else {
+            ""
+        };
         println!(
             "{}",
             format!(
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -3095,40 +3095,60 @@ fn validate_priced_messages(
         return Err("pricing data is unavailable for submission".to_string());
     };
 
-    let unpriced: Vec<String> = messages
-        .iter()
-        .filter(|message| {
-            let tokens = &message.tokens;
-            let token_bearing = tokens.input > 0
-                || tokens.output > 0
-                || tokens.cache_read > 0
-                || tokens.cache_write > 0
-                || tokens.reasoning > 0;
-            token_bearing
-                && !message.has_authoritative_cost()
-                && !pricing.covers_usage_with_provider(
-                    &message.model_id,
-                    Some(&message.provider_id),
-                    &message.tokens,
-                )
-        })
-        .map(|message| {
-            if message.provider_id.is_empty() {
-                message.model_id.clone()
-            } else {
-                format!("{}/{}", message.provider_id, message.model_id)
-            }
-        })
-        .collect();
+    // Counted rather than listed per message: a real submission repeats the
+    // same handful of ids thousands of times, and the raw list buried the
+    // actionable model names under hundreds of kilobytes of output (#1013).
+    let mut counts: std::collections::HashMap<String, usize> = std::collections::HashMap::new();
+    let mut order: Vec<String> = Vec::new();
 
-    if unpriced.is_empty() {
-        Ok(())
-    } else {
-        Err(format!(
-            "pricing is unavailable for submitted token usage: {}",
-            unpriced.join(", ")
-        ))
+    for message in messages {
+        let tokens = &message.tokens;
+        let token_bearing = tokens.input > 0
+            || tokens.output > 0
+            || tokens.cache_read > 0
+            || tokens.cache_write > 0
+            || tokens.reasoning > 0;
+        let unpriced = token_bearing
+            && !message.has_authoritative_cost()
+            && !pricing.covers_usage_with_provider(
+                &message.model_id,
+                Some(&message.provider_id),
+                &message.tokens,
+            );
+        if !unpriced {
+            continue;
+        }
+
+        let id = if message.provider_id.is_empty() {
+            message.model_id.clone()
+        } else {
+            format!("{}/{}", message.provider_id, message.model_id)
+        };
+        match counts.get_mut(&id) {
+            Some(count) => *count += 1,
+            None => {
+                counts.insert(id.clone(), 1);
+                order.push(id);
+            }
+        }
     }
+
+    if order.is_empty() {
+        return Ok(());
+    }
+
+    let summary = order
+        .into_iter()
+        .map(|id| match counts.get(&id).copied().unwrap_or(1) {
+            1 => id,
+            count => format!("{id} (x{count})"),
+        })
+        .collect::<Vec<String>>()
+        .join(", ");
+
+    Err(format!(
+        "pricing is unavailable for submitted token usage: {summary}"
+    ))
 }
 
 fn filter_messages_for_report(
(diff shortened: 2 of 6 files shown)


MoerAI:
Thanks — the P1 was right, and it was the exact hazard I described in the PR body and then failed to guard. I had protected the buckets the hinted row already prices but left the borrowed bucket itself unchecked, so an `azure_ai/grok-code-fast-1` row could indeed have inherited the canonical `xai/` cache-read rate and billed on a tariff neither provider charges. Fixed in 1cf9a51.

Borrowing is now gated on the two rows pricing the same deal: they must agree on every base bucket where both publish a rate, and they must share at least one such bucket, since without a single common rate there is no evidence they describe the same deal. `differently_priced_canonical_row_does_not_lend_its_cache_rate` covers it with the $3.50/$17.50 versus $0.20/$1.50 numbers from the review.

A welcome side effect: this restores the original `1.0` expectation of `incomplete_unhinted_result_does_not_replace_provider_pricing`, which I had changed to `3.0`. That fixture pairs a provider row pricing only input with a canonical row pricing only output, so under the equivalence rule the two share no base bucket, nothing is borrowed, and the existing assertion holds unmodified. The PR no longer changes any existing expectation.

On the long-context tiers — also correct, and the doc comment claimed the behaviour the code did not have. Each tier is now filled only when absent, so a row publishing an above-threshold rate for a bucket whose base rate it omits keeps its own tier (d3cb03d).

On the endpoint selection — correct too; cache read and cache write are independent fields, so preferring the first endpoint with a read rate could hide a same-price endpoint publishing both. Selection now takes the candidate publishing the most cache rates, ties keeping the earliest endpoint (8ef3d62).

Re-verified after the changes: `submit --dry-run` still exits 0 on the same local history, `cargo test -p tokscale-core --lib` is 1391 passed with the same 20 pre-existing Windows failures as a clean baseline of `56b2008c`, and `cargo clippy -p tokscale-core --all-targets -- -D warnings` plus `cargo fmt --all -- --check` are clean.

PR #1030, opened 2026-08-04
fix(sync): hold the antigravity and trae sync locks in the OS

### LINKED ISSUE #1010: Sync lock protocols use a non-atomic read-decide-unlink instead of an OS-held lock
Split out of the review on #1007 so it does not get lost. **Pre-existing, not introduced by that PR**, and equally present on Unix today — which is why it was not gated there.

## The problem

`SyncLockGuard::acquire` (`crates/tokscale-cli/src/antigravity.rs`) and the Trae equivalent (`crates/tokscale-cli/src/trae.rs`) both implement mutual exclusion with a PID file:

1. `create_new(true)` the lock file — fails if it exists
2. on `AlreadyExists`, read the stored PID and probe it
3. if the owner is dead, `remove_file` and retry

Steps 2 and 3 are not atomic, so:

- **Two contenders can both evict the same stale lock.** Both read the dead PID, both unlink, both then create and enter the critical section. The whole point of the lock — that two syncs never overlap on the manifest and delete each other's session artifacts — does not hold.
- **The lock file is visible before it means anything.** `create_new` succeeds, then the PID and timestamp are written on the next line. A contender that opens the file in that window reads an empty file. `read_sync_lock` returns `None` for that, so the contender skips the liveness check entirely and unlinks a lock that a live process just took.
- **The write errors are discarded** (`let _ = writeln!(...)`), so a lock whose PID never made it to disk looks identical to a lock nobody owns.
- **PID reuse allows false ownership.** After the owner dies, an unrelated process can land on the same PID and be mistaken for it, holding the lock indefinitely.

#1007 made the liveness probe fail-alive on every platform, which narrows the window but cannot close it — no PID probe can, because the race is in the protocol, not the probe.

## The fix

Use an OS-held exclusive lock. `autosubmit::try_acquire_run_lock` (`crates/tokscale-cli/src/commands/autosubmit.rs`) already does exactly this with `fs2::try_lock_exclusive`, and the kernel releases it on process death — which removes the need for both the PID bookkeeping and the stale-lock recovery loop.

Two things to preserve when converting:

- The comment at the antigravity call site records that **age-based eviction was removed on purpose**. Do not reintroduce a timestamp heuristic; an OS lock makes it unnecessary.
- `try_acquire_run_lock` returns `Ok(None)` on contention rather than erroring. The sync locks currently `bail!` with the owning PID in the message, which is a better diagnostic — keep an equivalent message.

## Why it was not fixed in #1007

#997 reports neither this race nor PID reuse, the defect is not Windows-specific, and rewriting both sync lock protocols is a much larger change than the reported bug. Gating the Windows fixes on it would have been the wrong trade.

MoerAI:
Fixes #1010.

## The problem

`SyncLockGuard::acquire` in `antigravity.rs` and its twin in `trae.rs` both decided ownership from the bytes in a pid file: `create_new` the lock, write the pid, and on a collision read that pid, probe whether it is alive, then unlink and retry when it is not.

That read-decide-unlink sequence is not atomic. Two contenders can find the same dead owner and both enter the critical section. A contender arriving between the `create_new` and the pid write reads an empty file, `read_sync_lock` returns `None`, the liveness check is skipped entirely, and it unlinks a lock a live process has just taken. A recycled pid can make an unrelated process look like the owner. The whole point of the lock — that two syncs never overlap on the manifest and delete each other's session artifacts — does not hold in any of those cases.

## The change

Ownership is now the kernel's exclusive lock on the file, taken with the same `fs2` pattern `autosubmit::try_acquire_run_lock` already uses. That removes the liveness probe and the stale-lock recovery loop along with it, because the kernel releases the lock when the owner dies. `is_lock_contention` became `pub(crate)` and is shared rather than duplicated, so the Windows `ERROR_LOCK_VIOLATION` classification it documents stays in one place.

Two things the issue asked to preserve are preserved. Age-based eviction is not reintroduced — an OS lock makes it unnecessary. The pid is still recorded and still named in the contention message when it is readable and alive; on Windows a handle cannot read a range another handle has locked, so the pid is usually absent there and the message degrades to the plain form.

The lock file is now deliberately left on disk instead of being unlinked on drop. Unlinking it would let a contender create a fresh file and lock that instead, which is the same hole in a different shape. `autosubmit`'s lock already behaves this way.

## Verification

Real surface, same scenario against both binaries: another process holds a genuine exclusive OS lock on `%APPDATA%\tokscale\antigravity-cache\sync.lock` (taken via .NET `FileStream.Lock`, which maps to `LockFileEx`) and has written no pid, then `tokscale antigravity sync` runs.

```console
# before, built from 56b2008c
$ tokscale antigravity sync
Error: Could not acquire Antigravity sync lock after 3 stale-lock recoveries; another process keeps recreating the lock file

# after
$ tokscale antigravity sync
Error: Another tokscale antigravity sync is in progress; aborting
```

The old message is the tell: it made three attempts to evict a lock a live process was holding, and then blamed a phantom process for recreating the file. On Windows the unlink fails because the file is open, so the eviction is merely loud; where unlink of an open file succeeds it takes the lock and both syncs proceed, which is the corruption the lock exists to prevent.

- New regression tests, one pair per lock: `sync_lock_guard_refuses_a_held_lock_that_has_no_pid_written_yet` / `test_acquire_refuses_a_held_lock_that_has_no_pid_written_yet` reproduce the empty-file window and fail on `main` (`acquire` returns `Ok` while another handle holds the lock), and `sync_lock_guard_excludes_a_second_sync_until_the_first_is_dropped` / `test_acquire_excludes_a_second_sync_until_the_first_is_dropped` assert the mutual exclusion the lock exists for.
- `cargo test -p tokscale-cli --bin tokscale` — 992 passed, 1 failed. That one failure, `tui::cache::tests::load_cache_falls_back_to_legacy_dot_cache_path`, also fails on a clean `56b2008c` and belongs to the Windows `HOME`/path group tracked in #1014.
- `cargo clippy -p tokscale-cli --all-targets` produces exactly the same 11 warnings as `56b2008c`, so no new lint debt. It does not pass under `-D warnings` on either revision, for pre-existing unused imports and Windows-inactive helper functions this PR does not touch.
- `cargo fmt --all -- --check` clean.

## Tests that changed shape

Four existing tests asserted the old protocol directly and could not survive it: the two that asserted a pid file alone blocks or releases a lock, and the two that asserted stale-pid eviction. They are replaced, not dropped, by tests of the property the lock is actually for — a held lock is never evicted, an unheld lock file never blocks anyone, and a second sync is refused until the first guard drops. `test_acquire_writes_pid_and_timestamp` keeps its assertion and only reads the pid back after the guard is dropped, since the lock blocks the read on Windows.

---
## Summary by cubic
Move Antigravity and Trae syncs to OS-backed exclusive locks with a companion `sync.os.lock`, while keeping a legacy-compatible `sync.lock` published atomically to avoid overlap during rolling upgrades. `purge-cache` now serializes with the sync to prevent unlink races and never removes a live lock.

- **Bug Fixes**
  - Use `fs2::FileExt::try_lock_exclusive` for the OS lock; share Windows contention detection via `commands/autosubmit::is_lock_contention`.
  - Preserve legacy owners: do not reclaim an existing `sync.lock`; refuse when its recorded pid is alive; keep lock files on disk; on drop, delete only our own record.
  - Publish `sync.lock` atomically (temp + hard link) so legacy binaries never see an empty inode; never overwrite an existing or indeterminate `sync.lock`.
  - Antigravity `purge-cache`: take a parent cache-operation lock, then the sync lock; delete everything except `sync.lock` to prevent unlink races while a sync runs.

- **Migration**
  - During upgrades, tokscale fails closed if a `sync.lock` remains (including after a crash or forced stop). Confirm no `tokscale antigravity sync` or `tokscale trae sync` is running, remove the exact printed `sync.lock` path, then retry; do not remove the lock while a sync may still be running.

<sup>Written for commit a692ec31b125bffe6aaef226412bf643a3f80920. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1030?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MoerAI:
Both P1s were right. Addressed in b0a699b and db83a1b, rebased onto the current `main`.

**Rolling upgrade.** A sync started by a pre-change binary owns the lock through the pid file alone and holds no kernel lock, so the exclusive lock was free while that sync was still writing the manifest. The recorded owner is now read *before* the lock is taken, and a live owner that is not this process is refused. Reading first is also what keeps it working on Windows, which refuses reads of a range once we hold it ourselves. Our own pid is skipped because every successful sync leaves it behind, and a genuine second sync from the same process is already refused by the lock itself. The decision is a pure function over (recorded pid, our pid, liveness predicate), so all four cases are covered without spawning processes.

**purge-cache.** Correct, and it was reachable on any platform where unlinking an open file succeeds: `remove_dir_all` unlinked the inode the running sync held its lock on, so the next sync created and locked a different `sync.lock` and ran alongside the first. Purging now takes the same lock. Verified against the real binary with another process holding the lock — `tokscale antigravity purge-cache` reports `Another tokscale antigravity sync is in progress; aborting`, exits 1, and the cache survives intact; with the lock released the same command deletes it and exits 0.

**Duplication.** Fair, and the rollout guard would have doubled it, so the two implementations are now one shared module parameterised by the wording each command uses. That is why this push touches both files more than the fix alone would.

**Pid diagnostic atomicity (P3).** Keeping as-is deliberately. The pid is a diagnostic, never an ownership signal — exclusion is entirely the kernel lock — so a torn or stale read can only make the message less specific, never let two syncs overlap. The `pid_is_alive` filter already drops a dead stale pid; a recycled live pid would name the wrong process in a message, which I judged not worth another write protocol to close.

Re-verified after the rebase: `cargo test -p tokscale-cli --bin tokscale` is 1003 passed with the same single pre-existing `tui::cache::tests::load_cache_falls_back_to_legacy_dot_cache_path` failure, `cargo clippy -p tokscale-cli --all-targets` produces the identical 11-warning set as `main`, and `cargo fmt --all -- --check` is clean.

Issue #1011, opened 2026-08-03
Claude: `input` adds a char estimate of tool_result content on top of API-reported input_tokens


lizhuojunx86:
## Summary

For Claude Code transcripts, `input` is not the API-reported number. Every
`tool_result` block without explicit token metadata gets
`estimate_tokens_from_chars` — `chars.div_ceil(4)`
(`sessions/claudecode.rs:1118-1126`, `:1206`) — and that estimate is added to
`input`. Claude Code never writes token metadata on tool_result blocks, so the
fallback fires on all of them.

The tool result is then sent back to the model as part of the next turn's
prompt, and that turn's API-reported `input_tokens` / `cache_creation_input_tokens`
/ `cache_read_input_tokens` already account for it. The same content is counted
twice: once by the API, once by the estimate.

This is the hazard your own suppression rule names. The bare-transcript branch
(`claudecode.rs:~458-463`) refuses the char estimate because it "would
double-count usage already tracked by the originating client's own parser."
Same double count, different route — here the originating parser is the API's
own usage field.

## Measured

1,618 transcripts, measured against merged `9ceae64` built from source, isolated
`HOME` per run. A/B: the same corpus with every `tool_result` content emptied,
which zeroes the estimate and changes nothing else.

| | intact | tool_result content emptied |
|---|---|---|
| `input` | 24,793,081 | 3,960,326 |
| `cost` | $7,331.39 | $7,215.86 |

So **20,832,755 of 24,793,081 reported input tokens (84%) are the estimate**,
across 40,068 unique tool results — a 6.26× inflation of the field.

**The cost impact is small: 1.6%** ($115.53). Input tokens are cheap next to
cache and output on this corpus, so the dollar figure barely moves. The token
figure is the misleading one, and it is the one a "tokens used" display shows.

Mechanism check, so the numbers aren't just a black-box difference: computing
`ceil(chars/4)` over unique `(session, tool_use_id)` externally predicts
24,968,754 against the 24,793,081 actually reported — 0.71% apart, the residual
being your cross-file dedup.

## Minimal repro

One assistant message, one `tool_use`, one `tool_result` whose content is 21
characters, `usage.input_tokens = 7`:

```
reported input = 13   (7 from the API + ceil(21/4) = 6)
```

## Possible directions

Yours to pick — both keep explicit tool-result token metadata honored, which is
the case the fallback was written for:

1. Extend the rule you already apply to bare transcripts: for Claude Code
[...]

PR #1037, opened 2026-08-04
fix(claude): stop char-estimating tool_result input tokens

Yuxin-Qiao:
## Summary
`parse_claude_file` no longer char-estimates `tool_result` input tokens for Claude Code transcripts.

Claude Code never writes token metadata on `tool_result` blocks, so the `ceil(chars/4)` fallback fired on all of them, stacking an estimate on top of the API-reported `input_tokens` that the next assistant turn already includes. Measured in #1011: 84% of reported `input` was the estimate (20.8M of 24.8M tokens across 1,618 transcripts).

## Changes
- `allow_char_estimate: false` for Claude Code transcripts (bare and project). Explicit tool-result token metadata is still honored.
- Regression test: API-reported `input_tokens` is no longer inflated by tool_result char estimates (minimal repro from #1011).
- Adapted the synthetic-notice test to carry explicit token metadata so it still verifies an explicit model + explicit metadata survive the synthetic suppression path.

## Rebase note
Rebased onto current main. The branch previously carried Windows test-suite repairs; those were superseded by upstream #1046 (which landed the hermetic fixture work and the separator-canonicalization fix), so this PR is now just the #1011 product fix plus its regression test.

## Test
- `cargo test -p tokscale-core --lib` — 1477 passed, 0 failed
- `cargo test -p tokscale-cli --test cli_tests` — 152 passed, 0 failed

### DIFF
--- a/crates/tokscale-core/src/sessions/claudecode.rs
+++ b/crates/tokscale-core/src/sessions/claudecode.rs
@@ -146,13 +146,6 @@ fn is_workflow_journal(path: &Path) -> bool {
         .any(|ancestor| ancestor.file_name().and_then(|n| n.to_str()) == Some("subagents"))
 }
 
-fn is_in_transcripts_dir(path: &Path) -> bool {
-    path.parent()
-        .and_then(|p| p.file_name())
-        .and_then(|n| n.to_str())
-        == Some("transcripts")
-}
-
 /// Locate the parent main-session JSONL for a sidechain transcript.
 ///
 /// Nested layout: `.../projects/<key>/<session>/subagents/agent-X.jsonl`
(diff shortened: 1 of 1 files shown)


Yuxin-Qiao:
Addressed the review feedback in c90b5180:
- lib.rs cache-reparse test now uses `EnvGuard` for `HOME` + `TOKSCALE_CONFIG_DIR`, so a panic restores the redirect.
- Converted `write_fake_credentials`, `write_cursor_credentials`, and `write_cursor_usage_cache` to `sandbox_config_dir(base)`; on Windows the fixtures now land under the same AppData root `cmd_with_home` pins, instead of `.config/tokscale`.
- Added a `#[cfg(windows)]` regression test asserting `CachedPath` canonicalizes `/` and `\` styles to the same value.

Local: `cargo test -p tokscale-core --lib` targeted tests and the full `cargo test -p tokscale-cli --test cli_tests` suite (152 passed) are green on macOS; the new Windows test will run on the Windows CI legs.

Yuxin-Qiao:
Correction to the previous round: the Windows coverage leg failed on the three converted-writer tests, so I reverted that part in 69164b84 and kept the other two fixes.

Why the revert: the CLI resolves these three roots as hardcoded HOME-relative paths, not through the config root that `TOKSCALE_CONFIG_DIR` redirects:
- `credentials.json` -> `home/.config/tokscale/credentials.json` (tokscale-cli/src/auth.rs:73)
- `cursor-credentials.json` -> `home/.config/tokscale/cursor-credentials.json` (tokscale-cli/src/cursor.rs:35)
- Cursor API cache -> `home/.config/tokscale/cursor-cache` (tokscale-cli/src/cursor.rs:43)

Since `cmd_with_home` redirects `HOME` itself, `.config/tokscale` under the sandboxed home is the right fixture root on every platform, including Windows (the same three tests passed on windows-latest at a512006b with the original writers). Only settings fixtures belong on the pinned root, because settings read through `paths::get_config_dir()` which honors `TOKSCALE_CONFIG_DIR`; `write_settings_json` was already converted to `sandbox_config_dir`.

Kept from the previous round:
- `EnvGuard` for `HOME`/`TOKSCALE_CONFIG_DIR` in the cache-reparse test
- the `#[cfg(windows)]` CachedPath separator regression test (passed on windows-latest in the failing run)

Local: full `cargo test -p tokscale-cli --test cli_tests` suite passes 152/152 on macOS.

junhoyeo:

@Yuxin-Qiao thanks. This is probably the largest single accuracy correction in the release: 84% of reported Claude input was the `ceil(chars/4)` estimate stacked on top of API-reported `input_tokens` (@lizhuojunx86's measurement in #1011 — 20.8M of 24.8M tokens across 1,618 transcripts).

Rebasing the branch down to just the product fix once #1046 superseded the Windows work made this much easier to reason about, and it's why it went in ahead of the others.

One caveat now recorded on #1011: the correction isn't retroactive. `parser_version(Claude)` deliberately wasn't bumped — the table in `message_cache.rs` warns that bumping it discards retained assistant turns that compacted transcripts no longer contain — so warm caches keep their inflated numbers until each transcript changes. Correct decision, just not the whole story for existing users.

Issue #1044, opened 2026-08-04
Submission protocol cannot express unknown cost, so degraded pricing would silently lower recorded spend


junhoyeo:
## Summary

The submission protocol has no way to say "I don't know what this cost", only "this cost is $0.00". Combined with the server overwriting `daily_breakdown.cost` per day, that makes it unsafe for the CLI to submit unpriced usage at zero cost — degraded pricing coverage would silently lower a user's recorded spend.

This blocks the fix direction for #1013, #1021 and #1035, all of which are the same complaint: `tokscale submit` aborts entirely when any token-bearing row has no published price.

## Why the obvious fix is blocked

The natural fix for those issues is to stop aborting and submit unpriced usage at cost `0.0`, reporting what was zeroed. Token counts still reach the leaderboard and no price is invented.

The problem is what happens on the server. `packages/frontend/src/app/api/submit/route.ts` upserts the daily row with an unconditional overwrite:

```sql
ON CONFLICT (submission_id, submitted_device_id, date) DO UPDATE SET
  tokens = EXCLUDED.tokens,
  cost = EXCLUDED.cost,
  ...
  active_time_ms = GREATEST(daily_breakdown.active_time_ms, EXCLUDED.active_time_ms),
```

(and the same `cost = batch.cost` in the batch `UPDATE` below it).

So if pricing coverage degrades — one of LiteLLM / OpenRouter / models.dev fails to load, a cache expires, a corporate proxy blocks the fetch — every row the surviving datasets cannot cover is submitted at `0.0` and **overwrites the cost the server already recorded for that day**. A transient network failure permanently lowers the user's spend, with only a warning line on stdout to show for it.

Note that `active_time_ms` on the very next line already has a monotonic guard for exactly this class of hazard. `cost` does not.

## Why a client-side guard does not close it

We attempted this purely in the CLI across three review rounds. Each round closed its target case and left a narrower one open:

1. Guard on "no pricing data loaded at all" → a **partial** outage (one dataset up, two down) still zeroes everything the survivor cannot cover.
2. Guard on "more than half of price-needing token volume is uncoverable" → a **provider-scoped** outage is diluted below the threshold by other providers' priced usage and passes.
3. Any per-submission ratio is the wrong shape anyway, because the server overwrites **per day**. A submission spanning a long range where one day is entirely uncoverable and the rest is priced passes any global check and still zeroes that day.

[...]

Issue #1035, opened 2026-08-04
submit fails: pricing unavailable for friendli/LGAI-EXAONE/K-EXAONE-236B-A23B, freerouter/nvidia/nemotron-3-ultra-550b-a55b, antigravity/unknown


minpeter:
## Bug

`tokscale submit` aborts with `Error: pricing is unavailable for submitted token usage` when local session data contains models that have no pricing entry. The submission fails entirely instead of skipping/excluding the unpriced models.

## Models triggering the error (deduplicated)

- `friendli/LGAI-EXAONE/K-EXAONE-236B-A23B` (FriendliAI-hosted K-EXAONE, used via oh-my-openagent / senpi)
- `freerouter/nvidia/nemotron-3-ultra-550b-a55b` (free OpenRouter route)
- `antigravity/unknown` (Antigravity sessions where the model id is not resolved)

The raw error repeats each model id once per session/message, making the output very noisy — the list should be deduplicated before printing.

## Expected behavior

Same class as #1013 (closed) and #1021 (open): submission should either

1. exclude unpriced/free models and submit the rest, or
2. submit them with null/zero cost,

rather than aborting the whole submit. And the error message should list each unpriced model once.

## Environment

- tokscale 4.9.0 (via `bunx tokscale@latest`)
- Linux x64
- Clients involved: senpi/oh-my-openagent (friendli, freerouter providers), Antigravity

minpeter:
<img width="1168" height="291" alt="Image" src="https://github.com/user-attachments/assets/812db691-8bfb-4dd0-a257-8c79ed188382" />

TomaszJanusz:
@junhoyeo I have an almost identical problem with the following models:

* `openai/gpt-5.1-codex`
* `openai/gpt-5.1-codex-max`
* `openai/codex-auto-review`

I will not submit a separate issue since they only differ in identifiers.

haunchen:
I traced one of the three ids here — `antigravity/unknown` — to a concrete parser cause, and opened #1051 for it. Posting the root cause since this thread has the symptom but not the mechanism.

Antigravity CLI does not always write `chatModel.#19` (responseModel) into the `GeneratorMetadata` blob. `parse_gen_metadata` in `sessions/antigravity_cli.rs` falls back to the literal string `unknown` when the field is absent, and `inferred_provider_from_model` cannot place `unknown`, so the provider falls back to `antigravity` and the row becomes `antigravity/unknown`.

[...]

PR #1051, opened 2026-08-05
fix(antigravity-cli): attribute turns whose blob omits responseModel

haunchen:
## Problem

`tokscale submit` aborted on every run against my local data:

```
Error: pricing is unavailable for submitted token usage: antigravity/unknown (x8)
```

Eight rows out of 634 token-bearing generations, spread over three of 222 conversations, blocked the entire submission. `validate_priced_messages` rejects the whole batch when any token-bearing message has neither an authoritative cost nor a pricing hit, so a single unattributable row is enough.

## Root cause

`parse_gen_metadata` reads the model from `chatModel.#19` (responseModel) and falls back to the literal string `unknown`. `inferred_provider_from_model` cannot place `unknown`, so the provider falls back to `antigravity` and the row becomes the unpriceable `antigravity/unknown`.

Antigravity CLI does not always write `#19`. Decoding every `gen_metadata` blob in my conversations directory with the module's own wire-format reader shows the field missing on some continuation and tool turns — the ones that also carry a large `cacheRead` and a tiny output. Those rows drop a whole group of chatModel fields (`1, 2, 8, 10, 16, 19`), not `#19` alone.

The rows are not information-poor, though. `#21` — the model display label — survives on every one of them, and sibling rows in the same database carry the machine id next to the identical label:

```
2b38a2ed…  idx 0-12  #19="gemini-3-flash-a"  #21="Gemini 3.5 Flash (High)"
           idx 13    #19=<missing>          #21="Gemini 3.5 Flash (High)"
758f622f…  idx 0-3   #19="gemini-3.6-flash"  #21="Gemini 3.6 Flash (High)"
           idx 4-7   #19=<missing>          #21="Gemini 3.6 Flash (High)"
affd2a9d…  idx 0-8   #19="gemini-3.6-flash"  #21="Gemini 3.6 Flash (High)"
           idx 9-11  #19=<missing>          #21="Gemini 3.6 Flash (High)"
           idx 12    #19="gemini-3.6-flash"  #21="Gemini 3.6 Flash (High)"
```

## Fix

Index the conversation before parsing it, then resolve a missing `#19` through that index:

1. the row's own `#19`;
2. the `#19` seen elsewhere in the same file next to the row's `#21` display label;
3. the file's sole `#19`, when the row carries no label at all and every row that has one agrees;
4. otherwise `unknown`, as before.

A label that identifies no model, or two different ones, resolves to nothing and the row still reports `unknown` rather than borrowing an id the evidence contradicts. Case 3 is the same idea as `antigravity.rs`'s existing `session_model` fallback, which the CLI parser never had.

**Display labels stay out of pricing.** The value handed back is always a `#19` machine id read from that same file; `#21` is only ever a join key between rows of one conversation. This keeps the rule the alias table already states — labels get renamed (`Gemini 3 Flash` → `Gemini 3.5 Flash (High)`) and could be localized, so they must not become pricing keys. The data shows why a label-to-id table would be the wrong shape here anyway: `Gemini 3.5 Flash (High)` is the label for the wire string `gemini-3-flash-a`, and the two track each other only through the alias table.

Indexing the file up front rather than carrying the previous row's model forward also makes recovery independent of row order, and keeps it correct across a mid-conversation model switch — a label that never appeared next to a `#19` resolves to nothing instead of inheriting whatever ran last.

## Testing

Six unit tests cover recovery by label, recovery when the gap is the first row, alias resolution of a recovered id, a label the file never identified, a label claimed by two models, and the no-label single-model and multi-model cases. They build minimal proto fixtures with the existing `enc_varint` / `enc_len` helpers — no real databases are checked in.

`cargo fmt --all -- --check` is clean, and `cargo test --workspace --all-features --no-fail-fast` matches the pre-change baseline on every target (`tokscale-core --lib` goes from 1476 to 1482 for the new tests). My Windows checkout has some pre-existing failures unrelated to this change — the `tui::cache` legacy-path test, the timezone group in `cli_tests`, and one Clippy `unneeded return` in a `#[cfg(target_os = "windows")]` block in `clients.rs` — all of which reproduce identically on an unmodified tree at 4.10.0.

End to end against real local data: with the three affected conversations in place, an unmodified build reproduces `antigravity/unknown (x8)`; with this change, `tokscale --no-spinner submit --dry-run -c antigravity-cli` succeeds, and so does a full-history dry run across every client.

## Relation to #1035 and #1044

This removes one concrete cause of unpriced rows rather than changing what submit does when it meets one. #1035 lists two other unpriced ids that this does not touch, and the broader "don't abort the whole batch" request there and in #1021 is blocked on #1044 — so this is orthogonal to both and closes neither.

---
## Summary by cubic
Fixes unpriced rows in `antigravity-cli` by recovering missing model IDs, so `tokscale submit` no longer aborts on `antigravity/unknown`. Adds a safety guard so the “single-model” fallback is used only when every label in the file is identified.

- **Bug Fixes**
  - Indexes each conversation to recover missing `chatModel.#19` via its `#21` display label from the same file; if a row has no label, uses the file’s sole `#19` only when every label resolves; otherwise returns `unknown`. `#21` is optional and rows with neither field are handled.
  - Uses display labels only as join keys; pricing always uses the recovered machine ID, and alias resolution still applies.
  - Order-independent and safe across model switches; ambiguous or unidentified labels, or rows with no model fields, return `unknown` rather than guessing.
  - Adds unit tests for label-based recovery, first-row gaps, alias resolution, ambiguous/unidentified labels, and safe single-/multi-model fallbacks.

<sup>Written for commit 73d5937dadd6b8dbecde42199edb7ebc3fc480ab. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1051?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/antigravity_cli.rs
+++ b/crates/tokscale-core/src/sessions/antigravity_cli.rs
@@ -20,7 +20,7 @@
 //!
 //! - `gen_metadata.#1`            → chatModel message
 //!   - `#19` (string, optional)  → responseModel (e.g. `gemini-3-flash-a`)
-//!   - `#21` (string)            → model display label (`Gemini 3.6 Flash (High)`)
+//!   - `#21` (string, optional)  → model display label (`Gemini 3.6 Flash (High)`)
 //!   - `#9.#4` = `{#1: seconds, #2: nanos}` → per-generation wall-clock time
 //!   - `#4`                      → usage message
 //!     - `#1` (varint, const)    → fixed system-prompt tokens (≈1132)
@@ -34,10 +34,12 @@
 //!
 //! `#19` is optional in practice: some continuation turns omit it while still
 //! writing `#21`. [`SessionModels`] recovers the machine id for those rows from
-//! the rest of the same conversation. `#21` serves only as a join key between
-//! rows of one file and is never used as a pricing key — it is a server-supplied
-//! name that gets renamed (`Gemini 3 Flash` → `Gemini 3.5 Flash (High)`) and
-//! could be localized.
+//! the rest of the same conversation. `#21` was present on every row observed so
+//! far, including the ones missing `#19`, but nothing here requires it — a row
+//! carrying neither field is handled too. `#21` serves only as a join key
+//! between rows of one file and is never used as a pricing key: it is a
+//! server-supplied name that gets renamed (`Gemini 3 Flash` → `Gemini 3.5 Flash
+//! (High)`) and could be localized.
 
 use super::utils::open_readonly_sqlite;
 use super::{normalize_workspace_key, workspace_label_from_key, UnifiedMessage};
@@ -118,7 +120,8 @@ struct SessionModels {
     /// that appear with more than one id are dropped: an ambiguous label is no
     /// better evidence than no label at all.
     by_display: HashMap<String, String>,
-    /// The file's only `#19` value, when every row carrying one agrees. Serves
+    /// The file's only `#19` value — set only when every row carrying one
+    /// agrees *and* every label in the file was identified by some row. Serves
     /// rows that have neither `#19` nor `#21`.
     sole_model: Option<String>,
 }
@@ -127,16 +130,21 @@ impl SessionModels {
     fn from_blobs(blobs: &[Vec<u8>]) -> Self {
         let mut by_display: HashMap<&str, Option<&str>> = HashMap::new();
         let mut distinct: HashSet<&str> = HashSet::new();
+        let mut unresolved_labels: Vec<&str> = Vec::new();
 
         for blob in blobs {
             let Some(chat_model) = message_field(blob, 1) else {
                 continue;
             };
+            let label = non_empty_string_field(chat_model, 21);
             let Some(model) = non_empty_string_field(chat_model, 19) else {
+                // Kept so `sole_model` below can tell whether this row was
+                // identified by some other row carrying the same label.
+                unresolved_labels.extend(label);
                 continue;
             };
             distinct.insert(model);
-            if let Some(label) = non_empty_string_field(chat_model, 21) {
+            if let Some(label) = label {
                 by_display
                     .entry(label)
                     .and_modify(|resolved| {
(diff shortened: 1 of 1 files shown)


haunchen:
Both valid, fixed in 73d5937d.

The second one is a real hole and it was my own rule applied inconsistently. `sole_model` counted *named* ids, not models: a conversation where one row carries a display label no row ever identifies ran at least two models while naming only one, so an unlabelled row there could be either — and it inherited the single named id. The labelled path already refused to guess in exactly that situation; the unlabelled one did not. The fallback is now gated on every label in the file resolving to a machine id, with a test covering the named-Flash / unidentified-Pro / no-fields case.

On the first: `#21` is now marked optional too. It was present on every row I observed, including all the ones missing `#19`, but the parser has never required it and the `sole_model` path exists precisely for rows carrying neither field — so the table implying otherwise was misleading. The prose now says both things.

channprj:
Looks good!

junhoyeo:

@haunchen @channprj thanks. Recovering the model from the `chatModel #21` display label — after noticing that Antigravity drops a whole field group (`1, 2, 8, 10, 16, 19`) on continuation and tool turns while `#21` survives on every one — was a genuinely good piece of reverse engineering, and the decoding evidence in the description made it easy to trust.

Worth calling out how you handled the review: the P1 about a conversation containing a model switch billing under the wrong model was a real hole, and `73d5937d` withholds the single-model fallback when any label is unresolved rather than papering over it. That's the right instinct.

One gap that came out of triaging #1035 afterwards: this covers the **CLI** parser only. `sessions/antigravity.rs:61` still has the bare `unknown` fallback for the IDE-backed path, so IDE-sourced rows are unaffected. I've asked the reporters there which path their data comes from before filing it separately — if you already know the IDE format well enough to say whether the same `#21` trick applies, that'd be useful.

Also relevant now: post-#1053 an unattributable row is excluded with a warning instead of aborting the submission, so a regression here would be quiet rather than loud.

PR #1052, opened 2026-08-05
feat(core): support TOKSCALE_SKIP_EXTERNAL_PRICING for submissions

pengtianabc:
Adds support for the `TOKSCALE_SKIP_EXTERNAL_PRICING` environment variable so that `tokscale submit` can succeed even when upstream pricing sources (LiteLLM, OpenRouter, models.dev) do not cover every model used, such as private or internal models.

When `TOKSCALE_SKIP_EXTERNAL_PRICING` is set to `1`, `true`, or `yes`:
- The submit path no longer fetches external pricing datasets over the network.
- Only `custom-pricing.json` is loaded.
- Messages whose model is not priced are reported with `cost = 0` instead of failing the whole submission with `pricing is unavailable for submitted token usage`.

The default behavior is unchanged: without the variable, `submit` still validates that every token-bearing message is covered by a known price.

A new unit test covers the skip path with a deliberately unpriced model, asserting that the graph builds successfully and costs are zero.

While preparing the test suite, two existing tests that call `build_graph_from_messages` directly were updated to pass the now-required `bucket_timezone` argument; this is a minimal compile fix with no production behavior change.

---
## Summary by cubic
Make `tokscale submit` lenient by default so unpriced models no longer block submissions. Unpriced tokens are submitted at zero cost with warnings, and validated provider‑reported costs are preserved; strict pricing is opt‑in.

- **New Features**
  - Default lenient pricing: unpriced usage stays in the graph and is reported as cost = 0 with a provider/model warning; authoritative costs from parsers like `trae`, `hermes`, `warp`, `codebuff`, `kilo`, `amp`, `mux`, `openclaw`, `crush`, `cursor`, and `roocode` are preserved.
  - Strict opt-in: `--strict-pricing` or `TOKSCALE_STRICT_PRICING` requires priced models and fails otherwise; strict overrides `TOKSCALE_SKIP_EXTERNAL_PRICING`.
  - `TOKSCALE_SKIP_EXTERNAL_PRICING`: load only `custom-pricing.json` (no network) and allow submissions with unpriced usage at zero cost.

- **Bug Fixes**
  - Tightened authoritative cost detection and flag preservation: only mark provider‑reported when raw costs are present, finite, and non‑negative; keep explicit zero when reported (e.g., `amp`, `roocode`), and ignore invalid/negative values (`trae`, `kilo`, `openclaw`, `mux`, `warp`, `crush`, `codebuff`). For `cursor`, only the old CSV without a Kind column treats positive‑cost rows as authoritative; new formats never do. Cache versions for `roocode`/`kilo` were bumped to reparse and retain correct flags.

<sup>Written for commit 3ccca4033344f1245f61e58d927b57b42f97bab6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1052?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -267,6 +267,8 @@ enum Commands {
             help = "Show what would be submitted without actually submitting"
         )]
         dry_run: bool,
+        #[arg(long, help = "Fail submission if any model lacks a known price")]
+        strict_pricing: bool,
     },
     #[command(about = "Manage periodic usage submission")]
     Autosubmit {
@@ -807,6 +809,7 @@ fn main() -> Result<()> {
             clients,
             date,
             dry_run,
+            strict_pricing,
         }) => {
             reject_unsupported_home_override(&cli.home, "submit")?;
             let (since, until) = build_date_filter(&date, &cli.home);
@@ -823,6 +826,7 @@ fn main() -> Result<()> {
                 until,
                 year,
                 dry_run,
+                strict_pricing,
                 SubmitMode::Interactive,
             )
         }
@@ -5505,7 +5509,7 @@ fn report_unpriced_submission_exclusions(
         println!(
             "{}",
             format!(
-                "  Warning: excluded {} unpriced {}/{} message(s) ({} tokens): {}.{}",
+                "  Warning: skipped pricing for {} unpriced {}/{} message(s) ({} tokens): {}.{}",
                 row.message_count,
                 row.provider_id,
                 row.model_id,
@@ -5555,7 +5559,7 @@ fn run_autosubmit_command(subcommand: commands::autosubmit::AutosubmitSubcommand
             };
 
             let (clients, since, until, year) = commands::autosubmit::submit_filters(&settings);
-            match run_submit_command(clients, since, until, year, false, SubmitMode::Autosubmit) {
+            match run_submit_command(clients, since, until, year, false, false, SubmitMode::Autosubmit) {
                 Ok(()) => {
                     commands::autosubmit::record_run_success(
                         chrono::Utc::now().timestamp_millis(),
@@ -5578,12 +5582,19 @@ fn run_submit_command(
     until: Option<String>,
     year: Option<String>,
     dry_run: bool,
+    strict_pricing: bool,
     mode: SubmitMode,
 ) -> Result<()> {
     use colored::Colorize;
     use std::io::IsTerminal;
     use tokio::runtime::Runtime;
     use tokscale_core::{generate_submission_graph, GroupBy, ReportOptions};
+
+    // 命令行参数或环境变量均可开启严格价格校验；默认使用宽松模式。
+    let strict_pricing = strict_pricing
+        || std::env::var("TOKSCALE_STRICT_PRICING")
+            .map(|v| matches!(v.as_str(), "1" | "true" | "TRUE" | "yes" | "YES"))
+            .unwrap_or(false);
 
     let auth_token = match auth::resolve_api_token() {
         Some(token) => token,
@@ -5651,16 +5662,19 @@ fn run_submit_command(
     let rt = Runtime::new()?;
     let mut graph_result = rt
         .block_on(async {
-            generate_submission_graph(ReportOptions {
-                home_dir: None,
-                use_env_roots: true,
-                clients,
-                since,
-                until,
-                year,
-                group_by: GroupBy::default(),
-                scanner_settings: tui::settings::load_scanner_settings(),
-            })
+            generate_submission_graph(
+                ReportOptions {
+                    home_dir: None,
+                    use_env_roots: true,
+                    clients,
+                    since,
+                    until,
+                    year,
+                    group_by: GroupBy::default(),
+                    scanner_settings: tui::settings::load_scanner_settings(),
+                },
+                strict_pricing,
+            )
             .await
         })
         .map_err(|e| anyhow::anyhow!(e))?;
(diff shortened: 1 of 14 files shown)


pengtianabc:
Review fails to converge: 3 rounds of revisions completed since initial review, new issues keep emerging.

junhoyeo:
@pengtianabc thanks for this, and genuinely for the four rounds of revisions — you also asked on #1053 whether this one was better, which was the right question to force. I've gone with #1053 for the abort fix. Two points on why, so the decision isn't opaque:

**1. The `cost = 0.0` default is blocked by an open design issue, not by code quality.** #1044 documents that the submission protocol has no way to express "I don't know what this cost" — only "this cost is $0.00" — while `packages/frontend/src/app/api/submit/route.ts` upserts `cost = EXCLUDED.cost` unconditionally. So unpriced usage submitted at zero overwrites the cost already recorded for that day, and any degradation in pricing coverage (one dataset failing to load, a stale cache, a proxy blocking the fetch) permanently lowers a user's recorded spend. #1044 also walks through three client-side guards we tried and abandoned, so this isn't something a CLI-side refinement can close. Making it the default in `submit` widens the exposure to everyone rather than to people who opted in. This is a sequencing problem — the `costIsComplete` flag and migration in #1044 need to land server-side first.

**2. Scope and verification.** The PR grew well past its title: 14 files, +538/−67, rewriting authoritative-cost detection across 11 parsers with two cache-version bumps that force a full reparse for RooCode/Kilo users. That's a second, riskier change riding along with the fix, and it's what the review kept finding new P1s in — the Trae and Hermes findings were about erasing genuine provider-reported costs, which is the same class of silent data loss as point 1. Your own comment that review wasn't converging is a fair read; the surface was just too large to stabilize in one PR. On top of that, the build/test/lint checks never ran here (first-time-contributor approval gate), so none of the parser work is verified on any of the 9 targets.

One piece I do want, though: **`TOKSCALE_SKIP_EXTERNAL_PRICING`** — custom-pricing.json only, no network — is a real gap for offline and private/internal models, and it's orthogonal to the lenient-vs-exclude question. As a standalone PR that only adds the env var and the dataset-loading skip (no default change, no parser edits) it's maybe 40 lines, and I'd review that quickly. The `EnvGuard` fix you already made in response to review would carry over as-is.

Closing this one in favour of #1053. Please do open the skip-external PR if you're up for it.

junhoyeo:
@pengtianabc — correcting myself on the last paragraph above, before you spend time on it. I shouldn't have invited a standalone `TOKSCALE_SKIP_EXTERNAL_PRICING` PR. We're deliberately holding the line on new environment variables, and I gave you the opposite impression.

Two things I should have checked before suggesting it:

**1. The switch already exists.** `TOKSCALE_PRICING_CACHE_ONLY` (`crates/tokscale-core/src/lib.rs`, in `load_pricing_for_local_parse`) is the same shape of knob — skip the network, use what's on disk. Adding `TOKSCALE_SKIP_EXTERNAL_PRICING` next to it would give us two near-duplicate ways to say "don't fetch pricing", differing only in whether the cached datasets or just `custom-pricing.json` survive. That distinction isn't worth a permanent public env var, and every one of these we add is one more thing that has to keep working forever.

**2. The actual gap is narrower than a new variable.** `generate_submission_graph` calls `PricingService::get_or_init().await?` directly rather than going through `load_pricing_for_local_parse`, so the existing `TOKSCALE_PRICING_CACHE_ONLY` never reaches the submit path at all. That's the real defect behind your use case — not a missing switch, but an existing switch that submit doesn't honor. If you want to pick this up, routing submit through the same loader so `TOKSCALE_PRICING_CACHE_ONLY` applies there too would be welcome, and it's smaller than what you'd have written.

Separately, for private/internal models specifically: `custom-pricing.json` is consulted ahead of every external dataset, so that need is already served today without any env var — there's a worked example in https://github.com/junhoyeo/tokscale/issues/1013#issuecomment-5177200628, including the constraint that a row must price every token bucket your messages actually populate.

Sorry for the churn. The #1053 decision itself is unchanged.

pengtianabc:
> @pengtianabc — correcting myself on the last paragraph above, before you spend time on it. I shouldn't have invited a standalone `TOKSCALE_SKIP_EXTERNAL_PRICING` PR. We're deliberately holding the line on new environment variables, and I gave you the opposite impression.
> 
> Two things I should have checked before suggesting it:
> 
> **1. The switch already exists.** `TOKSCALE_PRICING_CACHE_ONLY` (`crates/tokscale-core/src/lib.rs`, in `load_pricing_for_local_parse`) is the same shape of knob — skip the network, use what's on disk. Adding `TOKSCALE_SKIP_EXTERNAL_PRICING` next to it would give us two near-duplicate ways to say "don't fetch pricing", differing only in whether the cached datasets or just `custom-pricing.json` survive. That distinction isn't worth a permanent public env var, and every one of these we add is one more thing that has to keep working forever.
> 
> **2. The actual gap is narrower than a new variable.** `generate_submission_graph` calls `PricingService::get_or_init().await?` directly rather than going through `load_pricing_for_local_parse`, so the existing `TOKSCALE_PRICING_CACHE_ONLY` never reaches the submit path at all. That's the real defect behind your use case — not a missing switch, but an existing switch that submit doesn't honor. If you want to pick this up, routing submit through the same loader so `TOKSCALE_PRICING_CACHE_ONLY` applies there too would be welcome, and it's smaller than what you'd have written.
> 
> Separately, for private/internal models specifically: `custom-pricing.json` is consulted ahead of every external dataset, so that need is already served today without any env var — there's a worked example in [#1013 (comment)](https://github.com/junhoyeo/tokscale/issues/1013#issuecomment-5177200628), including the constraint that a row must price every token bucket your messages actually populate.
> 
> Sorry for the churn. The #1053 decision itself is unchanged.

Accepted.

PR #1058, opened 2026-08-06
feat(submit): let a submission declare its cost incomplete

junhoyeo:
Implements the server half of #1044.

## Problem

The submission protocol has no way to say *"I don't know what this cost"* — only *"this cost is $0.00"*. Both `daily_breakdown` write paths overwrote `cost` unconditionally:

```sql
ON CONFLICT (submission_id, submitted_device_id, date) DO UPDATE SET
  cost = EXCLUDED.cost,
  ...
  active_time_ms = GREATEST(daily_breakdown.active_time_ms, EXCLUDED.active_time_ms),
```

So any degradation in a client's pricing coverage — one dataset failing to load, a stale cache, a corporate proxy blocking the fetch — permanently lowered that day's recorded spend. `active_time_ms` on the very next line already had a monotonic guard for exactly this hazard. `cost` did not.

#1044 also records three client-side guards that were tried and abandoned; each closed its target case and left a narrower one open. The client is being asked a question it cannot answer — *is my pricing good enough to overwrite what the server already has?* Only the server knows what it has.

## Change

`daily_breakdown.cost_is_complete boolean not null default true` records whether a stored cost accounts for every token counted that day.

- **Payload**: `totals.costIsComplete` is optional; absent means complete.
- **Both write paths** (the `ON CONFLICT` arm *and* the batch `UPDATE` that #1044 named explicitly):

```sql
cost = CASE
         WHEN EXCLUDED.cost_is_complete THEN EXCLUDED.cost
         ELSE GREATEST(daily_breakdown.cost, EXCLUDED.cost)
       END,
```

The stored flag describes the cost **actually kept**, not the last writer, so it can't claim a number is unreliable when the guard just preserved a complete one:

```sql
cost_is_complete = CASE
                     WHEN EXCLUDED.cost_is_complete THEN true
                     WHEN EXCLUDED.cost > daily_breakdown.cost THEN false
                     ELSE daily_breakdown.cost_is_complete
                   END,
```

## Compatibility and deployment

Server-only and backwards compatible. The field is optional and absent means complete, so **every released CLI is unaffected** and keeps exact overwrite semantics, downward corrections included. Migration is a single `ADD COLUMN ... DEFAULT true NOT NULL` — no table rewrite on modern Postgres — generated with `drizzle-kit generate` (journal entry `idx: 26`, not hand-edited).

Per #1044, deployment is ordered: **this first, then the CLI release that sets the flag.** A CLI sending the flag against an older server would be silently ignored.

## Tests

- `bun run test` — 826 passed (84 files), up from 822
- `bun run typecheck` clean; `bun run lint` 0 errors
- **Mutation check:** removing the schema field fails `preserves a declared-incomplete flag through validation` and `rejects a non-boolean flag rather than coercing it`. The parsed value is asserted, not just validity — Zod strips unknown keys, so a validity-only assertion would pass against a schema that never declared the field and silently dropped it.

I did not run `test:migrations` locally: it invokes `drizzle-kit migrate`, which **applies** migrations, and `DATABASE_URL` here points at a real database. CI replays against a scratch `postgres:16`.

## What this does not cover

**The guard's SQL semantics are not exercised against a real Postgres by this PR.** The existing transaction double models `source_breakdown` only, not `cost`, so making it prove the guard would mean reimplementing the `CASE`/`GREATEST` logic inside the double and testing the double. Coverage here is schema-level plus the structural change. A postgres-backed route test is the right follow-up, and I'd rather flag that than imply an invariant is proven when it isn't.

## Trade-off worth challenging

A day flagged incomplete will not accept a legitimate downward correction until the user resubmits with healthy pricing. That's deliberate — over-reporting is recoverable, silently losing recorded spend is not — but it is a real behavior change and worth a second opinion if the leaderboard's semantics argue otherwise.

## Open question from #1044

Kept at day granularity, which is the level the server actually overwrites. Per-client (`ClientContributionSchema`) would record *which* client made the day incomplete — better for debugging and for showing users what to fix — at the cost of a wider schema and more per-row storage. Happy to widen it if you'd prefer.

---
## Summary by cubic
Let submissions and daily rows declare whether costs are complete to stop partial pricing from lowering recorded spend. The floor now applies during merge per client and per nested model (including alias-heal), and completeness composes per client, per day, and at the submission total to keep scalars and JSON consistent (per #1044).

- **New Features**
  - API accepts optional `totals.costIsComplete`; absent defaults to complete.
  - DB adds `daily_breakdown.cost_is_complete` and `submissions.cost_is_complete` (boolean default true).
  - Merge floors incomplete costs per client and per model, derives client totals from model floors, and applies on alias-heal; day completeness derives from the merged breakdown and submission completeness is `BOOL_AND` across days/devices; writes are plain overwrites that match recomputed totals and preserve incomplete siblings on filtered resubmits.

- **Migration**
  - Apply migrations 0026 and 0027.
  - Deploy server first; existing CLIs are unaffected, and newer CLIs sending the flag are ignored by older servers.

<sup>Written for commit 9ecd67fc3d7165f5774553f2cb7dd49ba24ad8a1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1058?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/__tests__/api/submit.test.ts
+++ b/packages/frontend/__tests__/api/submit.test.ts
@@ -567,6 +567,43 @@ describe('POST /api/submit - Client-Level Merge', () => {
       expect(breakdownCostIsComplete(merged)).toBe(true);
     });
 
+    it("floors cost on the alias-heal branch too", () => {
+      // Regression: the heal branch assigned the incoming client raw and
+      // `continue`d, skipping the floor entirely. An incomplete submission
+      // could then drop a folded $20 to $1 AND mark the day complete — the
+      // exact data loss this whole mechanism exists to prevent.
+      //
+      // The heal gate is evidence about tokens, not pricing, so the token
+      // count still heals; only the cost is held at the floor.
+      const { merged } = mergeClientBreakdownsWithRegressionGuard(
+        { kilo: client(200, 20) }, // normalized fold of kilocode+kilo
+        { kilo: client(100, 1) },
+        new Set(["kilo"]),
+        new Map([["kilo", 100]]), // heal floor = largest single component
+        false
+      );
+
+      expect(merged.kilo.tokens).toBe(100); // healed
+      expect(merged.kilo.cost).toBe(20); // floored, not $1
+      expect(merged.kilo.provenance?.costIsComplete).toBe(false);
+      expect(breakdownCostIsComplete(merged)).toBe(false);
+    });
+
+    it("still heals cost exactly when the healing submission is complete", () => {
+      // The floor must not break the heal's original purpose.
+      const { merged } = mergeClientBreakdownsWithRegressionGuard(
+        { kilo: client(200, 20) },
+        { kilo: client(100, 1) },
+        new Set(["kilo"]),
+        new Map([["kilo", 100]]),
+        true
+      );
+
+      expect(merged.kilo.tokens).toBe(100);
+      expect(merged.kilo.cost).toBe(1); // inflated fold replaced outright
+      expect(breakdownCostIsComplete(merged)).toBe(true);
+    });
+
     it("makes a day incomplete when any single client is floored", () => {
       // The submission-level BOOL_AND composes the same way one level up: one
       // floored device makes the account total a lower bound, not an exact
--- a/packages/frontend/__tests__/api/submit.test.ts
+++ b/packages/frontend/__tests__/api/submit.test.ts
@@ -5,8 +5,10 @@ import {
   validateSubmission,
 } from "../../src/lib/validation/submission";
 import {
+  breakdownCostIsComplete,
   deriveClientBreakdownProvenance,
   mergeClientBreakdownsWithRegressionGuard,
+  recalculateDayTotals,
   type ClientBreakdownData,
 } from "../../src/lib/db/helpers";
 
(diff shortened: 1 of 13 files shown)


junhoyeo:
Converting to draft. Adversarial review (Codex) found four real defects in this design, and they are not patchable in place — the guard is in the wrong layer.

**The root problem is defect 4.** On an incomplete decrease the SQL keeps the old scalar via `GREATEST(d.cost, batch.cost)` but replaces `source_breakdown` unconditionally. The scalar and the JSON then disagree, and different readers use different ones: profile daily totals read the scalar (`publicProfileData.ts:524-532`), client costs and filtered leaderboards sum the JSON (`:616-629`, `leaderboard/sourceBreakdown.ts:63-76`). The row's own two representations would contradict each other.

That is fatal to the approach, because the scalar is *derived* from the breakdown by `recalculateDayTotals`. Keeping a scalar the breakdown does not support is incoherent by construction.

**And defect 3 answers the open question this PR asked.** I proposed day-level granularity, noting per-client as an optional refinement. It is not optional. Given a day with incomplete client A and complete client B, a healthy filtered resubmit of only B carries `costIsComplete=true`, and `WHEN batch.cost_is_complete THEN true` marks the whole day complete — while A is preserved and still unpriced. Day-level state cannot represent that.

**Defect 2** is the same class: an incomplete rescan with more tokens and lower cost keeps `$10` and keeps `true`, but `$10` was complete for the *old* token set.

**Defect 5**: `STEP 3d` sums every device's row into `submissions.totalCost` with no completeness composition, so one device's floor silently becomes part of an account total presented as exact.

## Redesign

Move the floor out of SQL and into the client-breakdown merge, where this repo already defends per-client token regressions (`mergeClientBreakdownsWithRegressionGuard`):

- an incomplete submission floors **per-client cost** during the merge, so the recomputed scalar and the stored breakdown agree by construction — defect 4 dissolves
- completeness is carried **per client** in `source_breakdown[client].provenance`, so a filtered resubmit cannot clear another client's state — defect 3
- the day column becomes `AND` over its clients, computed in JS and written unconditionally, which also deletes both `CASE` expressions — defect 2
- `STEP 3d` composes with `BOOL_AND` — defect 5

Migration `0026` stands as-is; the column is still needed, just no longer written by a `CASE`.

Two of Codex's other angles cleared: the `::numeric(14,4)` casts make `GREATEST` numeric rather than lexicographic, and the migration is a fast metadata default on PG11+ (CI and compose are PG16), though it still takes `ACCESS EXCLUSIVE` and production's version is not pinned in the repo.

junhoyeo:
Reworked. The floor now lives in the client-breakdown merge rather than in SQL, which dissolves three of the four defects rather than patching them.

**Defect 4 (scalar vs. JSON divergence)** — gone by construction. The day scalar is recomputed from the merged breakdown by `recalculateDayTotals`, so flooring the breakdown makes the two agree automatically. Both write statements are plain overwrites again and **both `CASE` expressions are deleted**.

**Defect 3 (filtered resubmit clearing a sibling)** — completeness is now carried per client in `source_breakdown[client].provenance.costIsComplete`, and the day column is an AND over the merged clients. A healthy submit naming only client B leaves preserved client A's tag untouched.

This answers the open question the first revision asked: per-client is not a refinement over day-level, it is **required**. Thanks for forcing that.

**Defect 2 (tag can lie)** — the tag no longer depends on which cost won. A floored client stays incomplete even when its own number was kept, because the stored cost still may not cover the tokens now recorded beside it.

**Defect 5 (submission-level `SUM`)** — not addressed here, and called out in the commit's `Not-tested:`. `STEP 3d` still sums every device's cost with no `BOOL_AND`, so an account total can mix a complete and a floored device without saying so. That needs a `submissions` column and is a separate change; I did not want to grow this one further while its core was being redesigned.

Also fixed en route: three test files mocked `@/lib/db/helpers` wholesale, so any new export broke ~21 tests. They now spread the real module.

**Verification** — 832 tests pass (up from 826), typecheck clean, lint 0 errors. Mutation-checked: disabling the floor fails 3 of the 6 new merge tests.

Worth noting the redesign closed the testing gap I flagged on the first revision. The invariant is now exercised in plain JS against the real merge helper, instead of living in SQL that the transaction double (which models `source_breakdown` only, not `cost`) could never reach.

Confidence is **medium**, not high — this is a second design pass on a data-loss guard, and #1058's first version looked right to me too. Still in draft; a second adversarial pass on the new shape before merge would be reasonable.

junhoyeo:
Second adversarial review done (independent model). **One confirmed defect, five angles survived.** Fixed in `e66914d2`.

## The defect — and it was the one worth finding

The alias-fold healing branch assigned the incoming client raw and `continue`d, skipping `applyCostCompleteness` entirely:

```
stored  kilocode=100/$10 + kilo=100/$10  ->  normalized kilo=200/$20 (heal floor 100)
incoming kilo=100 tokens/$1, costIsComplete=false
->  heal accepted, stored $1, NO incomplete tag, day reported COMPLETE
```

An explicitly incomplete submission could cut a day from $20 to $1 **and mark it exact** — precisely the data loss this PR exists to prevent, reintroduced through a branch I did not consider when moving the floor into the merge.

The fix separates the two axes the branch had conflated. Heal eligibility is evidence about **tokens** (incoming ≥ the largest folded component); it says nothing about pricing coverage. So the token count still heals, while the cost goes through the floor and carries the tag. Keeping the folded cost is no worse than declining to heal, and the tag lets a later complete submission correct it exactly.

I rejected the simpler "don't heal at all when incomplete" — it would strand an alias fold permanently for anyone with a genuinely unpriced model.

## What survived

- **fold-preservation writeback** — only deliberately preserved folds get their raw entries back; no accepted replacement or new tag is overwritten
- **complete-path destructuring** — reached only for an accepted incoming client, so it cannot clear an untouched sibling
- **`tagBreakdownCostCompleteness` identity return** — no mutation path connects it to the in-place alias normalization, which operates on cloned data
- **scalar ↔ `source_breakdown` reconcilability** — holds on every path including the new-day INSERT and the fold writeback
- **`BOOL_AND` across devices** — same `submissionId` filter as the cost SUM; non-sticky recompute is correct

## Verification

835 tests pass (up from 833), typecheck clean, lint 0 errors. Mutation-checked: reverting the one-line branch change fails `floors cost on the alias-heal branch too` and nothing else.

## Assessment

Two review rounds, five defects then one. The remaining one was in a branch neither I nor the first round examined, which is the argument for having run the second pass rather than merging on green CI — this PR's first version was also green.

Still in draft. This is now the second time an early-return branch bypassed the floor, so I've left a `Directive:` trailer on the commit saying any future branch in that loop must route through `applyCostCompleteness`.

junhoyeo:
All three findings fixed in `9ecd67fc`.

## P1 — nested model costs (both bots found this independently)

Correct, and my previous fix stopped one level short. There are three levels and only two were consistent:

```
day.cost              = Σ client.cost         floored
client.cost           = max(existing, next)   floored
client.models[*].cost                         NOT floored  ← passed straight through
```

So a `model:`-filtered leaderboard summed the incomplete payload's nested costs and ranked usage at $0 while the client aggregate said $10 — the same scalar-vs-JSON divergence the merge-layer redesign was meant to eliminate, one level down. I claimed it was "gone by construction"; that was true for day↔client and false for client↔models.

Every model in the union is now floored, including one the incomplete payload dropped entirely — its usage did not stop existing because this submission could not price it.

The client total is **derived** from those floors rather than taken as `max(existing, next)`. Deriving is what makes the levels agree by construction: taking the max independently can land *between* them — existing `{A:6,B:4}=10` against incoming `{A:0,C:5}=5` floors to `{A:6,B:4,C:5}=15`, which `max(10,5)=10` would contradict.

## P3 — bound parameters

Right, 11 now. Both chunking comments said 10, and that number is what `INSERT_CHUNK_SIZE` is reasoned from, so the stale value was the actively misleading part. Fixed in both places.

## P3 — the vacuous test

Also right, and worth naming: it set summary and day cost to the same 12.5, and the mismatch check never reads `costIsComplete`, so it passed whether or not the flag did anything. It now sends a genuinely inconsistent summary and asserts the check still rejects — i.e. that the flag does not become a licence to send inconsistent numbers.

That is the second vacuous test I have written on this PR. The first (the Zod `accepts` cases) I caught by mutation-checking; this one I did not mutation-check. The three new nested-model tests were mutation-checked — all three fail with the floor removed.

838 tests pass, typecheck and lint clean.

Issue #1062, opened 2026-08-07
Fuzzy model matching attaches another model's rates instead of excluding, so unpriced usage submits at wrong or $0 cost


junhoyeo:
Found while triaging #1035 and #1021 against `main` (`2e49fb0c`). Not reported by either reporter — it is what happens *instead of* the failure they reported, and it is worse, because it is silent.

## Problem

Pricing lookup falls back to fuzzy matching. When a model has no real entry, the fuzzy match can land on an unrelated model and the row is then treated as **priced** — so #1053's exclusion never fires and the row submits with someone else's rates.

Two concrete cases, both verified with `tokscale pricing` against the live cached datasets:

**1. Zero-rate match — real tokens at $0.**

```
$ tokscale pricing nvidia/nemotron-3-ultra-550b-a55b
  Matched key: kenari/nemotron-3-ultra-550b-a55b
  Input: $0.00 / 1M    Output: $0.00 / 1M    (Models.dev)
```

`kenari/…` is a different provider's row at $0.00/$0.00. Since `55635ef4` an all-zero row "covers any usage shape" (`crates/tokscale-core/src/pricing/litellm.rs:138-140`), so `covers_usage_with_provider` returns true and the row is **not** excluded. It submits real token counts at $0.

**2. Wrong-rate match — real tokens at another model's price.**

```
$ tokscale pricing codex-auto-review
  Matched key: openai/gpt-5.1-codex-mini
  Input: $0.25 / 1M    Output: $2.00 / 1M
```

`openai/codex-auto-review` is billed at `gpt-5.1-codex-mini` rates. Nothing indicates the attribution is a guess.

## Why this matters more after #1053

Before #1053 an unpriceable row aborted the submission — loudly wrong, but wrong in a way you noticed. Now genuinely-unpriced rows are excluded with a named warning, which is honest. But a **fuzzy-matched** row takes neither path: it is silently accepted as priced. The failure mode moved from "blocks your submit" to "quietly records the wrong number".

It also collides with #1044. A `$0` row from case 1 is indistinguishable from a real free model, so the completeness flag that work introduces would mark it complete.

## Suggested direction

Not obvious, hence no PR:

1. Require the fuzzy match to agree on provider, or on a normalized model identity, before accepting it for **submission** (reporting can stay lenient — that split already exists via `GraphPricingRequirement`).
2. Treat a fuzzy match as not-covering for submission, so it flows into #1053's exclusion path with a clear reason.
[...]

PR #1066, opened 2026-08-07
docs(ci): describe the windows leg as it is today, not as it was on 08-03

NickAme03:
Item 3 of the remaining work listed in #1014: the comment block above the
`tests` job still describes the state of 2026-08-03.

It currently says the six root causes are open, that 33 failures are
outstanding, and that two of the groups "look like real Windows defects rather
than test artefacts". All three are now false. The leg is green on main, and
those two groups were traced to fixtures that built their JSON by interpolating
a path into a string literal, which produces unescaped backslashes and an
unparseable fixture on Windows only.

That matters beyond tidiness, for the reason given in the issue: this comment is
what a future reader consults before touching the leg.

What changed:

- the six groups are described as addressed, with the current numbers
- the two suspected product defects are described as what they turned out to be,
  with the mechanism in one sentence rather than a bare pointer
- the condition for flipping the gate is now stated: the leg stays advisory
  until the two un-redirected config roots are swept

What did not change: `continue-on-error` is untouched, and so is the closing
instruction to flip it when #1014 closes. Sweeping the config roots is item 1 on
your list and comes before the flip, so this PR does not pre-empt either.

No code, one comment block.

---
## Summary by cubic
Update the Windows test leg comment in `.github/workflows/test_coverage.yml` to match the current state and gate policy. It notes all six groups from #1014 are resolved and the leg is green on `main` (2,679 passed, 0 failed), clarifies the two suspected defects were fixture artefacts (`serde_json` rejecting unescaped Windows backslashes), and states to keep `continue-on-error` until the two un-redirected config roots are swept, then flip it when #1014 closes.

<sup>Written for commit da7e8311a2eb7ebd9aca04359c33c845b76b3793. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1066?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -57,16 +57,21 @@ jobs:
   #
   # The windows leg is advisory (`continue-on-error`) and Ubuntu stays a hard
   # gate. That split is deliberate and temporary. Running the suite for the
-  # first time found 33 pre-existing failures that no one has ever seen,
+  # first time found 33 pre-existing failures that no one had ever seen,
   # because cargo used to stop at the first failing binary and nine of the ten
-  # never ran at all. They are not one bug: they span at least six unrelated
-  # root causes, and two of those groups (the crush registry scan returning
-  # nothing, cc-mirror attributing the wrong provider) look like real Windows
-  # defects rather than test artefacts. Blocking every PR on them would mean
-  # either reverting this job or landing a 30-plus-fix change nobody can
-  # review, so the signal goes in first and the failures get fixed against a
-  # visible baseline. See #1014 for the enumerated groups.
+  # never ran at all. They spanned six unrelated root causes, enumerated in
+  # #1014, and all six have since been addressed: the leg is green on main,
+  # 2,679 passed / 0 failed across all 11 binaries.
   #
+  # Two of those groups were first read as real Windows product defects: the
+  # crush registry scan returning nothing, and cc-mirror attributing the wrong
+  # provider. Both turned out to be fixture artefacts. The tests built their
+  # JSON by interpolating a path into a string literal, and on Windows the
+  # backslashes land unescaped, so `serde_json` rejected the fixture and the
+  # parse error was swallowed into an empty result. Crush discovery was
+  # additionally verified working against a real Windows install. See #1014.
+  #
+  # The leg stays advisory until the two un-redirected config roots are swept.
   # This must not become permanent: an advisory job nobody reads is how #997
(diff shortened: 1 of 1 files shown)


junhoyeo:
@NickAme03 you are right, and main is currently wrong in exactly the way you describe. I traced it before acting.

#1046's `test(core): escape paths written into JSON fixtures` says it outright:

> Nine fixtures assembled a config file with `format!` and dropped a raw filesystem path inside a JSON string: the Crush `projects.json` registries in `scanner.rs` and every cc-mirror `variant.json` […] **The real writers escape correctly** (Crush is Go `encoding/json`, cc-mirror is Node `JSON.stringify`), so **invalid JSON was only ever produced here.**

And the three actual product bugs in that PR are elsewhere in it:

| commit | kind |
|---|---|
| `test(core): escape paths written into JSON fixtures` | crush + cc-mirror — **fixtures** |
| `fix(timezone): stop TZ from making a Windows device unpinnable` | product |
| `fix(pricing): stop the bare `default` token from fuzzy-matching a priced key` | product |
| `fix(cache): key a Windows cache entry by the file, not by the separator used` | product |

What made this easy to get backwards is #1046's own subject line — "the three product bugs it was hiding" — which invites reading the headline failures as the product bugs. They are a different three.

## Why this cannot merge as written

You opened this against the tree as it stood on 08-07. #1068 landed a few hours later and did two things this PR assumes have not happened: it removed `continue-on-error`, and it rewrote the same comment block. So the diff conflicts, and its closing line —

> The leg stays advisory until the two un-redirected config roots are swept.

— is now false. The leg is a hard gate. Merging this would reintroduce a comment describing a state the file no longer has, which is the failure mode the PR exists to fix.

## What I did

Opened #1073 with your correction, rebased onto the post-#1068 comment and reframed around the hard gate, with you as `Co-authored-by`. It also names the four genuine Windows defects explicitly (adding #1007's lock/PID probes to #1046's three) so the list cannot be misread the same way again, and notes that the `default` defect is #1062's class — a contentless token electing a priced row — since that shape has now cost this repo twice.

Closing this in favor of #1073. The diagnosis was yours; only the framing needed to move. Please do flag the next one that reads stale — you were right that this comment is what someone consults before touching the leg.

PR #1070, opened 2026-08-07
fix(pricing): refuse to price router labels, prefer original vendor on ties

junhoyeo:
Both halves of #1062. A design pass proposing provider-prefix gating was rejected by three independent adversarial reviews first — this is what survived.

## The issue's own reproduction was wrong

I filed #1062 citing `cursor/auto`. **No parser emits that string.** `sessions/cursor.rs:356` asserts a bare `auto` with the vendor in `provider_id`; four more default to the same bare label. The prefix came from the CLI warning's `client/model` display.

That error mattered enormously: every prefix-based gate keys off a vendor prefix, and bare ids have none — so on 97 real ids the proposed design excluded **zero**, including `auto` itself. It "passed" its blast-radius check by never firing.

## Router labels

```
before:  tokscale pricing auto  ->  morph/auto   $0.85 / $1.55
after:   tokscale pricing auto  ->  Model not found
         tokscale pricing morph/auto  ->  morph/auto   (unchanged)
```

Nothing in the session log records which model served an `auto` request, so any rate attached describes a different model. And this isn't saved by the missing-cache-bucket accident I originally described — `covers_usage` demands rates only for *populated* buckets, and Copilot Desktop's sessions table has no cache-write column at all, so that shape submits at Morph's real rates. 519 of 5037 rows in one live Cursor cache carry it.

Bare ids only. `custom-pricing.json` is consulted **before** this guard, so a user who knows their router's effective rate can still state it — and #1069 makes stating a $0 rate possible.

## Tie-break

`prefers_model_part_key` broke ties by shortest key with a hardcoded `anthropic/` preference in front. For model part `auto`, models.dev carries `morph/auto` ($0.85/$1.55), `llmgateway/auto` ($0.00), `kilo/openrouter/auto` ($0.00), `orcarouter/orcarouter/auto` ($0.00) — the 10-character paid row won **because it was short**.

The anthropic case is generalised to `is_original_provider`, which is the rule it was always an instance of: when several vendors publish the same model name, the vendor who made it is the one whose rates describe it.

## Rejected, with evidence

| Direction | Why not |
|---|---|
| Gate on provider-prefix agreement | Excludes 218 of 1,969 real ids whose key names the *same* model — `anthropic/claude-3-opus` → `llmgateway/claude-3-opus` is Anthropic's exact list price |
| Gate on `provider_id` | Disagrees on ~79% of priced usage; it's circular (24 parsers derive it from the model string) and degrades to a *client* label precisely when the model is unknown |
| Reject all fuzzy matches | 38% of ids resolve fuzzy-only onto keys naming the same vendor and model |
| Any of the above | Regresses free models, reverting 55635ef4 — `stepfun/step-3.7-flash:free` plus 6 more verified |

## Tests

1,492 passed / 0 failed; fmt and `clippy --all-targets -- -D warnings` clean. Two new tests: bare labels refuse while qualified ones resolve, and the tie-break prefers an original provider over a shorter reseller key.

## Not covered

The label list is evidence-based, not exhaustive — it's what five parsers demonstrably emit. Other bare product labels may be in circulation.

---
## Summary by cubic
Stops pricing router labels like `auto` (including normalized forms) and prefers the original provider when multiple vendors share a model name. Also lets `custom-pricing.json` declare free models with 0.0 rates.

- **Bug Fixes**
  - Router labels: block bare and normalized forms (`auto`, `agent_review`, `auto(high)`, unknown-prefix `cx/auto`) from pricing; real qualified ids (e.g. `morph/auto`) still price. `custom-pricing.json` can still set a rate.
  - Tie-breaks: prefer the original provider over the shortest key when keys share a model part.
  - Custom pricing: accept 0.0 input/output rates (free models). Rows with no rates and any negative values still reject.

<sup>Written for commit ea1a743f722c0e27358ccddbf0ce5ffa2ad5174a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1070?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/pricing/custom.rs
+++ b/crates/tokscale-core/src/pricing/custom.rs
@@ -78,12 +78,20 @@ impl CustomModelPricing {
             "output_cost_per_token",
         )?;
 
-        if !input_cost_per_token.is_some_and(|value| value > 0.0)
-            && !output_cost_per_token.is_some_and(|value| value > 0.0)
-        {
-            return Err(
-                "at least one of input or output pricing must be present and positive".into(),
-            );
+        // @keep: requiring a POSITIVE rate here locked users out of the one
+        // escape hatch they have for free models.
+        //
+        // A rate of 0.0 is a statement — "this model is free" — and it is
+        // exactly what a user needs to express for a free tier that no upstream
+        // dataset publishes. Demanding a positive rate made that unsayable, so
+        // free usage stayed unpriced and was excluded from submission with no
+        // way to correct it (#1021).
+        //
+        // Absence is still rejected: a row that names no rate at all says
+        // nothing, and silently reading that as free would invent a $0 total
+        // for genuinely unknown pricing.
+        if input_cost_per_token.is_none() && output_cost_per_token.is_none() {
+            return Err("at least one of input or output pricing must be present".into());
         }
 
         Ok(ModelPricing {
@@ -657,7 +665,7 @@ mod tests {
     }
 
     #[test]
-    fn drops_entries_with_zero_prices() {
+    fn keeps_free_models_but_still_drops_negative_prices() {
         let temp = TempDir::new().unwrap();
         let path = temp.path().join("custom-pricing.json");
         fs::write(
@@ -683,11 +691,18 @@ mod tests {
 
         let loaded = CustomPricing::load_from_path(&path);
 
-        assert!(loaded.lookup("all-zero").is_none());
+        // A free model is the case this file exists to let users express when
+        // no upstream dataset publishes the model (#1021). 0.0 is an assertion
+        // ("free"), not an absence, so an all-zero row is kept.
+        let all_zero = loaded.lookup("all-zero").expect("free model must load");
+        assert_eq!(all_zero.input_cost_per_token, Some(0.0));
+        assert_eq!(all_zero.output_cost_per_token, Some(0.0));
+
         assert_eq!(
             loaded.lookup("free-input").unwrap().output_cost_per_token,
             Some(0.000008)
         );
+        // Unchanged: a negative rate is nonsense, not a statement about price.
         assert!(loaded.lookup("negative-output").is_none());
     }
 
--- a/crates/tokscale-core/src/pricing/lookup.rs
+++ b/crates/tokscale-core/src/pricing/lookup.rs
@@ -365,6 +365,13 @@ impl PricingLookup {
         // `try_strip_unknown_suffix` below.
         let normalized_owned = strip_parenthesized_reasoning_tier(&lower).map(str::to_owned);
 
+        // A tier suffix does not turn a router into a model: `auto(high)`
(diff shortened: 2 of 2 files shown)


junhoyeo:
Fixed in `ea1a743f`. Both bots caught the same hole and it was real — verified before fixing:

```
auto(high)  ->  morph/auto   $0.85/$1.55    (tier suffix stripped to `auto`)
cx/auto     ->  morph/auto                  (unknown vendor prefix dropped)
```

The guard ran on the **raw** model id, but the resolver normalizes afterwards, so a router label arriving in any transformed shape still reached the model-part fallback.

Guards now sit at both normalization points. Placement is the interesting part: the prefix-strip check goes **inside** the fallback rather than at the entry, because reaching that branch already means no dataset key matched the qualified id. A real `morph/auto` resolves long before it — which is exactly what keeps the bare-label rule from over-reaching:

```
auto          -> not found
auto(high)    -> not found
cx/auto       -> not found
morph/auto    -> morph/auto     ← unchanged
claude-opus-4 -> anthropic/claude-opus-4
```

1,492 tests pass, fmt and clippy clean. The regression test now covers both transformed shapes alongside the bare ones.

The custom-pricing documentation point is addressed on #1069 (`ca39b518`) — all four READMEs said a rate "must be present and positive", which would have kept users away from the escape hatch that change adds.

**Note on ordering:** this branch was cut from `fix/custom-pricing-zero-cost` rather than `main`, so it carries that commit too. Merge #1069 first and this applies cleanly on top.

junhoyeo:
Correcting my previous comment, which overstated the transformed-shape cases.

**Valid:** the mechanism. `auto(high)` and `cx/auto` genuinely resolved to `morph/auto` at $0.85/$1.55 before the fix — reproducible, and the guard closes it.

**Overstated:** that these occur in practice. I presented them next to the bare-`auto` case as if both were equally real. They are not. Measured against a live corpus:

| shape | occurrences |
|---|---|
| bare `auto` | **519** of 5037 rows |
| `auto(tier)` | **0** |
| `cx/auto` or any prefixed form | **0** |

`cx/` is a real omniroute prefix (`lookup.rs:418`) and `(level)` a real CLIProxyAPI suffix, but combining either with a *router label* is something I assumed rather than observed — it needs a proxy fronting a router.

So the transformed-shape guards are **defensive**, not corrective: a proven mechanism on an unobserved input. They stay because they are ~10 lines, cost nothing (`morph/auto` still resolves, 1,492 tests green), and close the class rather than the instance. But the load-bearing fix in this PR is the bare label, and that one is measured.

Issue #1071, opened 2026-08-08
fix: bare routing labels (auto) resolve to another provider's model-part price instead of being excluded


thedavidweng:
## Summary

Bare routing labels — most importantly `auto`, the unknown-model label Kiro emits (`UNKNOWN_MODEL` in `sessions/kiro.rs`) and the default-model label Cursor/Copilot record in usage rows — resolve through the models.dev model-part index to a real provider's priced `*/auto` key. They are not models and must never carry a price, but today they do.

## Root cause

The two model-part index fallbacks (`exact_match_models_dev` and `exact_match_openrouter_model_part` in `crates/tokscale-core/src/pricing/lookup.rs`) match "some other provider's model whose model-part equals this id" — the same cross-provider trust class as fuzzy matching. Fuzzy matching is gated by `is_fuzzy_eligible`, which consults `FUZZY_BLOCKLIST`, and `auto` is on that list. But the model-part index is consulted earlier in `lookup_auto` and was never gated, so a bare `auto` bypassed the blocklist entirely.

In the current models.dev dataset four providers carry the model part `auto`: `morph/auto` ($0.85/$1.55), `kilo/openrouter/auto` ($0.00), `llmgateway/auto` ($0.00), `orcarouter/orcarouter/auto` ($0.00). `prefers_model_part_key` breaks the tie by shortest key, so `morph/auto` wins and the label bills at morph's rates. One live cursor usage export carried `auto` on 519 of 5037 rows (from #1062).

## Fix

Gate both model-part index fallbacks on `FUZZY_BLOCKLIST` membership. A bare generic token then never resolves through another provider's model part; the id's own full dataset key (e.g. `morph/auto` spelled out) is still an exact match. In `exclude_unpriced_submission_messages`, `auto` is now classified as a generic routing label (the `gemini-default` precedent, generalized to a shared `ROUTING_LABEL_UNPRICED_REASON`) and excluded honestly instead of submitted at morph's rates.

PR: #1072. Relates to #1062 and the #1021/#1035 zero-rate resolution issues.

PR #1072, opened 2026-08-08
fix(pricing): keep bare routing labels off model-part index fallbacks

thedavidweng:
Defense-in-depth for #1062, rebased onto current main. The companion issue is #1071.

## What

The models.dev model-part index matched bare ids against "some other provider's model whose model-part equals this id" — the same cross-provider trust class as fuzzy matching — but only the fuzzy path consulted `FUZZY_BLOCKLIST`. A bare `auto` (the unknown-model label Kiro emits, and the default-model label Cursor/Copilot record) therefore elected `morph/auto` at $0.85/$1.55, because the model-part tie-break rejected it as spam. 519 of 5037 rows in one live cursor cache carried it.

#1070 already fixed the live case at the resolver top (`is_routing_label` refuses `auto`/`agent_review` before the model-part index is consulted, and the tie-break prefers original vendors). This PR lands the second layer inside the model-part helpers themselves and keeps submission reporting honest.

## Changes

- `crates/tokscale-core/src/pricing/lookup.rs`: both model-part index fallbacks (`exact_match_models_dev`, `exact_match_openrouter_model_part`) skip `FUZZY_BLOCKLIST` tokens, as defense-in-depth; full dataset keys still resolve. A future provider publishing `someone/default` as a model part is caught here even though no parser emits a bare `default` today, and any path reaching the index without passing the resolver-top guard is covered.
- `crates/tokscale-core/src/lib.rs`: `GEMINI_DEFAULT_UNPRICED_REASON` generalized to `ROUTING_LABEL_UNPRICED_REASON` via `is_generic_routing_label`, aligning with `lookup.rs`'s `ROUTING_LABELS` (`auto`, `agent_review`) plus the historical `gemini-default` pair — so the submitted exclusion reason tells users "this label never named a model" instead of "your model is missing from the dataset".
- `crates/tokscale-core/src/pricing/litellm.rs`: the zero-row doc comment no longer calls kenari's nemotron row a "genuinely free model" — kenari's entire catalog is subscription-priced at zero (all 38 models.dev rows), and the same nemotron has 14 paid siblings.

## Verification

- New tests: `model_part_index_does_not_resolve_bare_generic_tokens` (red-green verified: fails without the gate, passes with it) and `submission_excludes_unpriced_auto_routing_label` (fixture quotes a cache-read rate so the pre-fix fallback would submit, making the test actually detect the regression).
- `cargo test -p tokscale-core --lib`: 1496 passed, 1 ignored; `cargo fmt --check` clean; clippy clean.
- Live CLI: `pricing auto` and `pricing agent_review` report "Model not found"; `morph/auto` full key and real models (`nvidia/nemotron-3-ultra-550b-a55b`, `openai/codex-auto-review`) resolve exactly as before.

The `github-copilot/` prefix gap, the `nvidia/` zero-rate tie-break election, and `codex-auto-review` attribution remain separate open issues (#1021, #1035) with different root causes, untouched by this PR.

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -3027,10 +3027,22 @@ fn build_graph_from_messages(
     Ok(result)
 }
 
-const GEMINI_DEFAULT_UNPRICED_REASON: &str =
+const ROUTING_LABEL_UNPRICED_REASON: &str =
     "generic routing label has no authoritative model-to-price mapping";
 const MISSING_MODEL_PRICING_REASON: &str = "no authoritative model-to-price mapping";
 const INCOMPLETE_MODEL_PRICING_REASON: &str = "pricing does not cover every populated token bucket";
+
+/// Routing labels name the router that served the request, never the model
+/// that answered it, so they have no authoritative model-to-price mapping.
+/// They are the same labels `lookup::is_routing_label` (lookup.rs) refuses at
+/// the resolver top — `auto` (the unknown-model label Kiro emits and the
+/// default-model label Cursor/Copilot record) and Cursor's `agent_review` —
+/// plus the historical `gemini-default` pair.
+fn is_generic_routing_label(provider_id: &str, model_id: &str) -> bool {
+    (provider_id.eq_ignore_ascii_case("google") && model_id.eq_ignore_ascii_case("gemini-default"))
+        || model_id.eq_ignore_ascii_case("auto")
+        || model_id.eq_ignore_ascii_case("agent_review")
+}
 
 fn has_positive_token_usage(tokens: &TokenBreakdown) -> bool {
     tokens.input > 0
@@ -3062,10 +3074,8 @@ fn exclude_unpriced_submission_messages(
             );
 
         if is_unpriced {
-            let reason = if message.provider_id.eq_ignore_ascii_case("google")
-                && message.model_id.eq_ignore_ascii_case("gemini-default")
-            {
-                GEMINI_DEFAULT_UNPRICED_REASON
+            let reason = if is_generic_routing_label(&message.provider_id, &message.model_id) {
+                ROUTING_LABEL_UNPRICED_REASON
             } else if pricing
                 .lookup_with_source_and_provider(
                     &message.model_id,
@@ -4388,8 +4398,8 @@ mod tests {
         parse_local_clients, parsed_to_unified, paths, pricing, retain_for_requested_clients,
         scanner, select_local_parse_pricing, unified_to_parsed, validate_priced_messages, ClientId,
         GraphPricingRequirement, GroupBy, LocalParseOptions, ReportOptions, TokenBreakdown,
-        UnifiedMessage, UnpricedSubmissionExclusion, GEMINI_DEFAULT_UNPRICED_REASON,
-        INCOMPLETE_MODEL_PRICING_REASON, MISSING_MODEL_PRICING_REASON, UNKNOWN_WORKSPACE_LABEL,
+        UnifiedMessage, UnpricedSubmissionExclusion, INCOMPLETE_MODEL_PRICING_REASON,
+        MISSING_MODEL_PRICING_REASON, ROUTING_LABEL_UNPRICED_REASON, UNKNOWN_WORKSPACE_LABEL,
     };
     use serial_test::serial;
     use std::collections::{HashMap, HashSet};
@@ -8383,7 +8393,73 @@ mod tests {
                 model_id: "gemini-default".to_string(),
                 message_count: 7,
                 total_tokens: 18,
-                reason: GEMINI_DEFAULT_UNPRICED_REASON,
+                reason: ROUTING_LABEL_UNPRICED_REASON,
+            }
+        );
+    }
+
+    #[test]
+    fn submission_excludes_unpriced_auto_routing_label() {
+        // `auto` is the unknown-model label Kiro emits and the default-model
+        // label Cursor/Copilot record in usage rows. A models.dev `morph/auto`
+        // paid row exists, so before the resolver refused routing labels the
+        // bare label resolved to it and slipped through submission at morph's
+        // rates (#1062). The fixture quotes a cache-read rate precisely so the
+        // pre-fix fallback covers all three populated buckets (7 input, 11
+        // cache-read, 0 output) and submits the row — without it, the row
+        // would fail coverage on the missing cache rate and the test would
+        // pass even with the bug. The label must instead be excluded with the
+        // routing-label reason.
+        let mut models_dev = HashMap::new();
+        models_dev.insert(
+            "morph/auto".to_string(),
+            pricing::ModelPricing {
+                input_cost_per_token: Some(8.5e-7),
+                output_cost_per_token: Some(1.55e-6),
+                cache_read_input_token_cost: Some(1.6e-7),
+                ..Default::default()
+            },
+        );
+        let pricing = pricing::PricingService::new_with_custom_and_models_dev(
+            pricing::custom::CustomPricing::default(),
+            HashMap::new(),
+            HashMap::new(),
+            models_dev,
+        );
+        let mut auto = UnifiedMessage::new(
+            "kiro",
+            "auto",
+            "amazon-bedrock",
+            "generic",
+            1_736_510_400_000,
+            TokenBreakdown {
+                input: 7,
+                cache_read: 11,
+                ..Default::default()
+            },
+            0.0,
+        );
+        auto.message_count = 7;
+
+        let graph = build_graph_from_messages(
+            vec![auto],
+            Some(&pricing),
+            GraphPricingRequirement::Submission,
+            std::time::Instant::now(),
+            &crate::bucket_tz::BucketTimezone::Local,
+        )
+        .expect("routing label must not abort submission");
+
+        assert_eq!(graph.summary.total_tokens, 0);
+        assert_eq!(graph.unpriced_submission_exclusions.len(), 1);
+        assert_eq!(
+            graph.unpriced_submission_exclusions[0],
+            UnpricedSubmissionExclusion {
+                provider_id: "amazon-bedrock".to_string(),
+                model_id: "auto".to_string(),
+                message_count: 7,
+                total_tokens: 18,
+                reason: ROUTING_LABEL_UNPRICED_REASON,
             }
         );
     }
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -3074,9 +3074,18 @@ fn exclude_unpriced_submission_messages(
             );
 
         if is_unpriced {
-            let reason = if is_generic_routing_label(&message.provider_id, &message.model_id) {
-                ROUTING_LABEL_UNPRICED_REASON
-            } else if pricing
+            // Resolution is consulted before the routing-label reason, not
+            // after. `custom-pricing.json` is read first by
+            // `lookup_with_source_and_provider`, and stating a rate for `auto`
+            // there is the user asserting the label does name something for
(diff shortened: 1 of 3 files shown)


junhoyeo:
Thanks @thedavidweng — I verified all three parts against the live datasets. Two land, one is already covered by #1070, which merged seven minutes after you opened this.

## The comment fix is right, and more right than it claims

I checked `kenari` against live models.dev, and the correction understates the case:

```
kenari models: 38, all-zero rows: 38
```

**Every one of kenari's 38 models is `{input: 0, output: 0}`** — `grok-4-5`, `claude-fable-5`, `gpt-image-2`, all of them. kenari is a subscription catalog end to end, so `kenari/nemotron-3-ultra-550b-a55b` was never "a genuinely free model"; it is the *same* case as `kenari/claude-opus-4-7`. The comment was contrasting two rows drawn from one world and presenting them as two.

Your sibling numbers check out exactly:

```
nvidia/nvidia/nemotron-3-ultra-550b-a55b      {'input': 0.5,  'output': 2.5,  'cache_read': 0.15}
llmgateway/nemotron-3-ultra-550b              {'input': 0.5,  'output': 2.5,  'cache_read': 0.15}
openrouter/nvidia/nemotron-3-ultra-550b-a55b  {'input': 0.6,  'output': 3.6,  'cache_read': 0.2}
togetherai/nvidia/nemotron-3-ultra-550b-a55b  {'input': 0.6,  'output': 3.6,  'cache_read': 0.2}
kilo/nvidia/nemotron-3-ultra-550b-a55b        {'input': 0.5,  'output': 2.2,  'cache_read': 0.1}
wandb/nvidia/NVIDIA-Nemotron-3-Ultra-550B...  {'input': 0.75, 'output': 2.75, 'cache_read': 0.15}
… 14 paid rows total, against 5 zero rows
```

**One nit on the replacement example.** You swapped in `openrouter/nvidia/nemotron-3-ultra-550b-a55b:free`, which is real and is `{0, 0}` — but it slightly undercuts the sentence it sits in. The paragraph's claim is that the two cases are *indistinguishable in the data*, and `:free` is a distinguishing marker sitting right there in the key. It is only a convention rather than a field, so the claim survives, but `opencode/nemotron-3-ultra-free` or a bare unmarked zero row would carry the point without the reader spotting the suffix and wondering.

## The exclusion reason is a real gain

I probed current main directly:

```
PROBE exclusion = UnpricedSubmissionExclusion {
    provider_id: "amazon-bedrock", model_id: "auto", message_count: 7,
    total_tokens: 18, reason: "no authoritative model-to-price mapping"
}
```

So post-#1070 `auto` is correctly *excluded*, but reported with the generic reason. Your `is_generic_routing_label` gives it `"generic routing label has no authoritative model-to-price mapping"`, which is the difference between a user thinking their model is missing from the dataset and understanding that the label never named a model. Worth having.

## The lookup.rs gate is already covered

#1070 landed an `is_routing_label` guard at the top of the resolver, before the model-part index is ever consulted. I ran **your** test verbatim against current main with your `lookup.rs` hunks omitted:

```
test pricing::lookup::tests::probe_1072_model_part_bare_routing_label ... ok
```

It passes without your change. And I checked whether the wider `FUZZY_BLOCKLIST` gate buys anything beyond `auto`, across all 6,231 models.dev keys and the OpenRouter set:

```
models.dev keys whose model-part == a FUZZY_BLOCKLIST token:
  auto   total=8  paid=1  ['morph/auto']
openrouter: (none)
```

`auto` is the only token in the list that collides at all. `mini`, `chat`, `base`, `claude`, `anthropic`, `gemini`, `model`, `router`, `default` have zero model-part collisions in either dataset.

That does not make the hunks wrong — a second layer inside the model-part helpers is defensible for a bug class that silently mispriced 519 rows, and a new provider publishing `someone/default` tomorrow would be caught by yours and not by #1070's. But it is defense against an unobserved shape, so it should land described that way rather than as the fix for #1062, which is already fixed.

## To land

- Rebase onto main (you branched from #1068, so #1069 and #1070 are missing).
- `cargo fmt` — Lint is red purely on formatting in the new test.
- Either drop the `lookup.rs` hunks or keep them with the comment reframed as defense-in-depth, noting that #1070's guard is what closes #1062.

The comment correction and the exclusion reason I want either way.

thedavidweng:
Thanks for the thorough verification — all three points addressed.

- **Rebased** onto current main (includes #1069 and #1070).
- **cargo fmt** — Lint should be green now.
- **lookup.rs kept, reframed as defense-in-depth.** The comments now credit #1070's resolver-top `is_routing_label` guard as what closes the bare-`auto` case, and describe the blocklist gates as the second layer: they enumerate nothing, so they catch `FUZZY_BLOCKLIST` tokens no parser emits today (`default`, `router`, `mini`, ...) if a future provider publishes one as a model part, and they protect any path that reaches the model-part index without passing the resolver-top guard. I also retargeted the lookup test to prove exactly that layer: it uses `someprovider/router` and `someprovider/default` fixtures (tokens #1070's list does not enumerate) and I verified red-green that it fails when the gate is removed and passes with it.

One alignment worth noting: `is_generic_routing_label` in lib.rs now mirrors lookup.rs's ROUTING_LABELS (`auto`, `agent_review`) on top of `gemini-default`, so the submitted exclusion reason and the resolver's refusal talk about the same labels.

On the litellm example — swapped to `opencode/nemotron-3-ultra-free` per your nit, and the paragraph now states the full strength of the correction: all 38 kenari rows are `{0, 0}` (subscription catalog end to end), with 14 paid siblings on the same nemotron.

thedavidweng:
Valid — fixed. The fixture now quotes `cache_read_input_token_cost` on the `morph/auto` row (alongside input and output), so the pre-fix model-part fallback covers all three populated buckets (7 input / 11 cache-read / 0 output) and would submit the row at morph's rates. The test therefore fails against the pre-fix resolver and passes now; I confirmed red-green locally before pushing.

junhoyeo:
Reviewed as part of a sweep across the open PRs. I specifically checked the regression this repo has hit before — a pricing PR opened against an older tree quietly reverting a `lookup.rs` guard — and this branch is clean on that count: its merge base is current `main`, and #1070's `is_routing_label` resolver-top guard and original-vendor tie-break both survive the diff intact.

I also enumerated what the new gate actually excludes. Against the cached LiteLLM dataset, three real keys have a model-part that is a `FUZZY_BLOCKLIST` token:

- `openrouter/openrouter/auto`
- `openrouter/switchpoint/router`
- `deepgram/base`

All three still resolve by full key, so the gate only affects *bare* queries — which is the intent. Worth naming `deepgram/base` in a comment though: a bare `base` is now permanently unpriceable, and that is a deliberate trade rather than an oversight, so it should read that way to the next person.

Four things worth fixing before merge:

- **Reason-string ordering.** The routing-label reason is checked before the incomplete-pricing reason, so a user who *has* priced `auto` in `custom-pricing.json` is still told no mapping exists. Custom pricing should win — it is an explicit user statement that the label does name something for them.
- **`is_generic_routing_label` hand-copies `ROUTING_LABELS` from `lookup.rs` and drops the `.trim()`.** Two copies of the same list in two files will drift, and the missing trim means ` auto ` is treated differently by the two paths. Sharing one constant would remove both problems.
- **The provider-hinted key-part matcher is not gated.** The two model-part index fallbacks now consult the blocklist, but a blocklisted bare token can still resolve cross-model when a provider hint is present — so the defense-in-depth has a seam. Worth either closing it or saying explicitly why the hint makes it safe.
- **`FUZZY_BLOCKLIST`'s doc comment still describes it as fuzzy-only** after this PR gives it authority over exact model-part resolution. That comment is exactly what someone will read before adding an entry, and adding an entry is now a heavier decision than it was.

Minor: the substituted example in `litellm.rs` contradicts the "null cache fields, byte for byte" claim it is meant to illustrate.

The `GEMINI_DEFAULT_UNPRICED_REASON` → `ROUTING_LABEL_UNPRICED_REASON` rename is safe as far as I can tell — I grepped `frontend/` and `packages/` for the old constant's *value* rather than its Rust name and found no consumer matching on it.

junhoyeo:
Thanks @thedavidweng — I pushed `5e18b0e9` to this branch fixing the reason-string ordering item, since it was the one with direct user-facing effect. The other three items from my review I deliberately left for you.

**The bug reproduced exactly.** `PricingService::lookup_with_source_and_provider` (`pricing/mod.rs:366`) consults `lookup_custom` *before* delegating to the inner resolver, so a user-priced `auto` does resolve as `Custom`. But the exclusion site tested `is_generic_routing_label` first and never got that far. `ROUTING_LABELS`' own doc comment already promises the opposite — "`custom-pricing.json` is consulted before this, so a user who knows their router's effective rate can still state it" — so the reason string was contradicting the documented escape hatch.

Failing output before the fix:

```
left:  reason: "generic routing label has no authoritative model-to-price mapping"
right: reason: "pricing does not cover every populated token bucket"
```

The fix swaps the first two branches of the reason chain in `exclude_unpriced_submission_messages` so the resolved-pricing branch is tested first; `MISSING_MODEL_PRICING_REASON` remains the fallback. Nothing regresses for genuinely unpriced labels — with no custom entry the resolver refuses `auto`/`agent_review` outright and returns `None`, so the routing-label reason still fires. Your own `submission_excludes_unpriced_auto_routing_label` and the pre-existing `submission_excludes_unpriced_generic_gemini_default_but_keeps_priceable_usage` both still pass unchanged.

Added `custom_priced_routing_label_reports_incomplete_pricing_not_missing_mapping` as the regression guard, verified red-green.

`cargo test -p tokscale-core` — 1497 passed, 0 failed across every target. `cargo fmt --all --check` clean, `cargo clippy --all-targets` no warnings. CI is green per-job: Lint, Tests (ubuntu-latest), Tests (windows-latest), and both Windows native builds.

**Still yours to decide** (I did not touch them): the duplicated `ROUTING_LABELS` list in `lib.rs` plus the missing `.trim()` — unlike `is_routing_label`, `is_generic_routing_label` does not trim, so a whitespace-padded id diverges between the two paths; the ungated provider-hinted key-part matcher; and the stale `FUZZY_BLOCKLIST` doc comment.

junhoyeo:
Pushed one more small follow-up to this branch (`7b8fda7f`), on top of the exclusion-reason ordering fix.

`is_generic_routing_label` in `crates/tokscale-core/src/lib.rs` restated lookup.rs's `ROUTING_LABELS` entries inline (`auto`, `agent_review`). That left two copies of one list in two files, so a label added to `ROUTING_LABELS` would keep being refused by the resolver while silently losing the routing-label exclusion reason. The copy also skipped the `.trim()` that `is_routing_label` applies, so the two paths disagreed on a whitespace-padded id: the resolver refused to price ` auto `, but the reason path did not recognise it as a routing label and reported "no authoritative model-to-price mapping" instead.

The fix is the small one: `lookup::is_routing_label` becomes `pub(crate)` and `is_generic_routing_label` calls it rather than restating the list, with the provider-scoped `gemini-default` arm trimmed to match. `gemini-default` handling is unchanged in behaviour — it stays in the reason and is still not part of the resolver gate.

Added `whitespace_padded_routing_label_is_classified_the_same_by_resolver_and_reason`, which asserts the two predicates agree on ` auto ` and then walks a padded label through submission with a fully priced models.dev `morph/auto` row present. Verified red before the change (`left: true, right: false`) and green after. `custom_priced_routing_label_reports_incomplete_pricing_not_missing_mapping`, `submission_excludes_unpriced_auto_routing_label`, and `submission_excludes_unpriced_generic_gemini_default_but_keeps_priceable_usage` all still pass. `cargo fmt --all -- --check` and `cargo clippy -p tokscale-core --all-targets` are clean, and `cargo test -p tokscale-core` is 1498 passed / 0 failed on the lib target with every other test binary green.

I deliberately left two other discussed items alone — they are your call, not blockers I wanted to take unilaterally: the ungated provider-hinted key-part matcher, and the stale `FUZZY_BLOCKLIST` doc comment.

PR #1073, opened 2026-08-08
docs(ci): stop crediting the windows leg with two defects it did not find

junhoyeo:
Closes the factual half of #1066.

## What is wrong today

The comment above the `tests` job — which I rewrote in #1068 while flipping the gate — says:

> six unrelated root causes, **including two real Windows product defects** (the crush registry scan returning nothing, cc-mirror attributing the wrong provider)

Both were fixture artefacts, not product defects. #1046's `test(core): escape paths written into JSON fixtures` is explicit:

> Nine fixtures assembled a config file with `format!` and dropped a raw filesystem path inside a JSON string: the Crush `projects.json` registries in `scanner.rs` and every cc-mirror `variant.json` […] **The real writers escape correctly** (Crush is Go `encoding/json`, cc-mirror is Node `JSON.stringify`), so **invalid JSON was only ever produced here.**

On Windows the temp dir is `C:\Users\RUNNER~1\AppData\Local\Temp\...`, so the fixture contains `\U`, `\A`, `\L`, `\T` — none of which are JSON escapes. Both readers treat unparseable config as absent rather than as an error, so the bug surfaced as eleven assertions about discovery finding nothing, several files from the cause.

## Why it was easy to get backwards

#1046's subject line says "the three product bugs it was hiding", which invites reading the headline failures as the product bugs. They are not the same three. Checking the sub-commits:

| | |
|---|---|
| `test(core): escape paths written into JSON fixtures` | crush + cc-mirror — **fixtures** |
| `fix(timezone): stop TZ from making a Windows device unpinnable` | product |
| `fix(pricing): stop the bare \`default\` token from fuzzy-matching a priced key` | product |
| `fix(cache): key a Windows cache entry by the file, not by the separator used` | product |

Plus #1007 (`fix(windows): stop lock and PID probes silently failing off Unix`). Four product defects; crush and cc-mirror are in none of them.

## Why this is worth a commit

This comment is what a future reader consults before touching the leg. "Two real Windows product defects" is the sentence that stops someone from deleting a test whose fixture merely looks wrong — and it points at the wrong two.

The correction also notes that the `default` defect is #1062's class (a contentless token electing a priced row), since that is now the second time this repo has paid for that shape and this is the first place a reader is likely to meet it.

## Relationship to #1066

@NickAme03 diagnosed this correctly in #1066 and is credited as co-author. That PR cannot merge as written: it was opened against the pre-#1068 tree, so it keeps `continue-on-error` and states "the leg stays advisory until the two un-redirected config roots are swept" — which stopped being true when #1068 landed. The factual correction is theirs; only the framing changed.

No behavior change. One comment block.

---
## Summary by cubic
Updates the Windows CI leg comment in `.github/workflows/test_coverage.yml` to stop misattributing two fixture-caused failures as product bugs, explicitly list the real defects surfaced by the leg, and correct the test binary count to 10. Aligns attribution with #1007, #1012, #1046, and #1014; no behavior or gating changes.

<sup>Written for commit ac457d0cb51e6211ba79532f4e909203b87f3841. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1073?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -56,11 +56,26 @@ jobs:
   # *executed* a test there, so Windows-only defects reached users unseen.
   #
   # The leg was advisory (`continue-on-error`) while the 33 pre-existing
-  # failures that first run exposed were worked through: six unrelated root
-  # causes, including two real Windows product defects (the crush registry
-  # scan returning nothing, cc-mirror attributing the wrong provider). Those
-  # landed in #1007, #1012 and #1046, and the suite now completes green across
-  # all 11 binaries.
+  # failures that first run exposed were worked through, across six unrelated
+  # root causes (#1014). Those landed in #1007, #1012 and #1046, and the suite
+  # now completes green across all 11 binaries.
+  #
+  # Most of the 33 were the harness lying, not the product: fixtures built
+  # JSON with `format!` and dropped a raw path inside a string, so on Windows
+  # `C:\Users\RUNNER~1\...` put `\U` and `\A` in the document and serde_json
+  # rejected it. Both readers treat unparseable config as absent, so the Crush
+  # registry scan returned nothing and cc-mirror sessions fell back to plain
+  # Claude - failures that read like product defects several files from the
+  # cause. The real writers (Go `encoding/json`, Node `JSON.stringify`) escape
+  # correctly, so only the tests ever emitted invalid JSON.
+  #
+  # Four genuine Windows product defects did surface underneath: lock and PID
+  # probes silently failing off Unix (#1007), TZ leaving a device unpinnable,
+  # the bare `default` token fuzzy-matching a priced key, and a cache entry
+  # keyed by path separator rather than by file (all #1046). The `default` one
+  # is the same class of bug as #1062 - a generic token with no model identity
+  # electing some provider's priced row - found here only because the Windows
+  # leg ran the test that caught it.
   #
   # It is a HARD GATE as of this change. The advisory phase was explicitly
   # temporary - an advisory job nobody reads is how #997 happened in the first
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -58,7 +58,7 @@ jobs:
   # The leg was advisory (`continue-on-error`) while the 33 pre-existing
   # failures that first run exposed were worked through, across six unrelated
   # root causes (#1014). Those landed in #1007, #1012 and #1046, and the suite
-  # now completes green across all 11 binaries.
+  # now completes green across all 10 test binaries.
   #
   # Most of the 33 were the harness lying, not the product: fixtures built
   # JSON with `format!` and dropped a raw path inside a string, so on Windows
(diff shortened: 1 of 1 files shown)


junhoyeo:
Reviewed alongside the rest of the open PRs. The core correction is right: #1046's commit body is explicit that the crush and cc-mirror failures were fixture artefacts and that "the real writers escape correctly", so calling them Windows product defects was wrong and worth fixing.

cubic's thread is also right on the merits, and I would take it. "Four genuine Windows product defects" reproduces the same category error one level up — the bare `default` fuzzy-match is a bug on every platform, as the very next line concedes. Something like "Four genuine product defects surfaced by the Windows leg" fixes it without losing the point.

Three more factual snags, which matter more than usual in a PR whose whole purpose is factual accuracy:

- **"all 11 binaries"** — the workspace has 10 test binaries. This came in from #1068 but it is re-typed inside this PR's added lines, so it is fair game here.
- **"Four" counts grouped items, not defects.** #1007 alone fixed three, and the comment collapses them into one clause while dropping the third — artifact paths written with the native separator — entirely. That one is arguably the most user-visible of the three.
- **"found here only because the Windows leg ran the test that caught it"** is not right. The test is ungated and ran on the Ubuntu leg on every PR. What exposed the bug was the Windows sandbox escaping to the real `%APPDATA%` catalog — which is itself a harness defect, i.e. the same category this comment separates out two paragraphs earlier.

On the AGENTS.md "no hardcoded counts" rule: I do not think it binds here, since it is scoped to AGENTS.md files rather than Rust or YAML comments. But a literal count in a comment that exists to be read years later will go stale the moment a fifth defect lands, and the fix is free — phrasing it as a list without the leading tally reads the same and cannot rot.

No behavior change, YAML still valid, gate semantics untouched.

junhoyeo:
Pushed `ac457d0c` correcting the four items above. All four checked out against primary sources; none was a false alarm.

- **"Four genuine Windows product defects"** → reworded so the Windows leg gets credit for *surfacing* the defects rather than owning them. #1046's own body says the `default` mispricing hit "every platform", so calling it a Windows defect repeated the exact error this PR exists to fix.
- **"all 11 binaries"** → **10**. Measured two ways: `cargo metadata --no-deps` enumerates 10 test targets, and `cargo test --workspace --all-features --no-run --message-format=json` emits exactly 10 artifacts with `profile.test == true`. The 11th line cargo prints is `Doc-tests tokscale-core`, which is not a binary — almost certainly where the 11 came from. The file was already self-inconsistent: the `--no-fail-fast` comment further down says "twelve failures in tokscale-cli meant the other nine binaries produced no result", i.e. 1 + 9 = 10.
- **The tally undercounted #1007.** It fixed three defects, not one, and the dropped third — `resolve_cache_relative_artifact_path` rejecting the native separator — is the most user-visible of them. `to_relative_artifact_path` uses `Path::to_string_lossy`, so on Windows every manifest entry is stored as `sessions\<id>.jsonl`, and `cleanup_stale_session_artifacts` runs on the main sync path (`antigravity.rs:233`). Any Windows sync with an artifact to retire failed *after* the manifest had already been written. All six defects are now enumerated per PR.
- **"found here only because the Windows leg ran the test that caught it"** → the test (`test_submit_excluding_only_generic_gemini_usage_does_not_promise_submission`, `cli_tests.rs:3768`) is an ungated `#[test]` that ran on Ubuntu every PR. What exposed the bug was the Windows sandbox escaping to the runner's real `%APPDATA%` catalog — a harness defect making a product defect visible, which is the same category this comment separates out two paragraphs earlier. Attribution rewritten to say that.

Also dropped the leading hardcoded tally so the paragraph cannot go stale when a fifth defect lands.

Comment-only, +20/-8 in one file. YAML re-validated with `yaml.safe_load`; no keys, gates, or `continue-on-error` semantics touched. Lint, Actionlint, Tests (ubuntu-latest) and Tests (windows-latest) are all green at `ac457d0c`.

PR #1075, opened 2026-08-08
fix(core): join client scan roots with native path separators

### LINKED ISSUE #1048: clients --json emits mixed-separator paths on Windows (C:\Users\me/.codex/sessions)
## What

`ClientDef::resolve_path_with_env_strategy` builds every client's scan root by concatenating with a hardcoded forward slash:

```rust
// crates/tokscale-core/src/clients.rs:189
pub fn resolve_path_with_env_strategy(&self, home_dir: &str, use_env_roots: bool) -> String {
    format!(
        "{}/{}",
        self.root.resolve_with_env_strategy(home_dir, use_env_roots),
        self.relative_path
    )
}
```

On Windows the root half comes back with native separators (`C:\Users\me`) while the joining separator and the whole relative half are forward slashes (the client table stores `.codex/sessions` and friends as `/`-joined literals). The result is a mixed-separator path:

```
C:\Users\me/.codex/sessions
```

That string reaches user-facing JSON verbatim. `tokscale clients --json` serializes it as `sessionsPath` (`crates/tokscale-cli/src/main.rs:4029`), and `additionalPaths[].path` has the same shape:

```json
{ "client": "codex", "sessionsPath": "C:\\Users\\me/.codex/sessions", "sessionsPathExists": true }
```

## Why this is cosmetic, not broken

Windows accepts `/` in paths, so the value resolves: `Path::new(&sessions_path).exists()` on the next line works, and every consumer inside the binary works. Nothing is misreported — the path is correct, it is just spelled two ways in one string. Filing this as a spelling question, not a bug.

## Why it was not fixed in #1046

#1046 (the Windows CI leg) took `crates/tokscale-cli/tests/cli_tests.rs` from asserting a `Path::join`-built expectation to asserting what the binary emits, via a `client_scan_path` helper. Two assertions now pin the mixed spelling: the codex `sessionsPath` and the claude `additionalPaths[0].path`.

That was the right call for that PR, for a reason worth recording here so this issue is not read as a regression report:

- The mixed spelling predates #1046. Nothing in that PR changed the emitter.
- `Path::join` was **also** mixed, so no native-separator contract was surrendered by matching the emitter. `join` only fixes the *first* separator — the relative half keeps its forward slashes either way:
  - `join`:    `C:\Users\me\.codex/sessions`
  - emitter:   `C:\Users\me/.codex/sessions`
- `--home C:/Users/me` produces a mixed path no emitter change prevents, because the root is whatever the user passed.

So the test pins today's behaviour rather than endorsing it, and changing the spelling is a product decision about user-facing output, separate from getting the Windows leg green.

## The decision to make

Should `clients --json` normalize the paths it prints, and to what?

1. **Leave as-is.** Mixed separators; correct, ugly, and already what any existing Windows script parses.
2. **Native separators** (`C:\Users\me\.codex\sessions`). Matches what a Windows user sees everywhere else, including `dir` output and Explorer.
3. **All forward slashes** (`C:/Users/me/.codex/sessions`). Uniform across platforms, easiest for consumers to split, and already valid on Windows.

Options 2 and 3 both change output that scripts may parse today, so whichever is picked should land deliberately rather than as a drive-by.

## Scope if fixed

- `crates/tokscale-core/src/clients.rs:189` — `resolve_path_with_env_strategy`, the single source of the spelling.
- `crates/tokscale-cli/src/main.rs:4029` — the `clients --json` emitter (and the human-readable renderer at `:4207`, which formats the same value).
- `crates/tokscale-cli/tests/cli_tests.rs` — `client_scan_path` and its two call sites; these must change in the same commit, since they exist to pin the current spelling.

Other `resolve_path` consumers use the value as a path rather than printing it, so they are unaffected by a display-only change but would follow a normalization done in the emitter itself.

Felictycf:
## Summary

`ClientDef::resolve_path_with_env_strategy` built every client scan root by hand-concatenating `"{root}/{relative}"` with a hardcoded forward slash. On Windows the root half arrives with native separators (`C:\Users\me`), so the result was a **mixed-separator path** (`C:\Users\me/.codex/sessions`) that reached user-facing `clients --json` output verbatim (`sessionsPath` and `additionalPaths[].path`).

## Fix

Join through `std::path::Path` instead, so every platform gets native separators:

```rust
std::path::Path::new(&root).join(self.relative_path).to_string_lossy().into_owned()
```

Unix behavior is unchanged (`/tmp/home` + `.codex/sessions` still resolves the same way); Windows now emits `C:\Users\me\.codex\sessions`.

## Test

Adds `test_resolve_path_joins_with_native_separators_not_hardcoded_slash`, which asserts the joined result equals what `Path::join` produces on the host platform (and, under `#[cfg(windows)]`, that no forward slash remains).

Full `tokscale-core` lib suite: **1497 passed, 0 failed**.

Fixes #1048

---
## Summary by cubic
Normalize all client, scanner, and headless scan roots to use native path separators, eliminating mixed Windows paths in `clients` output and internal scans. Unix is unchanged; Windows now prints backslash-only paths and uses consistent cache keys (fixes #1048).

- **Bug Fixes**
  - Join by pushing each relative component onto a `PathBuf` across clients and scanner; no mixed separators and the root is returned unchanged when `relative_path` is empty.
  - Route `PathRoot` fallbacks and env/XDG defaults through the same native join; fix `clean_reasonix_env_dir` to expand `~` with native separators.
  - Normalize headless roots; make the `tokscale clients` transcripts text assertion platform-aware.
  - Preserve an absolute `CODEBUFF_DATA_DIR` (including `/`) and accept trailing `/`; no trimming that turns `/` into a relative path; add tests to pin this.
  - Update tests/helpers to expect native separators (`native_join`, CLI helper, Kimchi/Senpi env-override expectations); fix `lib.rs` fixture comments.

<sup>Written for commit 2a8844578425050ed034f54b0d502177c7b189ff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1075?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -3214,13 +3214,18 @@ fn test_clients_command_includes_claude_transcripts_text() {
     let tmp = create_empty_fixture_dir();
     fs::create_dir_all(tmp.path().join(".claude/transcripts")).unwrap();
 
+    // The transcripts path is spelled with native separators (#1048): on
+    // Windows the `~` root renders with backslashes throughout.
+    #[cfg(windows)]
+    let expected = "additional: ~\\.claude\\transcripts ✓";
+    #[cfg(not(windows))]
+    let expected = "additional: ~/.claude/transcripts ✓";
+
     cmd_with_home(tmp.path())
         .arg("clients")
         .assert()
         .success()
-        .stdout(predicate::str::contains(
-            "additional: ~/.claude/transcripts ✓",
-        ));
+        .stdout(predicate::str::contains(expected));
 }
 
 #[test]
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -785,27 +785,20 @@ fn settings_json_path(base: &Path) -> std::path::PathBuf {
 
 /// A scan path spelled the way the binary spells it.
 ///
-/// `ClientDef::resolve_path` builds every client's root as
-/// `format!("{root}/{relative}")` — one separator, `/`, on every platform,
-/// with the relative half a `/`-joined literal in the client table. Windows
-/// accepts `/` in a path, so the value works; it just is not what
-/// `Path::join` produces. An expectation built with `join` disagreed on the
-/// first separator only (`...\.tmpXXXX\.codex/sessions` against the emitted
-/// `...\.tmpXXXX/.codex/sessions`) — and note that `join` did not give a
-/// natively-spelled path either, because the rest of the literal keeps its
-/// forward slashes. There was no native-separator contract to preserve, only
-/// two spellings of the same path, so match the one the binary actually
-/// emits.
-///
-/// That does make this helper a pin on the mixed spelling: `clients --json`
-/// prints `C:\Users\me/.codex/sessions` in `sessionsPath` and
-/// `additionalPaths[].path`, and the two assertions below now hold it there.
-/// The pin is deliberate but not an endorsement — whether user-facing path
-/// output should be normalized, and to native or to forward separators, is a
-/// product decision tracked in junhoyeo/tokscale#1048. Changing the emitter
-/// means changing this helper in the same commit.
+/// `tokscale clients --json` emits every client's scan root with native
+/// separators throughout: the root half comes from the platform (`C:\Users\me`
+/// on Windows), and the relative half is pushed component-by-component so no
+/// forward slash survives from the `/`-joined client-table literal. An
+/// expectation built with `Path::join` would disagree on the relative half's
+/// separators (`...\.codex/sessions` against the emitted
+/// `...\.codex/sessions`), because `join` only normalizes the junction.
+/// Changing the emitter means changing this helper in the same commit (#1048).
 fn client_scan_path(home: &Path, relative: &str) -> String {
-    format!("{}/{}", home.display(), relative)
+    let mut path = home.to_path_buf();
+    for component in Path::new(relative).components() {
+        path.push(component.as_os_str());
+    }
+    path.to_string_lossy().into_owned()
 }
 
 /// Writes a minimal clawdboard account export to `<dir>/export.json` and
--- a/crates/tokscale-core/src/clients.rs
+++ b/crates/tokscale-core/src/clients.rs
@@ -187,11 +187,18 @@ pub struct ClientDef {
 
 impl ClientDef {
     pub fn resolve_path_with_env_strategy(&self, home_dir: &str, use_env_roots: bool) -> String {
-        format!(
-            "{}/{}",
-            self.root.resolve_with_env_strategy(home_dir, use_env_roots),
-            self.relative_path
-        )
+        let root = self.root.resolve_with_env_strategy(home_dir, use_env_roots);
+        if self.relative_path.is_empty() {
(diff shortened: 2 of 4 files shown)


junhoyeo:
Reviewed and pushed a follow-up commit to this branch directly, since maintainer edits are enabled. The core change is right and the Windows motivation is sound — but cubic's `CODEBUFF_DATA_DIR` thread is correct, and it is a real regression rather than a nit.

**What was wrong.** `join_native` builds its result by pushing components onto a `PathBuf`, and `push` on an empty buffer emits no leading separator. The codebuff override still ran its root through `trim_end_matches('/')` first, so a root of `/` became `""` and the scan root came out as a cwd-relative `projects` instead of `/projects`. The `format!` this replaced produced `/projects` correctly, so this is a behavior change introduced by the PR.

**The fix** is to drop the trim entirely — `PathBuf` already collapses a trailing separator, which is the only thing the trim was buying. I added `test_scan_all_clients_codebuff_override_root_may_end_in_a_separator` pointing the override at a root written with a trailing `/` to pin that, plus `join_native_preserves_an_absolute_root` documenting the contract including the empty-root shape that produced the relative path.

Being straight about the test's limits: `join_native` itself was always correct, so that test does **not** fail if the trim is re-added at the call site. The comment I left at the call site is what carries that warning. Testing `CODEBUFF_DATA_DIR=/` end to end would need a readable `/projects` on the test host, which is not worth it.

`cargo test -p tokscale-core --lib scanner::` is 135/135 and `clients::` is 48/48, fmt clean.

## Two pre-existing sites with the same shape, left alone

Not introduced by this PR, so out of scope — but worth an issue, because they are the same bug:

```rust
PathBuf::from(trimmed.trim_end_matches('/')).join("agent/sessions")   // GJC_CONFIG_DIR / PI_CONFIG_DIR
PathBuf::from(trimmed.trim_end_matches('/')).join("gjc/sessions")     // XDG_DATA_HOME
```

Both turn a root of `/` into a relative path for exactly the same reason.

## Two other things worth your attention

- **The description and the code disagree.** The body says the fix joins through `Path::new(&root).join(self.relative_path)`, but the code pushes each component onto a `PathBuf`. Those differ in an important way — `join` with an absolute argument *replaces* the whole path, whereas the component push does not. The implementation is the safer of the two; the description is just stale and should be updated so a future reader does not reason from the wrong mechanism.
- **The new `clients.rs` test's only cross-platform assertion is a tautology.** The `native_join` test helper is a line-for-line copy of the production algorithm, so on the Linux and macOS legs the test passes unchanged on `main` and proves nothing. Only the `#[cfg(windows)]` "no forward slash remains" assertion is load-bearing. Asserting a hardcoded expected string instead would make it real on every platform.

Smaller: the `client_scan_path` doc comment in `cli_tests.rs` prints the identical string on both sides of a "would disagree" contrast, so the sentence does not demonstrate its point (cubic flagged this too); and the `test_headless_roots_default` fixture is half-converted, with `config_root` on `join_native` while the sibling `mac_root` two lines below still uses `format!`.

PR #1076, opened 2026-08-08
fix(pricing): alias bare minimax-m3 spelling to first-party key

### LINKED ISSUE #935: minimax-m3 pricing missing
Hi, the pricing of minimax-m3 is missing. based on its a pretty good and mutch used model the price should be added.

when i find other models who are not added but provided by ollama cloud, should i open other tickets too? is there some kind of relevance threshold or are they just added on request?

Kind regards

Felictycf:
## Summary

`tokscale submit`/pricing reported **"pricing is unavailable"** (or silently mispriced) for MiniMax M3. Ollama Cloud and other routers record the model with the bare lowercase id `minimax-m3`, but the authoritative dataset key is `minimax/MiniMax-M3` (litellm), with a separate `fireworks_ai/minimax-m3` hosted row. A bare id has no exact hit in the lookup indexes, so it fell through to fuzzy matching — which can land on the fireworks_ai hosted rate instead of the first-party one.

## Fix

Adds an alias in `pricing/aliases.rs`:

```rust
m.insert("minimax-m3", "minimax/MiniMax-M3");
```

`resolve_alias` is case-insensitive, so `minimax-m3`, `MiniMax-M3` and `MINIMAX-M3` all resolve deterministically to the first-party litellm key. Qualified ids (`minimax/minimax-m3`) already resolve via exact match and are deliberately left unaliased.

## Tests

- `resolves_minimax_m3_bare_and_case_variants` — alias unit test.
- `test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias` — integration test proving a bare `minimax-m3` message gets priced through the alias.

Full `tokscale-core` lib suite: **1497 passed, 0 failed**.

Fixes #935

---
## Summary by cubic
Alias the bare MiniMax M3 id `minimax-m3` to the canonical `minimax/MiniMax-M3` so pricing chooses the first‑party rate, even without a provider hint. Fixes #935 and prevents “pricing unavailable”, $0 totals, or hosted mispricing when routers send the bare id.

- **Bug Fixes**
  - Map `minimax-m3` (case-insensitive) to `minimax/MiniMax-M3`, avoiding `fireworks_ai/minimax-m3` and zero‑cost third‑party `models.dev` rows.
  - Tests: Added alias unit test and rebuilt the apply‑pricing integration test to cover the unhinted bare‑id path; it now asserts the resolved key/source and cost. Qualified ids like `minimax/minimax-m3` are unchanged.

<sup>Written for commit 287bb9f1bf3b2cef04b33b0235e23f9510aacc79. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1076?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -9117,31 +9117,88 @@ mod tests {
 
     #[test]
     fn test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias() {
-        // #935: Ollama Cloud and other routers report MiniMax M3 with the
-        // bare lowercase id `minimax-m3`. The authoritative litellm key is
-        // `minimax/MiniMax-M3`; the alias in aliases.rs must route the bare
-        // id there instead of leaving it to fuzzy matching (which can land
-        // on the separate `fireworks_ai/minimax-m3` hosted row).
+        // #935: routers report MiniMax M3 as the bare lowercase id `minimax-m3`,
+        // which is not a key in any dataset. When the session record carries no
+        // usable provider hint — parsers emit `unknown` for an absent provider
+        // and `normalize_provider_hint` drops it — nothing pins the lookup to
+        // MiniMax's catalog, so the bare id falls through to model-part/fuzzy
+        // matching over every row whose model part is `minimax-m3`.
+        //
+        // models.dev publishes that model part under dozens of third parties,
+        // several of them at 0.0/0.0 (`kenari/minimax-m3` and
+        // `nvidia/minimaxai/minimax-m3` both do today). Electing one of those
+        // prices real usage at exactly $0 — which is what "pricing missing"
+        // in #935 looks like from the user's side, since a row of explicit
+        // zeros still counts as "priced" downstream. The alias must pin the
+        // canonical first-party `minimax/MiniMax-M3` key instead.
         let mut litellm = HashMap::new();
+        // Real first-party rates.
         litellm.insert(
             "minimax/MiniMax-M3".into(),
             pricing::ModelPricing {
-                input_cost_per_token: Some(1e-7),
-                output_cost_per_token: Some(2e-7),
+                input_cost_per_token: Some(3e-7),
+                output_cost_per_token: Some(1.2e-6),
                 ..Default::default()
             },
         );
-        let pricing = pricing::PricingService::new(litellm, HashMap::new());
+        // The hosted reseller row that ships alongside it, at a deliberately
+        // far-apart rate so electing it could not be mistaken for the
+        // first-party result.
+        litellm.insert(
+            "fireworks_ai/minimax-m3".into(),
+            pricing::ModelPricing {
+                input_cost_per_token: Some(3e-5),
+                output_cost_per_token: Some(1.2e-4),
+                ..Default::default()
+            },
+        );
+        // The zero-cost third-party row that the bare id actually elects
+        // without the alias.
+        let mut models_dev = HashMap::new();
+        models_dev.insert(
+            "kenari/minimax-m3".to_string(),
+            pricing::ModelPricing {
+                input_cost_per_token: Some(0.0),
+                output_cost_per_token: Some(0.0),
+                ..Default::default()
+            },
+        );
+        let pricing = pricing::PricingService::new_with_custom_and_models_dev(
+            Default::default(),
+            litellm,
+            HashMap::new(),
+            models_dev,
+        );
+
+        // Fixture guards: both competing rows must really be in the dataset and
+        // resolvable, or this test would pass for the wrong reason.
+        let competing_zero = pricing
+            .lookup_with_source_and_provider("kenari/minimax-m3", None, None)
+            .expect("competing zero-cost models.dev row must be present");
+        assert_eq!(competing_zero.matched_key, "kenari/minimax-m3");
+        assert_eq!(competing_zero.pricing.input_cost_per_token, Some(0.0));
+        let competing_hosted = pricing
+            .lookup_with_source_and_provider("minimax-m3", None, Some("fireworks_ai"))
+            .expect("competing fireworks_ai row must resolve under its own hint");
+        assert_eq!(competing_hosted.matched_key, "fireworks_ai/minimax-m3");
+
+        // The behavior the alias exists to guarantee: the bare id resolves to
+        // the canonical first-party key, not to either competitor.
+        let resolved = pricing
+            .lookup_with_source_and_provider("minimax-m3", None, Some("unknown"))
+            .expect("bare `minimax-m3` must resolve");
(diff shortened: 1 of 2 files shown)


junhoyeo:
Reviewed as part of a sweep across the open PRs. The bug is real, but I think the alias target is the wrong shape, and the integration test does not cover the regression.

## The target is the only provider-qualified value in the table

```
$ rg -o 'm\.insert\("[^"]+", "([^"]+)"\)' -r '$1' aliases.rs | grep -c "/"
1
```

That one is `minimax/MiniMax-M3`. Every other target in the table is a bare model id — `glm-4.7`, `kimi-k2-thinking`, `gemini-3.1-pro`, `claude-opus-4-6`. Alias *keys* are sometimes qualified (`anthropic/claude-4-5-opus`), but no target ever was before this.

That matters because `resolve_alias` is not only a pricing lookup: several parsers call it to rewrite the stored `model_id`, so the resolved value becomes the model identity that is displayed, grouped in reports, submitted, and matched against `custom-pricing.json`. With this alias a MiniMax M3 user's report row changes from `minimax-m3` to `minimax/MiniMax-M3`, and a `custom-pricing.json` entry keyed on the bare id stops matching. `m.insert("minimax-m3", "minimax-m3")` or a bare canonical spelling keeps the fix inside pricing resolution where it belongs.

## The stated hazard may not be reachable

The body says the bare id "fell through to fuzzy matching — which can land on the fireworks_ai hosted rate instead of the first-party one." Worth double-checking that: with a `minimax` provider hint, the provider-scoped tier selects `minimax/MiniMax-M3` before fuzzy matching is consulted at all. If that is right, the justifying comment describes a failure that cannot occur, and the real fix is for the no-hint path only. Either way the comment should describe the path that actually breaks, since that is what the next person will reason from.

## The integration test passes without the alias

cubic's thread on `lib.rs` is correct, and it is the important one. The test supplies a provider hint that already selects `minimax/MiniMax-M3` by exact match, so it goes green on `main` with the alias removed and detects nothing. To make it load-bearing it needs a setup where provider-aware exact matching cannot apply — no hint, competing `fireworks_ai/minimax-m3` row present — and it should assert the resolved key and the resulting cost, not just that pricing was found.

cubic's second thread is also right: the new test was inserted between the Anthropic regression doc comment and the test it documents, so that comment now sits above the MiniMax test and misdescribes it.

Worth confirming the two candidate rows actually price differently before leaning on the mispricing framing — if `minimax/MiniMax-M3` and `fireworks_ai/minimax-m3` are price-identical, the user-visible bug is only the "pricing is unavailable" message, which is still worth fixing but is a different claim than mispricing.

junhoyeo:
Thanks @Felictycf — the production fix is correct and I've kept it exactly as you wrote it. I pushed 287bb9f1 to this branch to make the tests prove what they claim, because the regression test as written could not have failed.

## Why the old test could not fail

`test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias` built a litellm fixture containing exactly one row, `minimax/MiniMax-M3`, and sent a message whose provider hint was `"minimax"`. Provider-aware exact matching selects that key from the hint alone, so the assertion held whether or not `resolve_alias` knew about `minimax-m3`. The comment named `fireworks_ai/minimax-m3` as the row being guarded against, but that row was never added to the fixture.

I confirmed this by deleting the alias and re-running the original test unchanged — it still passed.

## What I found while building a test that does fail

The stated failure mode (fuzzy matching landing on `fireworks_ai/minimax-m3`) turns out not to be reproducible within litellm alone. `select_best_match` prefers `ORIGINAL_PROVIDER_PREFIXES` over `RESELLER_PROVIDER_PREFIXES`, and `minimax/` is in the first list while `fireworks_ai/` is in the second, so the first-party row already wins that contest. I verified this against the live litellm dataset across a matrix of ids and provider hints: alias on vs. alias off produced **identical** resolutions in every case.

The alias *is* load-bearing, just for a different reason — cross-source resolution when no provider hint pins the lookup. models.dev publishes the `minimax-m3` model part under dozens of third parties, and several of them price it at `0.0/0.0` today (`kenari/minimax-m3` and `nvidia/minimaxai/minimax-m3` both do). Against the real combined dataset, unhinted `minimax-m3` resolves to a models.dev third-party row without the alias and to litellm's `minimax/MiniMax-M3` with it. When the elected row is one of the zero-cost ones the usage prices at exactly `$0` — which is precisely the "pricing missing" symptom reported in #935, since a row of explicit zeros still counts as priced downstream.

That path is reachable in practice: several parsers emit `unknown` when a session record has no provider, and `normalize_provider_hint` drops `unknown`, leaving the lookup unhinted.

## What the new test pins

The fixture now contains all three rows the resolver has to choose between — first-party `minimax/MiniMax-M3`, the `fireworks_ai/minimax-m3` hosted row at a rate two orders of magnitude away, and a zero-cost models.dev row — with the provider hint left absent so the alias is the only thing that can decide the outcome. It asserts the resolved **key and source**, not just a cost, plus fixture guards that both competing rows are really present and resolvable so the test cannot silently go vacuous again.

## Red-green evidence

Alias commented out in `aliases.rs`:

```
running 2 tests
test pricing::aliases::tests::resolves_minimax_m3_bare_and_case_variants ... FAILED
test tests::test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias ... FAILED

---- tests::test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias stdout ----
assertion `left == right` failed
  left: "kenari/minimax-m3"
 right: "minimax/MiniMax-M3"

test result: FAILED. 0 passed; 2 failed; 0 ignored; 0 measured; 1495 filtered out
```

Alias restored:

```
running 2 tests
test pricing::aliases::tests::resolves_minimax_m3_bare_and_case_variants ... ok
test tests::test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 1495 filtered out
```

## Also in this commit

The rationale comment above the alias now describes the failure mode I could actually reproduce rather than the fireworks one. And the `resolves_minimax_m3_bare_and_case_variants` test was carrying a doc comment describing an Anthropic `-0` moving-alias regression, which it does not cover — that block moved to `anthropic_moving_aliases_and_copilot_spelling_resolve`, and the MiniMax test got its own.

## Verification

`cargo fmt --all -- --check` clean, `cargo clippy -p tokscale-core --all-targets` clean, `cargo test -p tokscale-core` green (1496 lib tests + 35 integration tests, 0 failures).

Issue #960, opened 2026-07-25
Timezone re-splits permanently inflate stored usage; design for correcting existing data


junhoyeo:
Rescanning the same local history under a different timezone permanently inflates a user's stored totals. This describes the mechanism, why the obvious fixes are unsound, and a three-layer plan that corrects existing data without guessing.

Context: `d9df8c9c` (monotonic guard) and #955 (per-device session metrics) both touched this area. Neither addressed this, and #955's floor makes the frozen values explicit.

---

## Mechanism

`compute_daily_active_time` (`crates/tokscale-core/src/sessionize.rs:288`) apportions each session across **local calendar days** via `chrono::Local`. Token attribution follows message timestamps into the same local-day buckets. So the day a unit of usage lands in is a function of the machine's timezone at scan time.

`mergeClientBreakdownsWithRegressionGuard` (`packages/frontend/src/lib/db/helpers.ts:154`) then defends every per-day, per-client decrease. Combined, a timezone change inflates:

A session crossing local midnight, scanned once under `Asia/Seoul` and once under `UTC`:

| | Jan 1 | Jan 2 | stored total |
|---|---|---|---|
| first scan (Seoul) | — | 1000 | 1000 |
| rescan (UTC) — usage moves to Jan 1 | 1000 | *guard defends the stale 1000* | **2000** |

Truth is 1000. The value is idempotent (it converges, it does not grow without bound) but it has **no correction path** — every subsequent submit re-defends it.

Same mechanism drives the `active_time_ms` case; `totalActiveTimeMs` observed settling at ~25.2M ms for a 4h session.

## Why the guard exists

Deliberate, and correct for its threat model (`helpers.ts:219`):

> A token decrease alone signals a parser regression (e.g. the CLI re-parsed only a subset of history). Preserve the existing row even when coverage metrics are equal, because equal coverage + fewer tokens still indicates data loss. The old AND-gate (tokens < existing AND lower coverage) let equal-coverage regressions slip through undetected.

The OR-gate was chosen over an AND-gate on purpose. The guard is not the bug — it cannot distinguish a legitimate re-split from a partial re-parse, because **the payload carries no signal that would let it**.

## Why stored data cannot simply be repaired

Each cell `(device, day, client)` holds `max over every submit ever made`. Max is lossy: the value the most recent complete scan actually reported is not recoverable from the database. Reconstructing it from what is stored would be fabrication.

[...]

PR #1083, opened 2026-08-09
feat(submit): expose reconciliation census gate inputs

IvGolovach:
## Summary

- add an admin-only `GET /api/admin/reconciliation/census` endpoint that exposes reconciliation gate inputs without changing served totals
- count a user as measured only when every expected device/client/origin/month cell has a matching high-water row and no durable census work remains
- classify divergence, report explicit cell comparisons, segment coverage, and return a bounded repair-candidate sample
- execute the production census SQL against a migrated PostgreSQL 16 fixture in CI
- treat a fully covered zero-token user as a valid clean `0/0` observation, while keeping undefined zero-denominator cases out of candidate ranking

Advances the evidence-gathering path for #960. It does not enable automatic repair or change leaderboard totals.

## Why

The project already records guarded daily totals and per-device high-water buckets, but operators had no single read surface that could prove whether the population was sufficiently measured before a recovery rollout. Treating an absent measurement as zero would manufacture severe regressions during warm-up, while treating a durable zero-valued measurement as absent would keep valid zero-token users permanently warming.

This report separates those states: expected cells must be explicitly covered, pending work blocks readiness, and a fully measured `0/0` user is clean rather than missing.

## Safety model

- concealed behind the existing admin-session guard; non-admin callers receive the same `404` used by the moderation surface
- one read-only PostgreSQL snapshot with no writable CTEs, locks, scheduled work, or repair side effects
- `private, no-store` response; bigint token totals remain strings; the database payload is validated before it leaves the server
- candidate limits are normalized and capped at 100
- incomplete coverage remains `pending` or `warming`; only explicit complete evidence can enter a divergence band

## Diff scope

- census report contract and SQL snapshot query
- authenticated read-only API route
- focused unit and route tests
- PostgreSQL integration fixture and CI invocation
- operator-facing recovery documentation

No schema migration, submission write path, public leaderboard path, dependency, or version manifest changed.

- base: `origin/main` at `9814fa49e8ba32b19d94ef2b1545b66b17944435`
- ahead/behind against the fetched base: 4 ahead / 44 behind
- merge base: `d1d2c2d62d8095b5f287c7986fa5ed5e3addb97d`
- GitHub reports the PR mergeable; final pull-request CI is the required synthetic-merge proof against current `main`
- introduced commits:
  - `f023a440a201576173834f7e1e8772a388e8a531 feat(submit): expose reconciliation census gate inputs`
  - `c3e253aef539eb8978172a7040c8e93ff4fcbf59 test(reconciliation): verify census SQL against postgres instead of a mock`
  - `dcb49c7662b41c440bbc40b52a0a956347bba5da chore(frontend): set the census integration flag in its own test script`
  - `5625cb07314f11846b18d8c358bf162d306dad4d fix(submit): classify covered zero-token census users`
- multiple commits are intentional post-review corrections within one census change; the final PR diff contains only the intended eight files
- version: not applicable — release tooling owns version synchronization

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: eight intended files
- no secrets, local environment files, caches, build output, coverage artifacts, or unrelated generated files are included

Validation mode: Mode 4 — the runtime change is a read-only admin query and the PR extends CI verification for its PostgreSQL contract.

- targeted census unit and route Vitest: PASS, 20 tests
- full frontend Vitest: PASS, 858 tests; the six opt-in PostgreSQL tests were collected but require the CI database flag
- frontend TypeScript typecheck: PASS
- targeted ESLint for every changed TypeScript runtime/test file: PASS
- PostgreSQL integration locally: not run — mandatory Frontend Migration Replay runs the fixture against PostgreSQL 16 on the final PR SHA

## Remote proof on final SHA

All checks passed on `5625cb07314f11846b18d8c358bf162d306dad4d`:

- Frontend Migration Replay, including the PostgreSQL 16 census fixture
- Frontend Vitest
- Frontend Container
- Actionlint
- Vercel preview
- Cubic review
- repository merge/WIP gates

## Runtime and migration notes

No migration changed. The PostgreSQL fixture runs after the existing migration replay and deletes its own users through schema cascades. Runtime code adds no write, queue, background poll, or served-total switch.

## Rollback

Rollback: revert this PR. DB downgrade: not applicable. Data repair: not applicable. Operational caveats: none known.

## Known residual risk

The census performs full-site aggregates, so execution cost scales with reconciliation history. The route is admin-only and unscheduled, and the candidate payload is bounded; production latency should still be observed before using it as a frequently refreshed dashboard.

### DIFF
--- a/.github/workflows/frontend_migration_ci.yml
+++ b/.github/workflows/frontend_migration_ci.yml
@@ -12,6 +12,8 @@ on:
       - "packages/frontend/drizzle.config.ts"
       - "packages/frontend/scripts/check-migrations.ts"
       - "packages/frontend/src/lib/db/**"
+      - "packages/frontend/src/lib/reconciliation/ratchetCensus.ts"
+      - "packages/frontend/__tests__/integration/ratchetCensusPostgres.test.ts"
       - ".github/workflows/frontend_migration_ci.yml"
   pull_request:
     branches: [main]
@@ -62,3 +64,8 @@ jobs:
 
       - name: Replay frontend database migrations
         run: bun run --cwd packages/frontend test:migrations
+
+      - name: Execute reconciliation census against PostgreSQL
+        env:
+          RATCHET_CENSUS_DB_INTEGRATION: "1"
+        run: bun run --cwd packages/frontend test:reconciliation:db
--- a/docs/ratchet-inflation-recovery.md
+++ b/docs/ratchet-inflation-recovery.md
@@ -910,6 +910,15 @@ rather than on "a census" generally. Do not build them against the active-time
 numbers above; those are an incidence signal on an unrepresentative slice, not a
 magnitude estimate.
 
+The admin-only `GET /api/admin/reconciliation/census` endpoint exposes the gate
+inputs without changing any served value. It counts a user as measured only when
+every month/client/origin cell visible in their guarded daily rows has a matching
+device high-water row and no durable census work remains. It also compares only
+explicitly observed Phase 4a cells; omission is never interpreted as zero. The
+endpoint reports coverage, divergence bands, client/origin/version segments and
+a bounded outlier sample. It intentionally does not choose a Phase 2 rollout
+threshold or perform a repair.
+
 **Do not size the work from the worst ratios.** The 2300x device is a
 partial-scan artifact. Four devices are cleanly artifact, eleven ambiguous.
 
--- a/packages/frontend/__tests__/api/ratchetCensusRoute.test.ts
+++ b/packages/frontend/__tests__/api/ratchetCensusRoute.test.ts
@@ -1,0 +1,105 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { NextResponse } from "next/server";
+
+const { mockRequireAdminSession, mockGetRatchetCensusReport } = vi.hoisted(() => ({
+  mockRequireAdminSession: vi.fn(),
+  mockGetRatchetCensusReport: vi.fn(),
+}));
+
+vi.mock("@/lib/moderation/guard", () => ({
+  requireAdminSession: mockRequireAdminSession,
+}));
+
+vi.mock("@/lib/reconciliation/ratchetCensus", async (importOriginal) => {
+  const actual =
+    await importOriginal<typeof import("@/lib/reconciliation/ratchetCensus")>();
+
+  return {
+    ...actual,
+    getRatchetCensusReport: mockGetRatchetCensusReport,
+  };
+});
+
+import { GET } from "@/app/api/admin/reconciliation/census/route";
+
+describe("GET /api/admin/reconciliation/census", () => {
+  beforeEach(() => {
+    mockRequireAdminSession.mockReset();
+    mockGetRatchetCensusReport.mockReset();
+  });
+
+  it("returns the concealed admin response without touching census data", async () => {
+    const concealed = NextResponse.json({ error: "Not found" }, { status: 404 });
+    mockRequireAdminSession.mockResolvedValue({ response: concealed });
+
+    const response = await GET(
+      new Request("https://tokscale.ai/api/admin/reconciliation/census")
+    );
+
+    expect(response.status).toBe(404);
+    expect(mockGetRatchetCensusReport).not.toHaveBeenCalled();
+  });
+
+  it("returns a private non-cacheable report to an admin", async () => {
+    mockRequireAdminSession.mockResolvedValue({ session: { id: "admin" } });
+    mockGetRatchetCensusReport.mockResolvedValue({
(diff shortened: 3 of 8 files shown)


junhoyeo:
Thanks @IvGolovach — your latest push already did most of the work here: the real Postgres integration suite, the `expectedBuckets`/`measuredBuckets` → `expectedCells`/`measuredCells` rename, and switching the route test to `importOriginal` so it exercises the real `normalizeRatchetCensusCandidateLimit`. I verified each of those and pushed two follow-up commits for the gaps that were left.

## How the SQL contract is now verified

I ran the suite against a migrated PostgreSQL 16 and then mutation-tested the statement: change one thing in the SQL, confirm a test goes red, restore it, confirm green. On the fixture as it stood, **3 of the 7 `::text` casts and 6 other properties survived mutation** — the test stayed green with the SQL broken:

| Mutation | Before | After |
| --- | --- | --- |
| drop `::text` on `byCliVersion.totalTokens` / `.measuredTokens` | passed | fails |
| drop `::text` on `candidates.highwaterTokens` | passed | fails |
| `cellCoverage` rounding `ROUND(..., 6)` → `ROUND(..., 1)` | passed | fails |
| user band `under` cutoff `0.95` → `0.5` | passed | fails |
| user band `clean` cutoff `1.05` → `3.0` | passed | fails |
| coverage join stops matching `t.origin = e.origin` | passed | fails |
| coverage join stops matching `t.bucket_key = e.bucket_key` | passed | fails |
| `fully_measured`: `measured_cells = expected_cells` → `>= 0` | passed | fails |

The causes were that `byCliVersion` was never asserted at all, and that `highwaterTotal` was `3000000000000000` — comfortably under 2^53, so it round-trips through a JS double unchanged and the cast was doing nothing observable.

`c3e253a` rebuilds the fixture around seven personas — one per divergence band, plus a partially measured user whose two unmeasured cells each carry a high-water row differing *only* by origin or *only* by bucket key. Token totals are now `90071992547409937` and `9007199254740993` (2^53 + 1), neither of which survives a double round trip, and both site and segment coverage ratios are non-terminating (`4/7` → `0.571429`, `6/7` → `0.857143`, `5/6` → `0.833333`) so the 6-digit rounding is genuinely pinned rather than accidentally agreeing.

The bigint case is now concrete. Removing the cast on the candidate high-water column produces exactly the silent corruption the cast exists to prevent:

```
-   "highwaterTokens": "9007199254740993",
+   "highwaterTokens": "9007199254740992",
```

**25 of 25 mutants are now killed, zero survivors** — covering all 7 `::text` casts, both coverage-rounding sites, all 4 user-band cutoffs, all 4 observed-cell band cutoffs, `MAX(ratio)`, the provenance-origin extraction, both coverage-join predicates, the `pending_work` and cell-equality terms of `fully_measured`, the `source_breakdown ? r.client` guard, the `pending` band precedence, and the `LIMIT` binding.

## What remains unverified

Two things, stated plainly rather than papered over:

1. **The census suite is not part of `bun run test`.** It asserts site-wide totals, so it needs a database it owns. It runs in Frontend Migration CI, whose path filter includes `ratchetCensus.ts` and the test file, so any edit to the SQL does trigger it — but a plain frontend unit-test run still proves nothing about the SQL.
2. **The fixture is single-device per user and single-month for most personas.** Multi-device users and cross-month high-water aggregation are exercised only through the `partial` persona's three cells, not exhaustively.

`dcb49c7` closes a related hole: `test:reconciliation:db` ran a suite that skips itself unless `RATCHET_CENSUS_DB_INTEGRATION=1`, so the script exited 0 while running nothing if that variable were ever dropped from the workflow. The script now sets it itself, so the job cannot go green on a silently skipped file.

## Rename and normalization

The rename is complete — no `expectedBuckets`/`measuredBuckets` remain in `src/`, `__tests__/`, or `docs/`. The surviving `bucket_key`/`bucket_width` identifiers are the real schema column names and are correctly left alone.

For the normalization mock, using the real function was the right call. I mutation-tested it: removing the `Math.min` cap, relaxing `Number.isInteger` to `Number.isFinite`, and dropping the `<= 0` guard each turn **both** the direct unit tests and the route tests red. One mutant survives — deleting the `value.trim() !== ""` check — but that one is behaviour-preserving rather than a coverage gap, since `Number("")` is `0` and the existing `<= 0` guard already rejects it.

## Gates

- `bun run --cwd packages/frontend test` — 858 passed, 6 skipped (the census suite), 87 files
- `bun run --cwd packages/frontend test:reconciliation:db` — 6 passed against PostgreSQL 16
- `bun run --cwd packages/frontend typecheck` — clean
- `bun run --cwd packages/frontend lint` — 0 errors (17 pre-existing warnings, none in these files)

No behaviour changes beyond the rename; `ratchetCensus.ts` is byte-identical to your version.

PR #1084, opened 2026-08-09
fix(pricing): separate estimates from submission-safe matches

### LINKED ISSUE #1062: Fuzzy model matching attaches another model's rates instead of excluding, so unpriced usage submits at wrong or $0 cost

IvGolovach:
## Summary

- attach explicit resolution evidence to pricing lookups, including resolution kind, candidate count, price consensus, model-identity proof, and transformations applied during matching
- keep lenient fuzzy pricing available for local estimates while refusing unsafe evidence at the leaderboard submission boundary
- judge borrowed bucket rates on the combined evidence of both the provider-scoped row and the canonical donor row
- expose the exact usage-aware composed result to submission diagnostics so an ambiguous borrowed rate is reported as ambiguous rather than merely incomplete
- surface pricing provenance in `tokscale pricing` text and JSON output

Closes #1062.

## Why

Catalog matching intentionally tolerates provider prefixes, aliases, and fuzzy neighbours because that is useful for estimates. Submission is a stricter trust boundary: once a rate is accepted as complete, the calculated cost is persisted and shown as authoritative.

Previously, lookup did not preserve enough evidence to distinguish a deterministic match from a guess. A provider-scoped row could also borrow a missing bucket rate from an ambiguous canonical row, be correctly rejected by the coverage gate, and still produce the wrong user-facing reason because diagnostics inspected a different plain lookup. This PR makes coverage and diagnostics consume the same usage-aware result.

## Resolution contract

`LookupResult` carries `ResolutionEvidence` describing exact/model-part/provider/fuzzy/custom resolution, candidate count, full rate-vector consensus, exact terminal identity, aliasing, normalization, prefix stripping, and submission safety. Evidence is preserved through the lookup cache and weakened when a result borrows rates from a less trustworthy donor.

Deterministic and user-authored custom pricing remain submission-safe. Fuzzy pricing remains visible for reporting, but can cover submission only when identity and rate evidence meet the stricter gate.

## Diff scope

- `crates/tokscale-core/src/pricing/lookup.rs` — evidence, candidate consensus, donor composition, and lookup-path annotation
- `crates/tokscale-core/src/pricing/mod.rs` — service-level usage-aware resolution and coverage
- `crates/tokscale-core/src/lib.rs` — submission exclusion reasons and regression coverage
- `crates/tokscale-core/src/pricing/custom.rs` — custom-pricing evidence
- `crates/tokscale-cli/src/main.rs` and `crates/tokscale-cli/tests/cli_tests.rs` — text/JSON provenance output

No pricing dataset, dependency, migration, workflow, public submission payload, or release manifest changed. The latest `main` was merged after resolving the sole overlap in the `lib.rs` test import list by preserving both the monthly-report types from `main` and the provenance-specific exclusion constants from this PR.

- base: `origin/main` at `e7293751e8ec2af7dc7d57acbd37c5401cb66885`
- ahead/behind: 7 ahead / 0 behind
- merge base: `e7293751e8ec2af7dc7d57acbd37c5401cb66885`
- GitHub reports the PR mergeable after semantic conflict resolution
- introduced commits:
  - `04016cbb48003a9df187bf101922a4f29c090154 fix(pricing): separate estimates from submission-safe matches`
  - `c2e4f691f9683b27cd533b2821c1442192e62a46 fix(pricing): judge borrowed rates on the evidence of the row they came from`
  - `20fb045475eb6d2f6cf4d7e70c9fc86475652f2e fix(pricing): report custom matches reached through model normalization as normalized`
  - `b79dd3ccf7e38cdf21dba2cfc7638dd65f542e6f fix(pricing): name unverified model identity separately from candidate price disagreement`
  - `f3eb17fa36de5f476a6b541391e0b3ad18664c68 fix(pricing): expose usage-aware submission evidence`
  - `d5e57d23d0ef3880e281191e1a74691b2c442508 chore(pricing): synchronize provenance logic with main`
  - `1fa9f03a665c2d4925a73ca134bad432a679f8c9 chore(pricing): integrate latest main`
- multiple commits are intentional same-scope review corrections plus base synchronization; the final PR diff contains only the intended six files
- version: not applicable — release tooling owns version synchronization; the version carried by the merge is identical to the validated base

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: six intended files
- no conflict markers or unmerged index entries remain
- no secrets, local environment files, caches, build output, coverage artifacts, migrations, or unrelated generated files are included

Validation mode: Mode 2R for the final conflict correction, following Mode 3 proof for the shared pricing and submission behavior. The only new resolution is a test-import union; mandatory remote CI will validate the final SHA before merge.

- `cargo fmt --all -- --check`: PASS
- `cargo test -p tokscale-core pricing::lookup::tests`: PASS, 202 tests on the final synchronized tree
- `cargo test -p tokscale-core submission_`: PASS, 12 tests on the final synchronized tree
- `git diff --check`: PASS
- `git diff --cached --check`: PASS before the merge commit
- full workspace validation locally: not run for the narrow post-review conflict correction; mandatory remote Linux/Windows tests, lint, native builds, launcher smoke, and container build provide final-SHA proof

## Remote proof on final SHA

All checks passed on `1fa9f03a665c2d4925a73ca134bad432a679f8c9`:

- Linux and Windows test suites
- lint
- all nine native build targets across macOS, Linux, Windows, and Android
- launcher smoke and TUI container
- Vercel preview
- Cubic review
- repository merge/WIP gates

GitHub reports `CLEAN` / `MERGEABLE`; every required final-SHA gate has completed successfully.

## Runtime safety

- no new network request, blocking lock, queue, or unbounded collection
- submission becomes more conservative while reporting keeps its lenient estimate path
- diagnostics inspect the exact result that made the row ineligible

## Rollback

Rollback: revert this PR. DB downgrade: not applicable. Data repair: not applicable. Excluded rows remain local and can be submitted after authoritative pricing becomes available.

## Known residual risk

Price consensus intentionally requires an exact complete rate-vector match. Economically equivalent entries with structurally different optional fields remain estimate-only; this is a conservative false negative rather than silent mispricing.


junhoyeo:
Thanks @IvGolovach — the estimate/submission split here is the right shape, and the `ResolutionEvidence` provenance it adds is what made the remaining gaps findable at all. I pushed three follow-up commits to `codex/pricing-provenance` on top of `04016cbb`.

## Submission could still publish an ambiguous rate

`resolve_for_usage` fills the buckets a provider-hinted row cannot price from the canonical unhinted row, then returns `LookupResult { pricing: filled, ..hinted }` — which keeps the **hinted** row's evidence. So a hinted row that is submission-safe on its own could borrow a rate from a canonical resolution the resolver had already refused to publish, and the combined row was reported as safe.

Reproduced against the branch before the fix, with a deterministic `azure_ai/atlas-chat` hint and an ambiguous fuzzy canonical match:

```
HINTED:    ("azure_ai/atlas-chat", kind: ModelPart, ..., submission_safe: true)
CANONICAL: ("vendor-a/atlas-chat-preview", kind: Fuzzy, candidate_count: 3,
            price_consensus: false, exact_model_identity: false, submission_safe: false)
RESOLVED:  ("azure_ai/atlas-chat", kind: ModelPart, ..., submission_safe: true, cache_read: Some(5e-7))
COVERS:    true
```

The published cache-read rate is `5e-7` from `vendor-a`, while `vendor-b` quotes `9e-7` for the same fuzzy match — the resolver picked one of two conflicting answers and submitted it.

The fix composes the evidence rather than rejecting the fill: `ResolutionEvidence::borrowing_from` judges the filled row on the weaker of the two resolutions it quotes. That keeps your separation intact in both directions — the estimate stays visible with the borrowed rate, only its publishability changes, and a borrow from a submission-safe canonical row still submits exactly as before (that is the #1013 behaviour the borrow exists for, covered by its own guard test).

Red before, green after:

```
---- borrowed_rates_from_an_ambiguous_canonical_row_are_not_submission_safe stdout ----
panicked at crates/tokscale-core/src/pricing/lookup.rs:5356:9:
a rate borrowed from an ambiguous canonical row must not be publishable through a submission-safe hinted row
test result: FAILED. 1 passed; 1 failed
```

```
test borrowed_rates_from_an_ambiguous_canonical_row_are_not_submission_safe ... ok
test borrowed_rates_from_a_submission_safe_canonical_row_still_submit ... ok
```

The guard test passes before and after, which is the point: this does not tighten submission for genuinely safe matches.

## Custom matches through synthetic normalization reported as unnormalized

`CustomPricing::lookup_with_key` falls back to `normalize_synthetic_model`, so a `glm-4.7` custom entry prices `hf:zai-org/GLM-4.7` — but `lookup_custom` hardcoded `normalized: false`, so provenance claimed the matched key was the id as written. `CustomLookupResult` now carries which path matched.

```
---- custom_match_through_synthetic_normalization_reports_normalized_provenance stdout ----
panicked at crates/tokscale-core/src/pricing/mod.rs:1727:9:
a match reached through synthetic-model normalization is normalized
```
after: `test custom_match_through_synthetic_normalization_reports_normalized_provenance ... ok`

## Exclusion reason contradicted the code that emitted it

A fuzzy lookup with a single candidate is excluded because nothing proves the priced key names the model that was used — not because candidates disagreed. There is only one candidate, so `price_consensus` is vacuously true and nothing can disagree, yet the row was reported as `"model price lookup is ambiguous across non-equivalent candidates"`. Same for multiple candidates that agree on price.

```
---- submission_excludes_single_candidate_fuzzy_price_for_unverified_identity stdout ----
assertion `left == right` failed
  left: "model price lookup is ambiguous across non-equivalent candidates"
 right: "model price match does not exactly name the requested model"
```
after: `test submission_excludes_single_candidate_fuzzy_price_for_unverified_identity ... ok`

Rather than adding a second condition next to the first, the reason is now derived from `ResolutionEvidence::submission_safety_gap()` — the same match that decides whether a row is publishable also names why it is not, so the diagnostic cannot drift from the rule. `is_submission_safe()` is now defined as `submission_safety_gap().is_none()`, so the safety rule itself is unchanged, and your `submission_excludes_ambiguous_fuzzy_price_with_specific_reason` test still gets the ambiguity reason it asserts.

## Gates

```
cargo fmt --all -- --check          clean
cargo clippy -p tokscale-core --all-targets   clean, no warnings
cargo test -p tokscale-core         1522 passed; 0 failed; 1 ignored
cargo test -p tokscale-cli -- pricing   18 passed; 0 failed
```

Both intermediate commits were checked to compile on their own, so the series stays bisectable.

Issue #994, opened 2026-07-31
Historical totals silently drift when Claude Code rewrites session files in place (repro + numbers)


lizhuojunx86:
## Summary

Claude Code rewrites session JSONL files in place on resume/compact: between
two scans of the same file on a 34-day corpus (~26k messages), one transcript
lost 5 assistant messages and the file mtime bumped. tokscale recomputes
totals from the live files on every run, so those messages' tokens silently
leave history — and since `tokscale submit` feeds the leaderboard, submitted
numbers drift the same way.

To be clear about what is already right: your per-message accounting is
exact. On a synthetic corpus with a known manifest (540 distinct assistant
messages, streaming multi-line records with repeated `message.id`), a cold
run of `tokscale@4.7.0` matches ground truth to the digit on output,
cacheRead, cacheWrite and message count, including transcripts under
`subagents/**`. The dedup key, the per-field max merge for streaming
partials, and the deep discovery walk all check out. This report is about
persistence only.

## The measurement

Rewrite one transcript in place, dropping its last 6 assistant messages (the
observable effect of a real resume/compact), then run again:

```
field         before          after           drifted prediction
output        1,115,321       1,104,872       1,104,872
cacheRead     24,447,923      24,192,633      24,192,633
cacheWrite    8,130,148       8,052,730       8,052,730
messages      540             534             534
```

Every field lands on the drifted prediction exactly, which pins the cause to
recompute-from-live-files rather than anything in parsing. The
source-message cache behaves as designed — `message_cache.rs` documents that
the cache must invalidate when the file changes, and
`load_or_parse_source_with_fingerprint_and_policy` (`lib.rs:745`) re-parses
and replaces on any fingerprint change — so the design tracks live file
content, and live file content is not stable.

## Repro

Self-contained, ~1 minute, no real `~/.claude` data involved:

```bash
git clone https://github.com/lizhuojunx86/traceguard
cd traceguard/usage-tracker-audit/tokscale-drift-check
./run_check.sh
```

It installs the published npm package (pinned), generates the corpus with an
exact manifest under an isolated fake `$HOME`, runs three passes
(cold / warm / post-rewrite), and compares. Exit 0 = totals survive the
rewrite, 1 = drift — so the same script doubles as a regression test after a
fix.

## Fix shape that has shipped elsewhere

Two trackers in this space already answered this with a keyed
freeze-at-first-sight store: splitrail added an append-only SQLite history
[...]

PR #1085, opened 2026-08-09
fix(cache): preserve completeness when merging retained and live messages

IvGolovach:
## Summary

- persist which Claude assistant messages were retained from an earlier transcript version so retained/live provenance survives cache round trips
- replace path-order, first-wins deduplication with a monotonic completeness merge for globally stable Claude response identities
- keep live metadata authoritative while combining token, duration, message-count, conflict, workspace, session, agent, and provider-cost evidence without double counting
- apply the same completeness contract when concurrent cache writers merge retained history for the same source
- preserve the current cache format, Prime Agent accounting, reconciliation census, and pricing-provenance behavior while integrating the Claude retention changes

This is a focused follow-up to #994.

## Why

Claude Code can replay one API response into a forked transcript. A retained copy from the original transcript and a live copy from the fork represent one logical turn, but the previous lexical path-order deduplication could keep an earlier streaming partial and discard the completed replay. The same weakness existed when concurrent cache writers merged duplicate identities.

The fix merges completeness monotonically and carries retained/live provenance through persistence, so a completed observation cannot be replaced by a partial one and one stable response identity still contributes at most once.

## Completeness contract

- token fields, reasoning tokens, duration, and message count merge by per-field maximum
- live records remain authoritative when retained and live copies meet
- missing workspace, session-title, and agent metadata may be filled from the other observation
- provider-reported cost is preserved when only one observation carries it
- derived fields and estimated pricing are refreshed after token values combine
- path-scoped and unkeyed messages remain ineligible for cross-file retention

## Diff scope

- `crates/tokscale-core/src/lib.rs` — retained/live provenance and cross-file Claude merge
- `crates/tokscale-core/src/message_cache.rs` — persisted retained identities, concurrent-writer completeness merge, and semantic integration with cache migration behavior
- `crates/tokscale-core/src/sessions/claudecode.rs` — shared monotonic completeness merge

No public API, submission payload, database schema, dependency, workflow, or release manifest changed. The latest `main` merged without textual conflicts, and the final diff remains limited to these three cache-related files.

- base: `origin/main` at `92a5fee9a7a25758b6e32fae777d608a434e6d5a`
- ahead/behind: 4 ahead / 0 behind
- merge base: `92a5fee9a7a25758b6e32fae777d608a434e6d5a`
- introduced commits:
  - `dc612d8a6ff383142405c9964dacde87b1e37980 fix(cache): merge completed Claude replays over partials`
  - `84efdf0f76010d7372bd54ac48a8de31f38555df fix(cache): rebuild Claude cache entries written before retention provenance`
  - `20bca4eddc9f8e42a50772ebd02d91d525885222 chore(cache): synchronize retention cache with main`
  - `0d0d77851ffb8341ac048d6aee67eea665f142be chore(cache): integrate latest main`
- multiple commits are intentional same-scope corrections plus base synchronization; the final PR diff contains only the intended three files
- version: not applicable — release provenance is inherited from `main`, with no version delta in the final PR diff

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: three intended files
- no conflict markers or unmerged index entries remain
- no secrets, local environment files, caches, build output, coverage artifacts, migrations, or unrelated generated files are included

Validation mode: Mode 2R for the final base synchronization, following Mode 3 proof for cache persistence and retained-history data integrity. The cache-completeness diff is unchanged, merged cleanly with current `main`, and mandatory remote CI will validate the final SHA before merge.

- `cargo fmt --all -- --check`: PASS
- `cargo test -p tokscale-core message_cache`: PASS, 65 tests
- `cargo test -p tokscale-core test_claude_`: PASS, 14 tests
- `cargo test -p tokscale-core test_prime_agent_legacy_cache_backfills_accounting_once`: PASS, 1 test
- `cargo +stable clippy -p tokscale-core --all-targets -- -D warnings`: PASS via the rustup-managed stable toolchain
- `git diff --check`: PASS
- `git diff --cached --check`: PASS before the merge commit
- full local workspace suite: not run for the narrow post-review base synchronization; mandatory remote Linux/Windows tests, lint, native builds, launcher smoke, and container build provide final-SHA proof

## Remote proof on final SHA

All checks passed on `0d0d77851ffb8341ac048d6aee67eea665f142be`:

- Linux and Windows test suites
- lint
- all nine native build targets across macOS, Linux, Windows, and Android
- launcher smoke and TUI container
- Vercel preview
- Cubic review
- repository merge/WIP gates

GitHub reports `CLEAN` / `MERGEABLE`; every required final-SHA gate has completed successfully.

## Review status

The sole review finding concerned legacy Claude cache entries that lacked retained-message provenance. It was corrected in `84efdf0f76010d7372bd54ac48a8de31f38555df`, covered by the warm-versus-cold parity and one-time rebuild regressions, and the review thread is resolved.

## Runtime safety

- behavior is scoped to the Claude parser namespace; Prime Agent cache behavior from `main` remains intact
- deleting a transcript still removes its cache entry through existing missing-file pruning
- no new network request, blocking lock, queue, or unbounded collection

## Rollback

Rollback: revert this PR. DB downgrade: not applicable. Data repair: not applicable. Existing cache entries remain readable and later scans can rebuild derived cache state from retained source entries.

## Known residual risk

The merge is intentionally monotonic and does not reconstruct event ordering. For contradictory non-monotonic metadata, the authoritative live record wins when provenance distinguishes it; otherwise the existing observation is retained.


junhoyeo:
Thanks @IvGolovach — the completeness merge is the right call, and the retained/live provenance you added is what makes it work. I pushed one follow-up commit to `codex/cache-completeness` (84efdf0f) covering the upgrade path.

## What was still broken

Provenance is recorded as message indices in `CachedSourceEntry::fallback_timestamp_indices`, which Claude entries written by earlier releases leave empty. So on the first warm scan after an upgrade, a legacy entry reports an empty retained set and every one of its rows reads as live. The cross-file merge then takes the `else` branch, the path-first copy stays authoritative, and `merge_message_completeness` only pulls per-field maxima across — so the stale copy's model attribution survives, and the cost priced from it survives with it. Tokens are fine (they are maxima, order-independent); everything else is not. Every existing install upgrades with a populated cache, so this is the first warm scan for everyone.

Measured on the new regression test — original transcript keeps turn one, fork replays turn two with the completed response, cache entries downgraded to the pre-provenance shape:

```
left  (warm): [("msg_001:req_001", "claude-3-5-sonnet", 100, 50, "0.0010500000"), ("msg_002:req_002", "claude-3-5-haiku",  2000, 999, "0.0055960000")]
right (cold): [("msg_001:req_001", "claude-3-5-sonnet", 100, 50, "0.0010500000"), ("msg_002:req_002", "claude-3-5-sonnet", 2000, 999, "0.0209850000")]
```

The replayed turn reports $0.005596 instead of $0.020985 — a 3.75x understatement on a submitted number, from a scan that never re-reads the file.

## The fix

A legacy entry and a current entry that simply retained nothing both leave the index vector empty, so the vector alone cannot tell them apart — and getting that wrong either strands the stale rows or re-parses every Claude transcript on every scan forever. `CLAUDE_RETENTION_PROVENANCE_MARKER` (`usize::MAX`, never a real message index) is appended whenever provenance is written, so the two cases separate.

An entry without the marker skips the warm-hit shortcut and takes the ordinary re-parse path exactly once. That path already does the right thing: `retain_observed_messages` diffs the cached messages against the live bytes, which *is* the retained set, and the entry is written back through `new_with_retained_message_keys` with the marker. Every scan after that is a plain warm hit.

The marker lives inside the existing vector rather than in a new struct field on purpose. `CachedSourceEntry` is bincode-encoded without field names, so a new field means a `CACHE_FORMAT_VERSION` bump — and per the warning on `parser_version(ClientId::Claude)` that discards every Claude entry, including the compacted assistant turns the cache is the only copy of. This migration rebuilds instead of discarding, so #994's retained history survives the upgrade.

## Evidence

- `test_claude_legacy_cache_entry_rebuilds_retention_provenance` (lib.rs) — writes a pre-provenance entry holding a retained turn, warm-scans, and asserts the full result (dedup key, model, tokens, cost to 10 decimals) equals a cold parse of the same bytes. Verified red before the fix with the output above, green after.
- One-time: the same test asserts the rebuild counter advances once per entry on the first warm scan and does not move on the second, and that both persisted entries report `needs_retention_provenance_migration() == false` afterwards with `msg_002:req_002` as the only retained key.
- `test_claude_entry_reports_retention_provenance_only_once_recorded` (message_cache.rs) — pins the marker semantics, including that a current entry with an empty retained set is not mistaken for a legacy one, and that Codex entries (which use the same vector for real fallback-timestamp indices) are never rebuild candidates.
- `test_claude_warm_cache_removes_synthetic_placeholder_before_submit_validation` now seeds its entry through `new_with_retained_message_keys` so it keeps exercising the warm-hit placeholder repair rather than the rebuild path. Assertions unchanged.

## Gates

```
cargo fmt --all -- --check   clean
cargo clippy --workspace --all-targets   clean
cargo test -p tokscale-core   1516 passed, 0 failed, 1 ignored (lib) + all integration suites green
```

The upgrade cost is one re-parse per Claude transcript, once — the same work a cold Claude scan does, but without losing the retained turns a version bump would have thrown away.

PR #1089, opened 2026-08-10
fix(workspace): name workspace rows by repo instead of a truncated path

IngTian:
## Problem

Workspace rows are labeled with the raw workspace key. For Claude Code that key is the launch directory with every non-alphanumeric byte replaced by `-` (`-Users-me-devpro-ing-app`), so it carries no `/` for `workspace_label_from_key` to split on and the label becomes the entire mangled path. Because every row then shares a long identical prefix and truncation cuts from the right, `--group-by workspace,model` renders rows that cannot be told apart — the Models table clipped labels to 18 cells and the Overview list to 35, which on a real machine collapsed 36 of 44 project directories to the same visible text.

## Approach

The slugification maps `/`, `.`, `+` and `-` all onto `-`, so it cannot be inverted by string surgery — a greedy split recovered only 18 of 45 real directories in testing. Instead, re-applying that same forward map to real directory names while walking the filesystem identifies the source path exactly, with backtracking for the ambiguous cases (a directory whose own name contains `-`, or `.`). On 46 real project directories, all 30 that still had recorded `cwd` data recovered their exact path, including names like `IngTian.github.io` and `feat+terrain-functions`.

Rows are now labeled `repo` or `repo ⑃ worktree`. Two fallbacks keep the edge cases honest: a key that resolves to no directory (a worktree since deleted from disk) still splits on the encoded `--claude-worktrees-` marker rather than reverting to the mangled string, and a parser-supplied label always wins so non-path keys such as Warp's workspace UUID keep the name only their parser knows.

## `--merge-worktrees`

Added `--merge-worktrees` (and `w` in the TUI, shown in the footer as `[w:worktrees]` / `[w:repos]`) to fold git worktrees into their parent repository. Because the rollup canonicalizes through the same decode, it also merges a repository that Claude Code and Codex/OpenCode recorded under different key formats — previously one repo could occupy a dozen rows that no flag could combine, since a dash-mangled slug and a real POSIX path never compared equal.

Workspace bucketing moved into `tokscale_core::workspace_bucket`, shared with the TUI's own aggregation so the two cannot drift on labeling or rollup behavior, and is skipped entirely for groupings that discard the label.

## Layout

The Workspace column stays a fixed `Length`, widening from 18 to 44 cells only at inner width >= 193. `Min` outranks `Length` in ratatui's solver, so a flexible workspace column pays for itself out of its neighbors on any row that cannot satisfy every column — an earlier revision of this branch clipped Cost by a cell at 18 distinct widths and squeezed Model from 20 cells to 4, which merely relocated the unreadable-truncation bug.

Labels are truncated to the width the solver actually *grants*, not the width the column requested: a `Length` is a request, and between 174 and 192 cells the row could not satisfy 44 while Model held its `Min(20)`, so the formatter and the table disagreed and the overflow was clipped with no ellipsis. The layout is solved once per frame and each cell truncates to the width its own column was actually granted, so correctness no longer depends on the threshold being exactly right. The Model cell had the same bug from the other direction — a constant 24-cell budget against a floating `Min(20)` the solver sets to 20 on a narrow row and 77 on a wide one, over-truncating at 97 of the widths in 78..=400. Model and Overview cells truncate by terminal cells rather than code points, since a path or model id can contain full-width graphemes.

## Verification

`cargo test --workspace --all-features` passes with 2,911 tests and 0 failures on this branch's own code; `cargo clippy --locked --workspace --all-features -- -D warnings` and `cargo fmt --check` are clean.

Checked against real usage data: workspace rows drop from 34 to 18 under `--merge-worktrees`, total cost is conserved to the cent between the two modes, and no row still renders a raw mangled slug.

`--merge-worktrees` is threaded into `TuiConfig`, so an interactive launch opens merged instead of silently dropping the flag.

New tests cover the decoder (names containing `-`, `.` and `+`; non-ASCII names; symlinked directories; POSIX and Windows drive roots), the deleted-worktree fallback, the non-path-key case, the cross-client rollup merge, and the painted Models screen. Two layout tests sweep every width in 78..=400 rather than sampling — one asserting the money columns are never narrowed, one asserting the truncation width equals the allocated width. A six-point sample passed while 18 widths in the same range were clipping Cost, so sampling is not enough here.

## Known limitations

- **Roughly half the session parsers (24 of 46 — including gemini, cursor, amp, droid, roocode, kilocode, goose, and Copilot's OTEL path) never record a workspace at all**, so their usage still aggregates into the single `Unknown workspace` row. Out of scope here; documented in the README.
- **Decoding reads the filesystem.** It is memoized per workspace key, so the cost is proportional to workspaces (~40 on my machine) rather than messages (~100k), and it is skipped for every grouping that does not display a workspace. Measured: `--group-by workspace,model` takes 3.65s against 2.60s for the default grouping on the same data. There is deliberately no directory-listing cache — the shared `/Users/<me>` prefix is re-enumerated per key, which is a small fraction of a report dominated by parsing session files.
- **Rollup is lexical, not git-aware.** `workspace_repo_root` looks for `.claude/worktrees/` or `.git/worktrees/` as a path prefix, so `--merge-worktrees` folds worktrees kept inside the repository (what agent CLIs create) and nothing else. A worktree checked out beside the repo (`git worktree add ../feature-x`) is linked to it only by a `gitdir:` pointer file, so it stays its own row; a repo reached through two different path spellings (a symlink and its target) likewise stays two rows, since identities are compared as strings — that part is pre-existing, `Separate` mode on `main` already keys on the raw string. Totals are unaffected either way: usage is split across rows, never lost or double counted. Closing these needs `gitdir:` parsing plus `canonicalize` on both sides, and `canonicalize` fails for the deleted worktrees this PR deliberately still labels, so it belongs in a follow-up with its own cross-platform tests.
- **Two distinct directories with the same basename** (`/tmp/a/proj` and `/tmp/b/proj`) display the same label. They remain separate rows with separate keys, so no cost is merged; only the display string is ambiguous.
- **The footer help row** needs ~14 more columns before the trailing `[r:refresh] • e • q` fits once the `[w:...]` chip is present. That row already truncates below 110 columns in the default grouping; this widens the window rather than creating it.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -310,7 +310,7 @@ TUIで`g`を押すか、`--light`/`--json`モードで`--group-by`を使用し�
 | **モデル** | `--group-by model` | ✅ | モデルごとに1行 — すべてのクライアントとプロバイダーを統合 |
 | **クライアント + モデル** | `--group-by client,model` | | クライアント-モデルペアごとに1行 |
 | **クライアント + プロバイダー + モデル** | `--group-by client,provider,model` | | 最も詳細 — 統合なし |
-| **ワークスペース + モデル** | `--group-by workspace,model` | | ローカル使用量をワークスペースキー、次にモデルでグループ化 |
+| **ワークスペース + モデル** | `--group-by workspace,model` | | ローカル使用量をワークスペースキー、次にモデルでグループ化。[`--merge-worktrees`](README.md#per-workspace-cost) を追加すると、リポジトリ内のワークツリーを親リポジトリに畳み込みます |
 | **セッション + モデル** | `--group-by session,model` | | `session_id` とモデルごとに1行 — 特定のエージェント CLI セッションにコストを帰属 |
 | **クライアント + セッション + モデル** | `--group-by client,session,model` | | クライアント・セッション・モデルごとに1行 — `session_id` で結合するマルチエージェントランナーに便利 |
 
--- a/README.ko.md
+++ b/README.ko.md
@@ -306,7 +306,7 @@ TUI에서 `g`를 누르거나 `--light`/`--json` 모드에서 `--group-by`를 �
 | **모델** | `--group-by model` | ✅ | 모델당 한 행 — 모든 클라이언트와 프로바이더 병합 |
 | **클라이언트 + 모델** | `--group-by client,model` | | 클라이언트-모델 쌍당 한 행 |
 | **클라이언트 + 프로바이더 + 모델** | `--group-by client,provider,model` | | 가장 세분화 — 병합 없음 |
-| **워크스페이스 + 모델** | `--group-by workspace,model` | | 로컬 사용량을 워크스페이스 키별로, 그 다음 모델별로 그룹화 |
+| **워크스페이스 + 모델** | `--group-by workspace,model` | | 로컬 사용량을 워크스페이스 키별로, 그 다음 모델별로 그룹화. [`--merge-worktrees`](README.md#per-workspace-cost)를 추가하면 리포지터리 내부의 워크트리를 상위 리포지터리로 병합합니다 |
 | **세션 + 모델** | `--group-by session,model` | | `session_id`와 모델당 한 행 — 특정 에이전트-CLI 세션에 비용 귀속 |
 | **클라이언트 + 세션 + 모델** | `--group-by client,session,model` | | 클라이언트, 세션, 모델당 한 행 — `session_id`로 조인하는 멀티 에이전트 러너에 유용 |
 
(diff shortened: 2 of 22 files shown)


junhoyeo:
Thanks — this shipped.

Your eight commits went in via #1122, which merged into `main` as part of #1124 (`cf622874`, `Co-authored-by: Zeying Tian`). #1122 itself shows as closed rather than merged because #1124 branched from its head and carried it forward, so the trail is a little confusing from the PR list — but the work is in.

The follow-up dealt with the rendered-width problem your change surfaced: labels that were unique as strings could still render identically once the Workspace column truncated them to 18 cells.

Closing as merged-via-#1124.

Issue #1031, opened 2026-08-04
Grok Build usage parsed as 0 for `tokscale models -c grok --today/--month` despite full updates.jsonl


QYworld:
## Summary

Grok Build sessions are found (223 messages), but token usage is reported as 0 for both `--today` and `--month` even though the session files contain complete per-turn usage in `updates.jsonl`.

## Environment

- tokscale CLI: 4.9.0 (bundled inside Token Monitor 0.40.0, macOS arm64)
- macOS: 26.5.2
- Grok Build invoked headless (yolo mode) from an agent daemon (not interactive terminal)

## Steps to reproduce

1. Have Grok Build write sessions under `~/.grok/sessions/<urlencoded-workspace>/<session-id>/updates.jsonl` (workspaces include dot-prefixed project directories).
2. Run:
   - `tokscale clients` → shows `Grok Build  sessions: ~/.grok/sessions ✓  messages: 223`
   - `tokscale models -c grok --today` → `Total: 0 messages, 0 tokens, $0.00`
   - `tokscale models -c grok --month` → `Total: 0 messages, 0 tokens, $0.00`

## Expected

`models -c grok` should report the tokens accumulated in those sessions (same order of magnitude as the usage recorded in `updates.jsonl`).

## Actual

0 messages / 0 tokens / $0.00, for both today and the full current month.

## Root-cause hypothesis (from source inspection)

`crates/tokscale-core/src/sessions/grok.rs` `GrokUsage::from_update` reads usage via `get_path(value, &["params", "update", "usage"])` with camelCase keys (`inputTokens`, `outputTokens`, `cachedReadTokens`, `costUsdTicks`, etc.). The on-disk data does contain exactly that shape, so the parser path looks right — yet the aggregate is empty.

One candidate mismatch I observed: top-level timestamp is written as a bare **seconds** integer (`"timestamp": <10-digit unix seconds>`), while several parser/test paths expect an `agentTimestampMs` field or ms-based values. If the timestamp path resolves to something outside the queried day/month bucket, messages would be silently excluded from `--today` / `--month` aggregates.

## Sanitized sample of a real usage line (values redacted)

```json
{
  "method": "_x.ai/session/update",
  "params": {
    "update": {
      "sessionUpdate": "turn_completed",
      "usage": {
        "inputTokens": "<int>",
        "outputTokens": "<int>",
        "totalTokens": "<int>",
        "cachedReadTokens": "<int>",
        "reasoningTokens": "<int>",
[...]

PR #1090, opened 2026-08-10
fix(droid): attribute session usage to the replies that spent it

Jackson57279:
## Problem

A Droid `*.settings.json` holds one cumulative `tokenUsage` total for the whole session, rewritten in place every time the agent spends tokens. `parse_droid_file` emitted a single record for it, anchored on `providerLockTimestamp` — which records when the provider was *selected*, not when the tokens were spent, and which Droid never updates as the totals climb.

For a session that stays open across days (a `/loop`, a long autonomous run), every token it ever spends is therefore attributed to the instant it started. The session reads as completely silent in `--today` and `--yesterday` while it is actively burning tokens:

```
$ tokscale --no-spinner --light --client droid --today
  Total: 0 messages, 0 tokens, $0.00
```

That machine had three Droid sessions mid-run. Here is the shape of the data — `providerLockTimestamp` frozen at session start while the totals kept being rewritten for three more days:

| `providerLockTimestamp` | file mtime | total tokens |
| --- | --- | --- |
| 2026-08-07T03:32:46Z | 2026-08-10T06:28:01Z | 203,852,824 |
| 2026-08-07T05:57:28Z | 2026-08-10T06:24:57Z | 155,503,988 |
| 2026-08-07T03:20:36Z | 2026-08-10T06:29:34Z | 149,761,299 |

Moving the anchor to the file's mtime makes those sessions visible again, but it does not make them *correct*: whichever single instant is chosen, one day absorbs every token the session ever spent. Checked against a provider-side API dashboard, anchoring at mtime reported **517,527,976** tokens for a day that had actually spent **24,226,571** — a 21× overstatement in place of the previous zero. A single cumulative blob stamped at one instant simply cannot describe multi-day spend.

## Fix

Droid records no token counts anywhere in the transcript, so there is nothing to read per call. But the transcript does say *when* each assistant reply happened and how much conversation was live at the time, and cost per call tracks that closely: nearly every token in these sessions is a cache read of the conversation so far. Weighting each assistant reply by the bytes accumulated since the last compaction recovers what one reply cost relative to another, which is all that apportioning the recorded total needs. Compaction resets the running weight, because it discards the transcript the following calls would otherwise re-read.

Each reply is weighted by the context standing *before* it, not including its own line. A reply's own bytes are output; the call's input and cache-read cost was already fixed by the conversation it read. Folding them in would let a verbose answer claim a larger share at the expense of other replies — including replies on other days, which makes it an attribution error rather than a rounding one.

The session's cumulative total is split across those replies and emitted as one record each. A multi-day session now reports against the days it actually ran, and hourly reports become meaningful for a client that previously had one timestamp per session.

Splitting allocates on the running total rather than rounding each share independently, so the parts sum back to exactly the total Droid recorded — daily figures cannot drift from the cumulative number.

Sessions whose transcript is missing, unreadable, or has no assistant reply yet keep the previous single-record behavior, anchored at mtime and floored at `providerLockTimestamp`, so usage is never dropped for want of a transcript. That fallback also fixes two smaller issues in the original anchoring: mtime beats a lock timestamp that Droid never updates, and a pre-epoch lock is now rejected rather than surviving into a 1969 bucket no date filter can reach.

**Cross-platform:** the fallback anchor relies only on `SystemTime::modified`, which every tier-1 target reports — unlike `created`, which is absent on many Linux filesystems. The timestamp and apportioning logic are pure functions, so their tests assert real ordering and conservation without depending on filesystem timestamp granularity.

## Verification

Measured against the same provider-side dashboard that exposed the problem:

| window | dashboard | before (lock anchor) | mtime anchor | this PR |
| --- | --- | --- | --- | --- |
| one day | 24,226,571 | 0 | 517,527,976 (21× over) | **24,547,118 (1.3% off)** |

Over a multi-day window the remaining gap is Droid's own under-recording, not attribution: 41 of 185 turns in that data ended in `"reason":"error"`, and an errored turn costs provider-side tokens that never reach `tokenUsage`. No parser of these files can see them.

Conservation holds against real data: summing every `tokenUsage` field across every `settings.json` on that machine gives 673,425,644, and `tokscale --client droid` reports exactly 673,425,644. The split moves usage between buckets without creating or losing any.

`cargo fmt --all -- --check`, `cargo clippy --locked --workspace --all-features -- -D warnings`, and `cargo test --workspace --all-features` are all clean.

The parser's test count goes from 6 to 20, covering apportioning conservation (parts sum to the original total), zero-weight handling, spreading across days, the compaction reset, a long reply not charging itself for its own output, both fallback paths, and the unusable-lock rules. I confirmed each regression test is not vacuous by restoring the old behavior and watching it fail — including one that initially passed under both and had to be rewritten around the ratio between shares rather than their order, since charging a reply for its own output leaves the two nearly equal (`[499, 501]`) rather than reversed.

### Scan cost

Reading the transcript is new work for this parser, so I measured it rather than assuming. Isolating parse cost with CPU time (wall clock on a loaded machine is dominated by fixed startup work), against a synthesized 197MB transcript holding 20,000 assistant replies:

| | CPU | peak RSS |
| --- | --- | --- |
| no Droid data (fixed overhead) | 0.22–0.24s | ~46MB |
| 197MB transcript, 20,000 replies | 0.64–0.67s | ~69MB |
| **attributable to parsing** | **~0.42s** | **~23MB** |

That is roughly 470MB/s, linear in transcript bytes, and the totals still come back exact. The ~23MB is the emitted records themselves (~1.2KB each) — the output the report consumes — not parser scratch space; a transcript is streamed a line at a time and nothing but the per-reply timestamp and weight is retained while scanning. Reaching even 1GB of resident records would take on the order of 800k replies, several gigabytes of transcripts. Cost is in line with the other JSONL parsers here; Grok's handles an 83MB `updates.jsonl` by design.

On the real corpus (20MB of transcripts) a cold scan is not measurably slower than `main`.

I deliberately did not add a size or turn cap that falls back to the single-record path. The cap would engage on exactly the largest, longest-running sessions — the ones this PR exists to fix — and silently return them to the 21× overcount, trading a bounded, measured cost for an unbounded correctness error with no signal to the user.

I also tried replacing the per-line `serde_json::Value` with a typed struct so reply bodies are skipped rather than materialized, and reverted it: peak RSS was unchanged (the memory is the emitted records, not the parse), the CPU difference was inside measurement noise, and it made the parser reject a whole line on any unexpected field shape instead of stepping over it — the opposite of the defensive parsing this codebase asks for, and the failure mode behind #1031.

## Cache invalidation

Bumps the Droid parser version to 4. This matters here for the same reason it did for Grok in #1031: a Droid session that has since ended never changes its bytes again, so its fingerprint keeps matching forever and warm caches would otherwise serve the old single-record shape indefinitely. Versions 2 and 3 only ever existed in pre-release builds of this change, and the bump past them keeps anyone who ran one from holding a superseded weighting. Droid is not covered by `retained_history_key_filter`, so the bump is merely cold, not lossy.

## Note on scope

The per-reply split is an estimate, not metering — Droid does not record per-call usage, so no parser can be exact here. It is reproducible from the files alone, which is why it is preferred over persisting cross-run snapshots of the cumulative counter: that would put history in the cache where re-parsing could not rebuild it, and make any future parser-version bump permanently lossy in the way `message_cache.rs` warns about for Claude.

## Disclosure

This patch is vibecoded — investigated, written, and tested by **Claude Opus 5** running as an agent in **Claude Code**, and reviewed by me before submitting. Happy to adjust anything that does not match how you would rather see this handled.

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1068,7 +1068,11 @@ fn parser_version(client: ClientId) -> u32 {
         // token it ever spent against the day it was started. A session that
         // has since ended never changes its bytes again, so its fingerprint
         // stays valid forever and only this bump discards the v1 anchor.
-        ClientId::Droid => 2,
+        // v2->v3: a session's cumulative total is no longer one record at one
+        // instant. It is now apportioned across the assistant replies in the
+        // sibling transcript, weighted by the context each reply read, so a
+        // multi-day session reports against the days it actually ran.
+        ClientId::Droid => 3,
         _ => 1,
     }
 }
@@ -2601,7 +2605,7 @@ mod tests {
         // A finished Droid session's settings.json is never rewritten again, so
         // its fingerprint keeps matching and only the version bump discards the
         // v1 lock-timestamp anchor.
-        assert_eq!(parser_version(ClientId::Droid), 2);
+        assert_eq!(parser_version(ClientId::Droid), 3);
     }
 
     #[test]
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1072,7 +1072,12 @@ fn parser_version(client: ClientId) -> u32 {
         // instant. It is now apportioned across the assistant replies in the
         // sibling transcript, weighted by the context each reply read, so a
         // multi-day session reports against the days it actually ran.
-        ClientId::Droid => 3,
+        // v3->v4: a reply is weighted by the context standing before it rather
+        // than including its own bytes, so a long answer no longer charges
+        // itself for its own output. Versions 2 and 3 only ever existed in
+        // pre-release builds of this change; the bump past them keeps anyone
+        // who ran one from holding the superseded weighting.
+        ClientId::Droid => 4,
         _ => 1,
     }
 }
@@ -2605,7 +2610,7 @@ mod tests {
         // A finished Droid session's settings.json is never rewritten again, so
         // its fingerprint keeps matching and only the version bump discards the
         // v1 lock-timestamp anchor.
-        assert_eq!(parser_version(ClientId::Droid), 3);
+        assert_eq!(parser_version(ClientId::Droid), 4);
     }
 
     #[test]
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1062,6 +1062,13 @@ fn parser_version(client: ClientId) -> u32 {
         // v2->v3: duplicate merging now upgrades the retained row when a later
         // copy carries an explicit cost, including zero.
         ClientId::MiMoCode => 3,
+        // Droid's cumulative session totals now anchor on the settings file's
+        // mtime (floored at providerLockTimestamp) instead of the lock
+        // timestamp alone, so a long-running session stops reporting every
+        // token it ever spent against the day it was started. A session that
+        // has since ended never changes its bytes again, so its fingerprint
+        // stays valid forever and only this bump discards the v1 anchor.
+        ClientId::Droid => 2,
         _ => 1,
     }
 }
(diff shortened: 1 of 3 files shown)


junhoyeo:
Thanks for this — the analysis here was the valuable part, and it shipped.

This landed in #1120, which carries your two commits with authorship intact (`Co-authored-by: Jackson57279` is on `151ecff2` in `main`), plus the corrections to the apportioning and the server-side bound the change needed before it could ship.

Closing as merged-via-#1120 rather than stale. The `providerLockTimestamp` finding — that Droid never updates it as the totals climb, so a multi-day session reads as silent — is what made the fix possible.

PR #1102, opened 2026-08-11
fix(claude): discover sessions under CLAUDE_CONFIG_DIR instead of a fixed ~/.claude root

### LINKED ISSUE #1101: Claude usage discovery ignores CLAUDE_CONFIG_DIR, unlike Codex's CODEX_HOME
## Summary

Claude Code's own CLI honors `CLAUDE_CONFIG_DIR` to relocate its config/session directory away from `~/.claude`, but tokscale's `claude` client registry hardcodes `PathRoot::Home` + `.claude/projects` and never reads that variable. Anyone who has relocated their Claude config gets zero Claude usage reported, silently.

This was found while investigating a downstream report of Claude usage under-reported by ~52% in [Javis603/token-monitor#375](https://github.com/Javis603/token-monitor/issues/375), which spawns `tokscale --json` and delegates all token counting/root discovery to it.

## Where this lives

`crates/tokscale-core/src/clients.rs` registers `codex` with `PathRoot::EnvVar { var: "CODEX_HOME", fallback_relative: ".codex" }`, but `claude` is registered as plain `PathRoot::Home` with `relative: ".claude/projects"`:

```rust
Claude = 1 => {
    id: "claude",
    display: "Claude Code",
    logo: Some("https://tokscale.ai/assets/logos/claude.jpg"),root: PathRoot::Home,
    relative: ".claude/projects",
    pattern: "*.jsonl",
    headless: false,
    parse_local: true,
    submit_default: true
},
```

Grepping the whole Rust workspace for `CLAUDE_CONFIG_DIR` returns zero matches — it is not handled anywhere.

There's a second, independent hardcoded reference in `crates/tokscale-core/src/scanner.rs`'s `built_in_extra_scan_paths_for`, which scans `<home>/.claude/transcripts` as a sibling extra path, again relative to `home_dir` directly rather than through the client's resolved root.

## Impact

Any user who sets `CLAUDE_CONFIG_DIR` (documented, first-class Claude Code behavior) gets **zero** Claude Code sessions scanned — not partial underreporting, complete silence, since `tokscale clients`/`tokscale --client claude` looks in the wrong directory entirely.

## Proposed fix (PR attached)

Register `claude` the same way `codex` already is:

```rust
Claude = 1 => {
    id: "claude",
    display: "Claude Code",
    logo: Some("https://tokscale.ai/assets/logos/claude.jpg"),root: PathRoot::EnvVar {
        var: "CLAUDE_CONFIG_DIR",
        fallback_relative: ".claude",
    },
    relative: "projects",
    pattern: "*.jsonl",
    headless: false,
    parse_local: true,
    submit_default: true
},
```

...and resolve the `transcripts` extra scan path through that same registry root instead of a second `home_dir`-relative literal, so both move together. I've opened a PR with this change, new regression tests for both the registry root and the scanner's extra-path resolution, and an accompanying fix to `claude_diagnostics.rs` (which independently hardcoded both paths for the `clients` diagnostics output).

## A second, unrelated finding — not fixed in the attached PR

While investigating the downstream report above, the reporter also flagged that on their machine ~78% of Claude token volume lives in the Claude desktop app's agent-mode session root (`%LOCALAPPDATA%\Claude-3p\local-agent-mode-sessions` on Windows; analogous paths under `~/Library/Application Support/Claude-3p/...` on macOS and `${XDG_DATA_HOME}/Claude-3p/...` on Linux), which tokscale does not scan at all — there is no reference to `Claude-3p` or `local-agent-mode-sessions` anywhere in this repo. They report the JSONL schema there matches Claude Code's own transcript format. This is a separate, larger change (a new source root, possibly a new scan entry point) that I have not attempted here since I can't validate the schema claim against real data — flagging it in case a maintainer or another contributor wants to pick it up. Happy to help validate against a sanitized sample if someone can share one.

## Environment

- tokscale version inspected: v4.13.0 (current `main` as of this report, commit `246765b`)
- Confirmed via direct inspection of `clients.rs`/`scanner.rs`, not user reproduction on this machine

Excelius-Wang:
## Summary

Fixes #1101. Claude Code's own CLI honors `CLAUDE_CONFIG_DIR` to relocate its config directory, but the `claude` client here was registered as a fixed `PathRoot::Home` + `.claude/projects`, so relocating via that variable silently drops all Claude usage. `codex` already has this exact pattern for `CODEX_HOME` via `PathRoot::EnvVar`; this gives `claude` the same treatment.

- `crates/tokscale-core/src/clients.rs`: registers `claude`'s root as `PathRoot::EnvVar { var: "CLAUDE_CONFIG_DIR", fallback_relative: ".claude" }` with `relative: "projects"`, matching the existing `codex` pattern exactly. Unset behavior is unchanged (`~/.claude/projects`).
- `crates/tokscale-core/src/scanner.rs`: `built_in_extra_scan_paths_for`'s `.claude/transcripts` extra path was a second, independent `home_dir`-relative literal — resolved it through the same client registry root instead so it moves together with `CLAUDE_CONFIG_DIR`, and threaded the existing `use_env_roots` flag through to it (mirroring how `Grok`'s root is resolved just above the same call site).
- `crates/tokscale-cli/src/claude_diagnostics.rs`: the `clients` diagnostics output independently hardcoded both `.claude/projects` and `.claude/transcripts` for its existence checks; switched to the same resolved root so diagnostics don't disagree with what's actually scanned.

## Test plan

- [x] New tests: `test_claude_root_uses_claude_config_dir_env_var`-equivalent coverage — `test_claude_defaults_to_home_dot_claude_without_env_override` and `test_claude_honors_claude_config_dir_env_override` in `clients.rs`, mirroring the existing Codex env-var tests
- [x] New test: `test_scan_all_clients_claude_honors_claude_config_dir` in `scanner.rs`, verifying both the `projects` and `transcripts` paths are discovered under a custom `CLAUDE_CONFIG_DIR` and a stray `~/.claude` entry is ignored
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --locked --workspace --all-features -- -D warnings`
- [x] `cargo test --workspace --all-features` — all passing except one pre-existing, unrelated failure (`test_submit_excluding_only_generic_gemini_usage_does_not_promise_submission`) that fails identically on unmodified `main` in this sandbox because it requires a live fetch to `models.dev` that the sandbox's network policy blocks

---
## Summary by cubic
Use `CLAUDE_CONFIG_DIR` to discover Claude Code sessions instead of a fixed `~/.claude` root, and keep `projects` and `transcripts` in sync with it. Diagnostics now follow the same env strategy, so `--home` overrides ignore `CLAUDE_CONFIG_DIR`; default behavior is unchanged when the env var is unset.

- **Bug Fixes**
  - Resolve Claude’s root via `CLAUDE_CONFIG_DIR` (fallback to `~/.claude`) and set `relative: "projects"`.
  - Scanner resolves the `transcripts` path from the same root and respects `use_env_roots`.
  - Diagnostics resolve through the same root and respect `use_env_roots`, so reported paths match the scanner and `--home` overrides don’t follow `CLAUDE_CONFIG_DIR`.

<sup>Written for commit cf5256b87f06829cc7174bfb499b38d9a90fc0d8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1102?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/claude_diagnostics.rs
+++ b/crates/tokscale-cli/src/claude_diagnostics.rs
@@ -103,15 +103,18 @@ fn diagnostic_paths(home_dir: &Path, desktop_paths: Vec<PathBuf>) -> Vec<Diagnos
         })
         .collect();
 
+    // Resolve through the client registry (CLAUDE_CONFIG_DIR-aware) instead of
+    // a fixed `.claude` offset from home, so this diagnostic reports the same
+    // paths the scanner actually reads.
+    let claude_root = PathBuf::from(
+        tokscale_core::ClientId::Claude
+            .data()
+            .root
+            .resolve(&home_dir.to_string_lossy()),
+    );
     for (label, path) in [
-        (
-            "claudeCodeProjects",
-            home_dir.join(".claude").join("projects"),
-        ),
-        (
-            "claudeCodeTranscripts",
-            home_dir.join(".claude").join("transcripts"),
-        ),
+        ("claudeCodeProjects", claude_root.join("projects")),
+        ("claudeCodeTranscripts", claude_root.join("transcripts")),
     ] {
         let exists = path.exists();
         paths.push(DiagnosticPath {
--- a/crates/tokscale-cli/src/claude_diagnostics.rs
+++ b/crates/tokscale-cli/src/claude_diagnostics.rs
@@ -24,12 +24,13 @@ const CLAUDE_DESKTOP_MESSAGE: &str =
 
 const CLAUDE_DESKTOP_HELP: &str = "Claude Desktop chat storage and Claude data exports do not expose a documented per-message token ledger. Use `tokscale usage` for Claude subscription quota bars; organization/API billing requires Anthropic Admin Usage/Cost API outside local scanning.";
 
-pub fn diagnostics_for_clients_row(home_dir: &Path) -> Vec<ClientDiagnostic> {
-    claude_diagnostics(home_dir, true)
+pub fn diagnostics_for_clients_row(home_dir: &Path, use_env_roots: bool) -> Vec<ClientDiagnostic> {
+    claude_diagnostics(home_dir, use_env_roots, true)
 }
 
 pub fn diagnostics_for_empty_explicit_report(
     home_dir: &Path,
+    use_env_roots: bool,
     clients: &Option<Vec<String>>,
     claude_message_count: i32,
 ) -> Vec<ClientDiagnostic> {
@@ -37,7 +38,7 @@ pub fn diagnostics_for_empty_explicit_report(
         return Vec::new();
     }
 
-    claude_diagnostics(home_dir, false)
+    claude_diagnostics(home_dir, use_env_roots, false)
 }
 
 fn explicitly_requests_claude(clients: &Option<Vec<String>>) -> bool {
@@ -46,7 +47,11 @@ fn explicitly_requests_claude(clients: &Option<Vec<String>>) -> bool {
         .is_some_and(|ids| ids.iter().any(|id| id == "claude"))
 }
 
-fn claude_diagnostics(home_dir: &Path, include_info: bool) -> Vec<ClientDiagnostic> {
+fn claude_diagnostics(
+    home_dir: &Path,
+    use_env_roots: bool,
+    include_info: bool,
+) -> Vec<ClientDiagnostic> {
     let mut diagnostics = Vec::new();
 
     let desktop_paths: Vec<PathBuf> = claude_desktop_storage_paths(home_dir)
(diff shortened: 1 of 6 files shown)


junhoyeo (approved):
Thanks!

Excelius-Wang:
Good catch — fixed in cf5256b8: threaded `use_env_roots` through `diagnostics_for_clients_row`/`diagnostics_for_empty_explicit_report` down to `resolve_with_env_strategy`, so a `--home` override now ignores `CLAUDE_CONFIG_DIR` in diagnostics the same way it already does in the scanner. Also added `CLAUDE_CONFIG_DIR` to the test env-hermeticity lists and a regression test (`test_clients_home_override_ignores_conflicting_claude_config_dir_env`).

PR #1105, opened 2026-08-12
feat(clients): add Cherry Studio support

sakuralmx:
### Summary

Adds [Cherry Studio](https://cherry-ai.com/) (an Electron desktop AI client) as a tracked client. Cherry Studio's Agent / Claude Code mode writes **standard Claude Code transcripts** under its per-user app-data directory, which this PR scans, parses, and submits like any other client.

### Key design points

- **Parser** (`sessions/cherrystudio.rs`): Cherry Studio appends the **same API call 3–4 times** to the transcript as the streaming response progresses (different `uuid`, identical `usage`). Naively summing every assistant row triple-counts each call. The parser dedupes **consecutive rows within a session whose usage signature** (`model` + the four token buckets) is identical, keeping one record per API call. Verified against the platform's per-hour billing with <1% error.
- **Dual transcript roots**: Cherry Studio V2 migrated active transcripts to `<config_dir>/CherryStudio/Data/Agents/.claude/projects`, while the legacy `CherryStudio/.claude/projects` root keeps untransferred history. Both are scanned; same-name sessions are deduped by path relative to the scanning root with the **V2 copy winning** (it receives new writes), V1-only files fill in the rest.
- **`PathRoot::AppData`**: new root variant resolving via `dirs::config_dir()` (`%APPDATA%` on Windows, `~/Library/Application Support` on macOS, XDG config home on Linux). Without env roots (hermetic tests, explicit `--home`) it now resolves under the given home like every other root, so an AppData-rooted client cannot leak the machine's real per-user data into a test scan.
- **Frontend**: registered in `SUPPORTED_CLIENT_TYPES` + display name / logo / color (Cherry Studio pink `#FA7298`), with the required `.github/assets/client-cherrystudio.png` logo.

### Files

- `crates/tokscale-core/src/clients.rs` — `ClientId::CherryStudio = 45`, `PathRoot::AppData`
- `crates/tokscale-core/src/sessions/cherrystudio.rs` — parser (new)
- `crates/tokscale-core/src/scanner.rs` — V1/V2 dual-root scan + relative-path dedup
- `crates/tokscale-core/src/lib.rs`, `sessions/mod.rs` — wiring
- `crates/tokscale-cli/src/main.rs`, `tui/client_ui.rs`, `tui/data/mod.rs` — CLI filter + hotkey `G`
- `packages/frontend/src/lib/types.ts`, `constants.ts`, `.github/assets/client-cherrystudio.png`
- `README.md` + ko/ja/zh-cn — supported-client list and data-location matrix

### Verification

- `cargo fmt --all -- --check` ✅
- `cargo clippy --locked --workspace --all-features` — 0 errors ✅
- `cargo test --workspace --all-features` — 1618 passed; the 2 failures on Windows are pre-existing platform issues (macOS-only dead code under `#[cfg(target_os = "macos")]`, and a cache-path test whose `HOME` redirect is ignored by Windows native home resolution — both pass on the macOS/Linux CI)
- `bun run --cwd packages/frontend test -- __tests__/lib/clientRegistry.test.ts` — 5/5 ✅
- `bun run --cwd packages/frontend typecheck` ✅
- Real-data run: `tokscale --no-spinner --client cherrystudio` reports ~30.5B tokens / ~31K messages from live Cherry Studio transcripts, matching the independently verified platform billing within 1%

---
## Summary by cubic
Adds Cherry Studio as a tracked client and fixes 3–4x overcounting of streamed responses. Old: summing each assistant replay row triple-counted calls and could move historical calls to “today” when a later replay lacked a timestamp. New: dedupe across stable IDs (`requestId` primary, `message.id` secondary, `uuid` fallback) after reading the full transcript, merge usage by per-bucket maxima, and preserve the latest valid event timestamp so timestamp-less replays do not reset call time.

- Core (`crates/tokscale-core`): adds `ClientId::CherryStudio` and a parser (`sessions/cherrystudio.rs`) that forms alias-connected components for the whole transcript via iterative union-find, keeps distinct requests even if a message ID is reused, retains sparse rows without IDs, merges streaming snapshots by per-field maximums, and selects metadata from the latest valid event timestamp (falling back to file mtime only when no event timestamp exists). Adds `PathRoot::AppData` with explicit `--home` mapping to platform app-data layouts.
- Scanner (`crates/tokscale-core`): scans both Cherry Studio roots, classifies V2 exactly, dedupes by relative path with V2 winning, and ignores ancestor-name false positives.
- CLI/TUI/tests (`crates/tokscale-cli`): supports `--client cherrystudio` and TUI hotkey G; tests verify alias-based dedupe and that historical timestamps survive timestampless replays across cold/warm cache.
- Frontend/docs (`packages/frontend` + READMEs): registers `cherrystudio` in `SUPPORTED_CLIENT_TYPES`, adds display name, logo, and color `#FA7298`; docs list V1/V2 paths and `--client cherrystudio`.

<sup>Written for commit 9055099a31d41edc9fec55f4f2519065ce36fbf5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1105?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sakuralmx:
Addressed the review feedback from the AI reviewer:

- **cherrystudio.rs: cache-only rows are no longer dropped** — removed the early `input == 0 && output == 0` skip that ran before the cache buckets were read; an all-zero check now runs only after all four buckets are summed, so a request with cached input but zero uncached input/output is counted.
- **cherrystudio.rs: saturating bucket sum** — the four-token-bucket total now uses `saturating_add` so malformed transcripts with extreme values cannot overflow (debug panic / release wrap).
- **cherrystudio.rs: delimited OpenAI `sol` matching** — `lower.contains("sol")` would misclassify models like `solar-pro` as OpenAI; the branch now requires the `sol` token at the end of the model id (`ends_with("sol")`), matching the actual `gpt-5.6-sol` naming.
- **READMEs: data-location rows now list both transcript roots** — the Cherry Studio rows in all four READMEs document the V2 root (`<appdata>/CherryStudio/Data/Agents/.claude/projects`, where current writes land) alongside the legacy V1 root, matching what the scanner resolves.

Verification: `cargo fmt --check`, `cargo clippy` (0 errors), cherrystudio unit tests pass, and a real-data run reports the expected totals.

junhoyeo (approved):
lgtm 👍

PR #1108, opened 2026-08-12
fix(cache): read source cache shards lazily and release consumed payloads

### LINKED ISSUE #1100: Memory leak
```
npx -y tokscale@4
```

tokscale 4.13.0

The memory keeps rising, and then the process exits

Image

Image

Return to v3 work is fine

```
npx -y tokscale@3
```

junhoyeo:
Fixes #1100.

## Diagnosis

`npx -y tokscale@4` is not leaking in the Rust sense — a tracking-allocator probe over four repeated in-process scans of a ~1M-message home shows the live heap returning to the same 6.2 MB baseline every time (`retained_delta=0.0MB`). What grows is the *peak* of each scan, and the TUI pays that peak again on every auto-refresh (the reporter's footer shows `R:auto 60s`).

The peak was proportional to the entire persisted source-message cache rather than to what the scan actually reads, because `SourceMessageCache::load` eagerly deserialized every client namespace and every shard, then kept every entry alive for the whole scan while the scan's own output accumulated on top of it. Measured against a real 358 MB on-disk cache, a scan that produces a *single* message peaked at 1164.7 MB; the same scan against an empty cache peaked at 12.9 MB. A live-heap timeline of a full scan shows 1215 MB allocated in the first 8 seconds (the cache load) and held flat as a floor for the remaining 76 seconds while the output grew to ~2.1 GB. That is why v3 is fine and v4 is not: v3 has no source-message cache, so it never paid this floor, and the floor grows for the life of the install as the cache accumulates.

## Change

- Read a namespace's shards the first time a lane asks for one of its sources instead of reading every namespace up front. Missing-source pruning moves into that per-namespace load, since nothing reads the whole cache any more.
- Hand a clean entry's message payload to its one consumer (`SourceMessageCache::take`) instead of cloning it out of a copy the cache keeps for the rest of the scan. A metadata husk stays behind so invalidation and shard rewrites still see the entry's path and fingerprint.
- Two cases keep their payload: an entry actually carrying retained history (a Claude entry with a non-empty retained set, or a pre-provenance entry that cannot say which of its rows are retained), where the cache is the only copy and a second lookup of a drained entry would retire those turns (#994); and entries already marked dirty, which `save_if_dirty` writes back from memory. A Claude entry whose retained set is empty is fully reproducible from the live file, so it drains like any other.

`save_if_dirty` already re-reads each shard it rewrites from disk and only overlays the entries a scan marked dirty, so a drained clean entry is never needed in memory again — that is what makes releasing it safe, and `saving_after_a_take_keeps_the_drained_entry_on_disk` pins it.

## Measured before/after

Peak heap per scan, tracking global allocator, same machine, same 368 MB cache snapshot restored before each run, real ~1M-message home:

| scan | before | after | change |
| --- | --- | --- | --- |
| `-c droid` (1 message) | 1164.7 MB | 6.3 MB | −99.5% |
| `-c codex` (334k messages) | 1477.0 MB | 328.3 MB | −77.8% |
| `-c claude` (254k messages) | 1839.6 MB | 757.2 MB | −58.8% |
| all clients (993k messages) | 2252.5 MB | 1473.2 MB | −34.6% |

Warm-scan wall clock over the same runs: `-c codex` 15.4 s → 1.9 s, `-c droid` 15.2 s → 0.0 s, `-c claude` 67.3 s → 38.1 s. A full all-clients scan is unchanged within run-to-run noise (85.6 s → 89.8 s against a 75.9–91.7 s spread across repeats), since it still reads every namespace.

End-to-end TUI (`tokscale --refresh 20`, 5.5 minutes, same cache snapshot), macOS `vmmap` physical footprint: peak 1843 MB → 1638 MB, and the between-refresh floor drops from ~1126 MB to ~300–900 MB.

The all-clients row is the smallest win because the TUI does read every namespace; the floor now decays as the scan's output grows instead of being held whole.

## Output equality

Byte-identical `tokscale --json -c claude` output between `main` and this branch, cold and warm, over an 800k-message synthetic Claude corpus (only `processingTimeMs` differs). A live-corpus probe that appends to five transcripts and compacts one between eight successive scans returns the identical message sequence (900/1800/…/7200) on both, so retention across compaction is unchanged.

## Tests

Seven new regression tests in `message_cache`:

- `load_reads_only_the_namespaces_a_scan_touches`
- `take_releases_the_entry_payload_but_keeps_its_metadata`
- `saving_after_a_take_keeps_the_drained_entry_on_disk`
- `take_keeps_the_payload_of_an_entry_carrying_retained_history`
- `take_keeps_the_payload_of_a_pre_provenance_claude_entry`
- `take_drains_a_claude_entry_with_no_retained_history`
- `remove_loads_the_namespace_it_invalidates`
- `loading_a_namespace_prunes_entries_whose_source_is_gone`
- `a_lazy_namespace_load_does_not_overwrite_this_scan_s_entries`

## Validation

- `cargo fmt --all -- --check` — exit 0
- `cargo clippy --locked --workspace --all-targets --all-features -- -D warnings` — exit 0
- `cargo test --workspace` — 2871 passed, 0 failed

## Known limitations

- Stale entries in a namespace this machine never scans again are no longer pruned, because nothing reads that namespace. They cost disk, not memory.
- A source looked up twice in one scan sees a drained husk and re-parses it: output is unchanged, the scan is slower. Retention namespaces are exempt, so this can never lose history.
- This does not make the TUI's refresh incremental. A full all-clients refresh still materializes every message, so a large enough history can still spike; the reporter's own cache was not available to verify the absolute numbers on their machine.

---
## Summary by cubic
Read source-message cache shards lazily per client and hand cached payloads to parsers instead of cloning them. Fixes #1100 memory spikes by only deserializing what a scan needs and not keeping a second in-memory copy; Claude entries now drain unless they actually carry retained history.

- Bug Fixes
  - Load shards on first access per namespace; prune missing sources during that load. Load a namespace on `remove` to record the replaced fingerprint.
  - Added `SourceMessageCache::take(...)` to move an entry’s messages to its consumer while keeping lightweight metadata for invalidation and shard rewrites; parser paths now use `take` and stop cloning payloads.
  - Keep payloads only when needed: entries already marked dirty, or entries that actually hold retained history (non-empty retained set or pre-provenance Claude). Drained clean entries are preserved via on-disk shard reads during saves.
  - Further reduced peak heap: `-c claude` 890.7 → 757.2 MB; all clients 1617.5 → 1473.2 MB.

<sup>Written for commit 615d24cfaadd868d4e31ad097eaf730839796ead. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1108?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -785,11 +785,15 @@ fn parse_all_messages_with_pricing_with_cache_policy(
         }
     }
 
+    /// Takes the entry's messages by value: the cache handed this entry to
+    /// its one consumer (see `SourceMessageCache::take`), so cloning here
+    /// would put a second copy of the transcript alongside the first for no
+    /// reason.
     fn cached_messages(
-        cached: &message_cache::CachedSourceEntry,
+        cached: message_cache::CachedSourceEntry,
         pricing: Option<&pricing::PricingService>,
     ) -> Vec<UnifiedMessage> {
-        let mut messages = cached.messages.clone();
+        let mut messages = cached.messages;
         apply_pricing_to_messages(&mut messages, pricing);
         messages
     }
@@ -973,7 +977,7 @@ fn parse_all_messages_with_pricing_with_cache_policy(
             Option<&message_cache::SourceFingerprint>,
         ) -> Option<message_cache::FingerprintStatus>,
     {
-        let cached = source_cache.get(identity, path);
+        let mut cached = source_cache.take(identity, path);
         // An entry written before retention provenance existed cannot say
         // which of its rows the live transcript already dropped, so serving it
         // warm presents a retained copy of a response as a live one — and the
@@ -984,13 +988,14 @@ fn parse_all_messages_with_pricing_with_cache_policy(
         // the entry back with the provenance marker, so the next scan is a
         // plain warm hit.
         let rebuild_retention_provenance = cached
+            .as_ref()
             .is_some_and(message_cache::CachedSourceEntry::needs_retention_provenance_migration);
         #[cfg(test)]
         if rebuild_retention_provenance {
             RETENTION_PROVENANCE_REBUILDS.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
         }
         let Some(fingerprint_status) =
-            fingerprint_from_path(path, cached.map(|entry| &entry.fingerprint))
+            fingerprint_from_path(path, cached.as_ref().map(|entry| &entry.fingerprint))
         else {
             let (mut messages, _) = parse(path, None);
             apply_pricing_to_messages(&mut messages, pricing);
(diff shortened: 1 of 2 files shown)


junhoyeo:
Valid — fixed in 615d24cf. `take` now pins only entries that actually carry retained history: a Claude entry with a non-empty retained set, or a pre-provenance entry whose retained rows cannot be identified yet (the suggested `retained_message_keys().is_empty()` check alone would have drained those, and a second lookup would then re-derive them from the live bytes and retire them). The check is `CachedSourceEntry::holds_retained_history`, which reads the index vector directly instead of materializing the key set.

Measured on a real 254k-message Claude corpus: warm `-c claude` peak heap 890.7 MB -> 757.2 MB, and the full all-clients scan 1617.5 MB -> 1473.2 MB (2252.5 MB on main). Covered by `take_drains_a_claude_entry_with_no_retained_history`, `take_keeps_the_payload_of_an_entry_carrying_retained_history`, and `take_keeps_the_payload_of_a_pre_provenance_claude_entry`.

PR #1120, opened 2026-08-13
fix(droid): attribute session usage to the replies that spent it, and bound it server-side

junhoyeo:
Maintainer follow-up to #1090 by @Jackson57279. Their two commits are carried here with authorship intact; the rest are the corrections and the hosted-side guard the change needs before it can ship.

## What #1090 does

Droid records only a cumulative `tokenUsage` per session. Anchoring that total at a single instant cannot describe a session that ran for days: one day absorbs everything the session ever spent. Against a provider dashboard, a machine with three multi-day sessions reported 517M tokens for a day that had actually spent 24M. #1090 splits the session total across its assistant replies, weighted by the conversation each reply read, so a multi-day session reports against the days it actually ran.

## Commits

1. `fix(droid): anchor session usage at last write instead of provider lock` (@Jackson57279's original direction, rebased)
2. `fix(droid): treat a pre-epoch provider lock as an unusable anchor`
3. `fix(droid): attribute session usage to the replies that spent it` (@Jackson57279)
4. `fix(droid): weight a reply by the context it read, not its own output` (@Jackson57279)
5. `fix(droid): correct and bound the transcript-weighted usage split` - three defects in the apportioning: a pre-epoch transcript timestamp anchored a share of the session in a 1969 bucket no date filter reaches; output and reasoning were apportioned on the context each reply *read* rather than what it *produced*; and the transcript read was unbounded (measured 148ms per parse in release, 1.28s in debug, on a 104.7MB transcript with 10,000 replies, paid again on every scan while the session is live). Caps the read at 32 MiB and coalesces the turn list to 1024 runs. Droid parser version to 5.
6. `fix(frontend): bound every re-attributing parser by the device high-water` - the hosted guard, below.
7. `test(cli): pin the Droid submission generation to the server registry`
8. `fix(droid): keep the reply count, and fail closed on an unreadable transcript` - three review findings on the bounded read: a coalesced run reported one message instead of the calls it stands for; a failing `metadata()` mapped to length 0 and so waived the ceiling in exactly the case where the size was unknown; and a transcript whose read stopped early (`lossy_lines` ends silently on EIO) was apportioned as if it were whole, moving the unread tail's tokens onto the days the prefix covers. Droid parser version to 6.
9. `fix(droid): count one session, not one per apportioned record` - review finding. The split emits one record per assistant reply, and each carried its own message count. `sessionize` opens a new interval whenever two records sit more than the idle gap (3 min) apart and counts every interval that reports messages, so a session whose replies are minutes apart became one session per record: 40 replies ten minutes apart reported `sessionCount` 40, and a 1200-reply session coalesced to 1024 runs reported 600. `timeMetrics.sessionCount` is computed before `parserPlans` in the submit route and ratcheted with `GREATEST`, so the inflated figure bypasses the device high-water and sticks on a public profile. The session's authoritative reply total now rides on exactly one record and the rest are count-neutral, the same split `copilot_desktop` applies to its shutdown fragments: the number of calls stays exact, the session count returns to one. Also drops the unreachable `total_context > 0 && total_output > 0` guard - both weights are `.max(1)`.
10. `fix(droid): invalidate caches holding a count on every apportioned record` - a v6 entry carries a count on every record, and a finished session never rewrites its `settings.json`, so only a version bump discards it. Droid parser version to 7.

## The hosted hazard, and why commit 6 is not optional

Re-attribution moves tokens between days without changing the lifetime total, and the hosted merge cannot read it as a move. `mergeClientBreakdownsWithRegressionGuard` defends each stored `(day, client)` cell against a decrease, so the days that fall keep their old figure while the days that rise are written. The device's stored total gains exactly what moved, permanently, and again on every resubmit.

A single 321,000-token session re-attributed across three days stores 621,000. That is a regression test in this PR (`shows why the day-by-day merge alone inflates a re-attributed session`), not a hypothetical.

The device/client lifetime high-water added for Copilot already answers this: credit only growth bounded by both the per-cell delta and the cumulative aggregate, so a pure reshuffle contributes nothing while genuinely new usage still lands. The submit route hardcoded it to one client. Commit 6 drives it from `SUPPORTED_VERSIONED_PARSERS` and registers Droid there; Copilot behaviour is unchanged, including the `baseline-new` case that still merges its snapshot directly.

Droid is registered at **generation 1** - the generation every installed CLI already declares (`SUBMISSION_PARSER_VERSION`), not the message-cache `parser_version` this branch bumps to 7. The two Droid shapes differ in *where* a token lands, never in the lifetime total, so no installed generation has to be frozen out; registering above 1 would freeze every CLI in the field instead. Commit 7 pins that cross-language contract with a test, because the value comes from a shared default that a bump made for any other client would silently change.

### Two behaviour changes that follow, both shared with Copilot

- A CLI old enough to send no `scanScope` has its Droid rows ignored once the device holds a Droid high-water. An undeclared generation cannot be trusted to advance one.
- A date-filtered scan (`--since`/`--until`/`--year`) cannot establish or advance the high-water, so it freezes Droid rather than crediting a partial snapshot. The next unfiltered submit re-credits everything it covered.

Both are the existing Copilot rollout's tradeoff, applied to a second client. Flagging them explicitly since they are user-visible.

### Two limits this change does not remove

Both are stated rather than fixed; changing either needs its own decision.

- **The 32 MiB transcript ceiling routes the largest sessions back to the single mtime-anchored record.** That is the exact misattribution this PR exists to fix, and it survives for any session whose `*.jsonl` exceeds the cap - a long-running multi-day session is precisely the shape most likely to cross it. The alternative was an unbounded per-scan read that a live session pays again on every scan (measured 148 ms per parse in release, 1.28 s in debug, on a 104.7 MB transcript). The fallback total is still exact; only the day it lands on is wrong.
- **The per-cell high-water can permanently withhold credit for later genuine usage.** The guard bounds each `(day, client)` cell by growth, so a re-attribution that moves a day's share *down* leaves that cell's stored figure above what the device now reports for it. Real usage on that day afterwards is credited only once it exceeds the old stored figure. The direction is undercount, not inflation, and it is the same tradeoff the Copilot rollout already carries - but it is a real, silent loss for the affected days, not a transient one.

### What this does not fix

The high-water is per `(device, client)`. Rows already inflated by a re-attribution submitted *before* this deploys are not rolled back - the baseline preserves them and only bounds future growth. If any such submissions exist in production, they need a separate repair pass; nothing in this PR can distinguish them from real usage after the fact.

## Gates

Run at `4a0b4b30`, which merges current `main` (including the `windows-latest` `PathRoot::AppData` fix from #1121):

```
cargo fmt --all -- --check                                                      exit 0
cargo clippy --locked --workspace --all-targets --all-features -- -D warnings   exit 0
cargo test --locked --workspace --all-features --no-fail-fast                   exit 0  (2952 passed, 0 failed, 12 suites)
packages/frontend: npx vitest run                                               exit 0  (910 passed, 6 skipped, 88 files)
packages/frontend: npx tsc --noEmit                                             exit 0
packages/frontend: npx eslint .                                                 exit 0  (17 pre-existing warnings, none in touched files)
```

### Session-count evidence, from the built binary

Fixture: one Droid session, 40 assistant replies ten minutes apart, `inputTokens` 4000 / `outputTokens` 400.

```
# built at 3559a092 (before commit 9)
tokscale time-metrics --json --home $H --client droid   ->  session_count 40
tokscale models --json --home $H --client droid         ->  totalMessages 40, input 4000, output 400

# built at 4a0b4b30
tokscale time-metrics --json --home $H --client droid   ->  session_count 1
tokscale models --json --home $H --client droid         ->  totalMessages 40, input 4000, output 400

# same home, old binary first (writes a v6 cache), then the new one
old -> session_count 40 ; new over that cache -> session_count 1   (parser version 7 discards it)
```

Covered by `test_apportioned_session_still_sessionizes_as_one_session` and `test_reply_count_rides_on_exactly_one_record`.

The `Test & Coverage / Tests (windows-latest)` failure reported earlier came from `main`, not this branch; #1121 fixed it and this branch now carries that merge.

Refs #1090.

### DIFF
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -8232,6 +8232,21 @@ mod tests {
         );
     }
 
+    /// Droid is bounded by the server's device/client lifetime high-water
+    /// (`SUPPORTED_VERSIONED_PARSERS` in packages/frontend/src/lib/db/parserHighWater.ts),
+    /// which accepts generation 1 for it. The submission generation is not the
+    /// cache `parser_version`: re-attribution changes which day a token lands
+    /// on, never the lifetime total, so no installed generation has to be
+    /// frozen out. Declaring anything else here freezes every Droid submission
+    /// server-side until the registry is bumped in lockstep.
+    #[test]
+    fn submit_scan_scope_declares_the_droid_generation_the_server_registers() {
+        let clients = vec!["droid".to_string()];
+        let scope = submit_scan_scope(Some(&clients), true).expect("droid scope");
+
+        assert_eq!(scope.parser_versions.get("droid"), Some(&1));
+    }
+
     #[test]
     #[cfg(target_os = "macos")]
     #[serial_test::serial]
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1063,6 +1063,13 @@ fn parser_version(client: ClientId) -> u32 {
         // v2->v3: duplicate merging now upgrades the retained row when a later
         // copy carries an explicit cost, including zero.
         ClientId::MiMoCode => 3,
+        // Droid's cumulative session totals now anchor on the settings file's
+        // mtime (floored at providerLockTimestamp) instead of the lock
+        // timestamp alone, so a long-running session stops reporting every
+        // token it ever spent against the day it was started. A session that
+        // has since ended never changes its bytes again, so its fingerprint
+        // stays valid forever and only this bump discards the v1 anchor.
+        ClientId::Droid => 2,
         _ => 1,
     }
 }
@@ -2946,6 +2953,14 @@ mod tests {
         // fingerprint forever, so only the version bump discards the truncated
         // v6 parse and forces a cold reparse.
         assert_eq!(parser_version(ClientId::Grok), 7);
+    }
+
+    #[test]
+    fn test_droid_usage_anchor_parser_version_invalidates_v1_entries() {
+        // A finished Droid session's settings.json is never rewritten again, so
+        // its fingerprint keeps matching and only the version bump discards the
+        // v1 lock-timestamp anchor.
+        assert_eq!(parser_version(ClientId::Droid), 2);
     }
 
     #[test]
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1073,7 +1073,12 @@ fn parser_version(client: ClientId) -> u32 {
         // instant. It is now apportioned across the assistant replies in the
         // sibling transcript, weighted by the context each reply read, so a
         // multi-day session reports against the days it actually ran.
-        ClientId::Droid => 3,
+        // v3->v4: a reply is weighted by the context standing before it rather
+        // than including its own bytes, so a long answer no longer charges
+        // itself for its own output. Versions 2 and 3 only ever existed in
+        // pre-release builds of this change; the bump past them keeps anyone
+        // who ran one from holding the superseded weighting.
+        ClientId::Droid => 4,
         _ => 1,
     }
 }
@@ -2964,7 +2969,7 @@ mod tests {
         // A finished Droid session's settings.json is never rewritten again, so
         // its fingerprint keeps matching and only the version bump discards the
         // v1 lock-timestamp anchor.
-        assert_eq!(parser_version(ClientId::Droid), 3);
+        assert_eq!(parser_version(ClientId::Droid), 4);
     }
 
     #[test]
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1087,7 +1087,11 @@ fn parser_version(client: ClientId) -> u32 {
         // the transcript ceiling, and a transcript that could not be read whole
         // takes the single-record path rather than apportioning the session
         // over the prefix that was read.
-        ClientId::Droid => 6,
+        // v6->v7: the apportioned records are attribution fragments of one
+        // session, so the session's reply count now rides on exactly one of
+        // them. v6 entries carry a count on every record, which `sessionize`
+        // reads as one session per record.
+        ClientId::Droid => 7,
         _ => 1,
     }
 }
@@ -2978,7 +2982,7 @@ mod tests {
         // A finished Droid session's settings.json is never rewritten again, so
         // its fingerprint keeps matching and only the version bump discards the
         // v1 lock-timestamp anchor.
-        assert_eq!(parser_version(ClientId::Droid), 6);
+        assert_eq!(parser_version(ClientId::Droid), 7);
     }
 
     #[test]
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1082,7 +1082,12 @@ fn parser_version(client: ClientId) -> u32 {
         // takes the single-record path. Versions 2 through 4 only ever existed
         // in pre-release builds of this change; the bump past them keeps anyone
         // who ran one from holding a superseded split.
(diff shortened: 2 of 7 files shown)


junhoyeo:
All three review findings are fixed in 3559a092.

**Reply counts on coalesced runs** (@chatgpt-codex-connector) — correct, and the thread above is against `c144d0cb`. `TranscriptTurn` now carries `replies`, `coalesce_turns` sums it across a run, and the emitted record sets `message_count` from it. A 1200-reply session reports 1200 calls across at most 1024 records. Covered by `test_coalesced_session_still_reports_every_reply_it_made` and `test_uncoalesced_replies_each_count_as_one_message`.

**`metadata()` failure waived the ceiling** (@cubic-dev-ai) — `unwrap_or(0)` made the one case where the size is unknown the case that authorized an unbounded read. Now fails closed: no size, no split, single-record fallback with the identical total.

**Partial read apportioned as a whole transcript** (@cubic-dev-ai) — worth spelling out why this mattered beyond the read itself: the parse is cached under a fingerprint that a finished session never invalidates again, so a split derived from a truncated read would stick until the transcript changed. The parser now compares consumed bytes against the declared length and falls back on a short read. Reading past the declared length stays fine, since a live session appends while the parse runs.

Droid parser version to 6.

Gates at `3559a092`: `cargo fmt --check`, `cargo clippy --locked --workspace --all-targets --all-features -D warnings`, and `cargo test --locked --workspace --all-features` (2945 passed) all green locally, plus the frontend package's `vitest run` (910 passed), `tsc --noEmit`, and `eslint`.

`Tests (windows-latest)` is red on two Cherry Studio tests from #1105, not on anything here. #1105's head only ran cubic, Vercel and WIP — the Rust matrix never ran on it — so `main` is red on Windows and needs a separate fix.

junhoyeo:
Two more commits at `4a0b4b30`, plus a merge of current `main`.

**`fix(droid): count one session, not one per apportioned record`** — the apportioned records are attribution fragments of one session, but each carried its own message count. `sessionize` opens a new interval whenever two records sit more than the idle gap (3 min) apart and counts every interval reporting messages, so one Droid session became N sessions: 40 replies ten minutes apart reported `sessionCount` 40, a 1200-reply session coalesced to 1024 runs reported 600. `timeMetrics.sessionCount` is computed before `parserPlans` in the submit route and ratcheted with `GREATEST`, so it bypasses the device high-water and the inflated figure sticks on a public profile.

This does not walk back the coalescing finding above — the session's total call count is still exact. It now rides on exactly one record with the rest count-neutral, which is what `copilot_desktop` already does for its shutdown fragments (`copilot_desktop.rs:462-468`). Number of calls unchanged, session count back to one.

Evidence from the built binary, one session with 40 replies ten minutes apart:

```
built at 3559a092:  session_count 40, totalMessages 40, input 4000, output 400
built at 4a0b4b30:  session_count  1, totalMessages 40, input 4000, output 400
```

Also removed the unreachable `total_context > 0 && total_output > 0` guard — both weights are `.max(1)`, so a non-empty turn list always has positive weight.

**`fix(droid): invalidate caches holding a count on every apportioned record`** — a v6 entry carries a count on every record and a finished session never rewrites its `settings.json`, so only a version bump discards it. Droid parser version to 7. Verified: old binary populates the cache (`session_count` 40), new binary over that same cache reports 1.

The PR description now also states two limits this change does not remove: the 32 MiB transcript ceiling still routes the largest sessions back to the single mtime-anchored record, and the per-cell high-water can permanently withhold credit for later genuine usage on a day whose share moved down. Both are stated rather than fixed.

Gates at `4a0b4b30`: fmt, clippy `-D warnings`, `cargo test --locked --workspace --all-features` (2952 passed, 0 failed), frontend vitest (910 passed, 6 skipped), `tsc --noEmit`, eslint (0 errors). All 23 CI checks green, including `Test & Coverage / Tests (windows-latest)` — that failure came from `main` and #1121 fixed it.

Issue #888, opened 2026-07-14
Feature request: backfill/import historical usage from an aggregate export (e.g. clawdboard)


dvddvd300:
## Problem

`tokscale submit` computes totals from **raw local session files** (`~/.claude/projects`, `~/.codex/sessions`, etc.). That works great going forward, but it can't recover history that's no longer on disk:

- Claude Code's `cleanupPeriodDays` defaults to **30 days**, so transcripts older than a month are auto-deleted locally.
- Other tools/machines may have generated usage that was never on the machine where tokscale runs now.

Result: a user with, say, 5 months of real usage can only ever submit the last ~30 days, because the raw sessions behind the earlier months are gone. `scanner.extraScanPaths` / `TOKSCALE_EXTRA_DIRS` don't help — they expect **raw session directories**, not summarized data, and the raw data no longer exists.

Meanwhile, competing dashboards (e.g. clawdboard) accumulate this history **server-side** and can export it as clean daily aggregates. There's currently no sanctioned way to bring that history into tokscale.

## Proposed feature

A first-class, clearly-labeled backfill import, e.g.:

```
tokscale import --format clawdboard ./clawdboard-export.json
tokscale submit --backfill   # uploads imported aggregates, flagged as backfilled
```

Key design point (integrity): imported data should be **tagged distinctly** server-side (e.g. `origin: "backfill"` / `source: "import:clawdboard"`) so it never masquerades as CLI-verified data. That lets the leaderboard decide how to treat it — separate "verified" vs "backfilled" totals, an asterisk on rank, a separate historical view, exclusion from competitive ranking, etc. The goal is *recovering real history transparently*, not bypassing Level 1 validation.

## Example input schema (clawdboard export, anonymized)

```jsonc
{
  "exportedAt": "2026-07-14T17:45:44.315Z",
  "profile": {
    "name": "example-user",
    "email": "user@example.com",
    "githubUsername": "example-user",
    "createdAt": "2026-04-20T22:52:32.172Z"
  },
  "dailyAggregates": [
    {
      "date": "2026-05-11",
      "source": "codex",
      "machineId": "<uuid>",
      "inputTokens": 157910,
      "outputTokens": 5224,
      "cacheCreationTokens": 0,
      "cacheReadTokens": 112640,
      "totalCost": "0.5859",
      "premiumRequests": 0,
      "modelsUsed": ["gpt-5.5"],
      "modelBreakdowns": [
        {
          "modelName": "gpt-5.5",
[...]

PR #1128, opened 2026-08-14
feat: add ccusage import format

aron-intframe:
Adds `--format ccusage` to `tokscale import`, alongside the existing clawdboard importer. Addresses part of #888.

## Why ccusage and not another dashboard

The importer exists because raw sessions expire and aggregates outlive them. clawdboard is one place those aggregates survive, but `ccusage daily --json` is the format almost everyone already has a copy of: it is what people generate to submit to viberank and similar boards, and the file usually stays on disk long after `~/.claude/projects` has been rotated away. Supporting the ccusage shape directly covers the common recovery path without asking anyone to sign up for a third-party dashboard first.

## The part that needed care

A clawdboard row is scoped to one client — the row carries `source`, so attribution is a lookup. A ccusage unified row is not: one date is one row, every agent that ran that day is merged into it, and the participants are only named in `metadata.agents`. Attributing such a row to a single client silently moves spend between clients. On my own export a single day carries both `claude-opus-5` and `gpt-5.6-sol`; collapsing it would have shifted $10 of Codex spend onto Claude for that day alone.

So the resolver works per model, not per row, and it treats `metadata.agents` as ground truth rather than a hint:

- one agent that day → every model is trivially that client, no inference at all, which also means a private or unrecognized model name still lands correctly
- several agents → the model family picks among the agents that are actually listed
- a model matching none of them → left as `unknown` and reported in `unknown_clients`

That last case is deliberate. Folding an unmatched model into the first listed agent would produce a clean-looking import that is quietly wrong; the leaderboard rejects `unknown`, so surfacing it is what lets the caller fix the export instead of uploading bad attribution.

## Shape variations handled

- `period` (unified report) and `date` (per-agent reports, older releases) — a row with neither is rejected rather than silently dropped
- `reasoningOutputTokens`, which Codex reports separately. Dropping it understates the day and breaks the server's token-sum check
- `totalCost` as a number, or as a string when a dashboard has round-tripped the file

Reconciliation, clamping, cost sanitization, future-date and suspect-cost warnings reuse the existing helpers, so both importers report through the same `ImportOutcome` and the CLI summary needed no changes.

## Verification

Round-tripped a real 104-day export (2026-04-24 → 2026-08-15) recovered from a decommissioned machine:

```
Date range: 2026-04-24 to 2026-08-15
Active days: 104
Total tokens: 70,763,790,393
Total cost:   $91283.97
Clients:      claude, codex, unknown
```

Totals match the source file exactly. The split lands at `claude $51,680.46` / `codex $39,346.88`, and the only `unknown` is a $256.63 `Other` bucket that the upstream dashboard had already collapsed — which is precisely the case the resolver is meant to flag rather than guess at.

13 new unit tests cover the mixed-agent split, reasoning tokens, both date fields, the string-cost variant, single-agent trust, unattributable models, absent metadata, reconciliation, clamping, ordering, and dispatch. `cargo test --workspace --all-features`, `cargo clippy --locked --workspace --all-features -- -D warnings`, and `cargo fmt --all -- --check` all pass.

## Out of scope

This does not change the upload boundary. Import still only normalizes to a file, and the module docs still point at #888 for the server-side tagging that backfilled data would need before it could be submitted. I would rather this land as a way to *recover* history than as a way to route around that distinction.

---
## Summary by cubic
Adds `--format ccusage` to `tokscale import` to ingest `ccusage daily --json` exports and recover from `cc.json` when raw sessions are gone. Old: only `clawdboard`; new: `ccusage` supported. No change to output shape or the upload boundary (import still writes a normalized file only).

- Resolves unified rows per model using `metadata.agents`: single-agent days attribute directly; multi-agent days pick by model family; no match becomes `unknown` and is surfaced.
- Handles schema variants: accepts `period` or `date` (rejects rows with neither); accepts `totalCost` as number or string and `ccusage-codex` `costUSD`; reads reasoning from `metadata.reasoningOutputTokens` or the flat field; subtracts reasoning from `output` so totals align with `totalTokens`.
- Skips empty-day rows to avoid minting phantom `unknown` clients.
- Normalizes agent ids to match families (`claude-code`→`claude`, `gemini-cli`→`gemini`, `ccusage-codex`→`codex`) so models resolve to their own clients.
- Reconciles per-model sums against aggregates and declared `totalTokens`; warns on divergence; reuses existing clamping, cost sanitization, and warning paths (`ImportOutcome` unchanged).
- CLI help lists `ccusage`; default format remains `clawdboard`.

<sup>Written for commit fd3dc2d2b9c4e6bc7ec4ac80205c822684d19548. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1128?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/import.rs
+++ b/crates/tokscale-cli/src/commands/import.rs
@@ -1,10 +1,11 @@
 //! Import historical usage from third-party aggregate exports.
 //!
-//! Currently supports the clawdboard.ai account export ("daily aggregates").
-//! The importer normalizes that export into tokscale's native [`GraphResult`],
-//! which the CLI then writes out as standard tokscale JSON (identical in shape
-//! to `tokscale graph`) for review, archival, or a future server-supported
-//! backfill.
+//! Supports the clawdboard.ai account export ("daily aggregates") and
+//! `ccusage`'s own `daily --json` output (the `cc.json` that viberank and
+//! similar dashboards ingest). The importer normalizes either into tokscale's
+//! native [`GraphResult`], which the CLI then writes out as standard tokscale
+//! JSON (identical in shape to `tokscale graph`) for review, archival, or a
+//! future server-supported backfill.
 //!
 //! Motivation: `tokscale submit` computes totals from *raw* local session
 //! files. Once those files are gone (Claude Code deletes transcripts after
@@ -28,7 +29,7 @@ use tokscale_core::{
 };
 
 /// Import source formats understood by [`parse_export`].
-pub const SUPPORTED_FORMATS: &[&str] = &["clawdboard"];
+pub const SUPPORTED_FORMATS: &[&str] = &["clawdboard", "ccusage"];
 
 /// Result of a successful import.
 pub struct ImportOutcome {
@@ -73,6 +74,7 @@ pub struct ImportOutcome {
 pub fn parse_export(format: &str, json: &str) -> Result<ImportOutcome> {
     match format {
         "clawdboard" => parse_clawdboard_export(json),
+        "ccusage" => parse_ccusage_export(json),
         other => bail!(
             "unsupported import format '{}' (supported: {})",
             other,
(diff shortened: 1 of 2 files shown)


junhoyeo:
Reviewed and merging with a fix commit on top — thanks, this is a genuinely useful addition and the module docs, the reuse of the existing `ImportOutcome` reporting surface, and the `SUPPORTED_FORMATS` wiring were all right.

Your gates reproduced clean here (14/14 tests, clippy `-D warnings`, fmt). What I did do was point it at a real `cc.json` and at ccusage's documented schemas, and that turned up four divergences worth recording:

**Reasoning was counted twice.** ccusage's `totalTokens` is `input + output + cacheCreation + cacheRead` — reasoning is never a term of its own. That holds in 14/14 rows of the real export I tested and in every documented example. tokscale's buckets are additive and `total()` sums `reasoning` as a fifth term, so your own `CCUSAGE_MIXED` fixture imported as **8700 tokens against the 8000 that same row declares**. `output` now keeps only the non-reasoning remainder, matching what `dsh.rs`, `grok.rs`, `senpi.rs`, `zcode.rs` and `reasonix.rs` each do for this shape (`senpi.rs` carries a dedicated regression test for exactly this mistake).

**Reasoning was read from the wrong place.** The unified report nests it under `metadata`, not beside the token fields — only `ccusage-codex` uses the flat spelling. So on real unified exports the field never populated and reasoning was silently dropped, which is the undercount you added it to prevent. Both spellings are accepted now.

**An empty day minted a phantom client.** ccusage emits a row per day in the range. The real export has `2025-11-21` with `modelsUsed: []`, `modelBreakdowns: []`, 0 tokens, $0 — and the no-breakdowns fallback turned it into an `unknown` client plus the "leaderboard would reject this" warning. Before/after on the same file:

```
-  Clients: claude, unknown     Models: 5
-  Warning: unrecognized client id(s): unknown...
+  Clients: claude              Models: 4
```

Totals were unaffected either way (267,186,765 / $453.81, matching the file exactly).

**`ccusage-codex` spells cost `costUSD`** (ccusage#831), so those days imported at $0. Worth noting my first attempt at this was wrong in a way your kind of test caught: `rename_all = "camelCase"` derives `costUsd`, which matches nothing — it needs an explicit `rename`.

Also `gemini-cli` never resolved: `model_family` answers in canonical client ids and `resolve_client` compares the two as strings, so an agent name that keeps its CLI suffix dropped that agent's whole day into `unknown`.

I added the row's declared `totalTokens` as a reconciliation cross-check. It's the one statement a ccusage file makes about itself, and it's what catches a bucket being mapped through twice — worth having before these numbers ever reach a leaderboard.

One limitation I documented rather than fixed: `resolve_client` splits by model *family*, not by the client that made the call. On a day listing both `claude` and `opencode`, all Claude models land on `claude` even though OpenCode routinely drives them. The unified row doesn't carry enough to tell them apart — ccusage's `--by-agent` output does, since each agent gets its own `modelBreakdowns`. That's the natural follow-up if you're interested; I didn't want to build it against a shape I couldn't capture a real sample of.

`cargo test --workspace`: 3065 passed / 0 failed.

aron-intframe:
Late reply, and the lateness is mine — this deserved one at the time.

All four were right, and the first is the one I keep. 14/14 green while the fixture itself declared 8000 and we imported 8700. Our tests were self-consistent and had never once been pointed at a real `cc.json`. That is not a Rust mistake, it is a process one, so we changed the process rather than the diff: nothing goes out now unless it has been reconciled against a real file, a documented schema, or a sibling module in the same repo, with the declared value and ours recorded side by side. `senpi.rs` already carrying a regression test for exactly the mistake we made is the part that stung.

That rule is what #1190 was built under — 34 days of our own Claude Code and Codex logs, 39 agent-day cells, 30.1B tokens, reconciled against what the export declares before anything was opened. You merged it with no fix commit on top, which is the only measure of the change I trust.

Thanks for writing the four up in that detail instead of just pushing the fix. That comment did more for us than the merge did.

PR #1131, opened 2026-08-16
fix(antigravity-cli): resolve routing-label rows from display labels

### LINKED ISSUE #1116: Antigravity CLI: gemini-default routing label is excluded despite a concrete display label in the same row
## Summary

Antigravity CLI messages whose `chatModel.responseModel` (`#19`) is the routing label `gemini-default` are excluded from leaderboard submission as unpriced — even though the same row carries a concrete display label (`#21`, e.g. `"Gemini 3.5 Flash (Medium)"`) that names the actual model that answered. The parser preserves the routing label verbatim and never consults the display label for these rows, so ~9.5M tokens were silently dropped.

I understand the current behavior is deliberate (see #1019, #1026, #1071): a routing label names the router, not the model, so aliasing `gemini-default` to a specific model was reverted in favor of honest exclusion. This issue is not about re-aliasing the label — it's about the fact that for Antigravity CLI the actual model is **already present in the same row** (display label `#21`), and tokscale ignores it for pricing.

## Observed behavior

`npx tokscale@latest submit`:

```
Warning: excluded 182 unpriced google/gemini-default message(s) (9,491,597 tokens): generic routing label has no authoritative model-to-price mapping.
```

Affected data in `~/.gemini/antigravity-cli/conversations/*.db` (`gen_metadata` blob): 8 conversations, 182 messages. Display labels carried in the very same rows:

| Display label (#21) | Messages |
|---|---|
| Gemini 3.5 Flash (Low) | 109 |
| Gemini 3.5 Flash (Medium) | 44 |
| Gemini 3.5 Flash (High) | 29 |

Token breakdown of the excluded usage: input 2,507,742 / output 37,487 / cache-read 6,910,666 / reasoning 35,702.

## Why this happens

Each Antigravity CLI `gen_metadata` row embeds a `chatModel` protobuf with:

- `#19` `responseModel` — machine id. For tiered Gemini Flash responses this is the routing label `gemini-default`.
- `#21` — display label, e.g. `"Gemini 3.5 Flash (Medium)"`.

In tokscale:

1. The Antigravity CLI parser preserves `#19` verbatim whenever present — see `gemini_default_response_model_is_preserved` in `crates/tokscale-core/src/sessions/antigravity_cli.rs`. The display label is only used to *recover* a **missing** `#19` (`SessionModels::by_display`), never to override a routing label.
2. At submit time, `exclude_unpriced_submission_messages` classifies `google` + `gemini-default` as a generic routing label via `is_generic_routing_label` (`crates/tokscale-core/src/lib.rs`) and excludes the rows.
3. The resolver refuses the label on its own: `default` is in `FUZZY_BLOCKLIST` (`crates/tokscale-core/src/pricing/lookup.rs`).

Net result: a row that unambiguously names a concrete, priceable model in `#21` is dropped because its `#19` is a routing label — even though the parser already knows how to map that display label to the real machine id, and uses that exact mechanism for rows with a missing `#19`.

## Suggested fix

When a row's `#19` is a routing label (i.e. `lookup::is_routing_label` or the `google/gemini-default` pair), fall back to the display-label resolution already built for missing `#19`: look up the `#21` label in the per-file `SessionModels::by_display` map and price the recovered canonical model (e.g. `gemini-3.5-flash` at LiteLLM rates).

This reuses existing, well-tested machinery (including the ambiguity handling that drops labels mapping to multiple priced models), and it tracks future routing changes automatically — when Google retargets the `default` label to a newer Flash, pricing follows the actual served tier instead of going stale. It also only applies where the display label is authoritative (same file), so it stays consistent with the "labels are never pricing keys on their own" rule.

## Workaround

An exact custom override in `~/.config/tokscale/custom-pricing.json` keyed `gemini-default` is accepted (resolution runs before the routing-label exclusion). But it is a static pin that cannot follow the routing label to newer models — the maintenance burden this issue would remove.

## References

- #1019 — original report: `google/gemini-default` blocked submission
- #1026 — merged fix: aliased `gemini-default -> gemini-2.0-flash-lite` (later superseded by the routing-label exclusion)
- #1071 — routing labels (`auto`) are excluded rather than mispriced; `gemini-default` is the precedent

## Environment

- tokscale version: latest (`npx tokscale@latest`)
- Client: Antigravity CLI (data under `~/.gemini/antigravity-cli/conversations/*.db`)
- Model family: Gemini 3.5 Flash (Low/Medium/High) via the `gemini-default` routing label

azestkingscrown:
## Problem

Antigravity CLI writes the generic routing label `gemini-default` as `chatModel.responseModel` (`#19`) whenever the router picked the model, leaving the concrete tier only in the sibling `#21` display label (e.g. `"Gemini 3.5 Flash (Low)"`). The parser preserved the label verbatim, so those rows were excluded from `tokscale submit` as unpriced — despite carrying a tier-identifying display label in the same row.

On real data this silently dropped **~218M tokens / ~$74** of Gemini usage across 9 Antigravity CLI conversations (reported in #1116).

## Root cause

1. The Antigravity CLI parser (`crates/tokscale-core/src/sessions/antigravity_cli.rs`) keeps `#19` verbatim when present. `gemini-default` is not a concrete model id, so it later trips the submission-time routing-label exclusion (`is_generic_routing_label` in `lib.rs`).
2. `SessionModels::from_blobs` treated `gemini-default` as a real model id, so a file mixing routing labels and real ids would report a false ambiguity and drop the display-label mapping entirely.

## Change

Resolve routing-label rows in three steps, in `parse_gen_metadata`:

1. A concrete `#19` is used as before.
2. Otherwise recover the machine id from a sibling row sharing the display label (the existing `SessionModels::recover` mechanism).
3. Otherwise map the display label itself through a small verified lookup table (`display_label_to_model_id`).

A routing label is never treated as evidence of a concrete model, so it no longer occupies the `by_display` slot and can't poison the ambiguity check against a real sibling id. Rows that still resolve to nothing keep `gemini-default` verbatim, so the submission-time exclusion continues to apply rather than guessing a model.

Display labels are server-supplied names (the app's `_getModelLabel` switch emits them from `MODEL_PLACEHOLDER_*` ids) and could be renamed or localized, so only labels verified against real user databases are mapped — `Gemini 3.5 Flash (Low)` → `gemini-3.5-flash`, `(Medium)` → `gemini-3.5-flash-medium`, `(High)` → `gemini-3.5-flash-high`. Anything unknown returns `None` and preserves the routing label.

## Verification

- `cargo clippy --locked --workspace --all-features -- -D warnings` — clean
- `cargo test --workspace --all-features` — all pass (25 antigravity_cli tests including 5 new ones)
- Built the CLI and ran `tokscale models` against real local data: the `gemini-default` row disappears and the usage resolves to `gemini-3.5-flash` ($57.09) and `gemini-3.5-flash-medium` ($16.92), increasing the reported total by ~$74 — matching the previously excluded amount.

Fixes #1116

---
## Summary by cubic
Resolves `antigravity-cli` routing-label rows by recovering concrete Gemini model IDs from display labels or sibling evidence, so previously excluded usage is priced. Also maps "Gemini 3.5 Flash (Low)" to `gemini-3.5-flash-extra-low` to align with the alias table and avoid splitting Low-tier usage across model rows.

- Parser changes in `tokscale-core`: updates `parse_gen_metadata`; adds `is_antigravity_routing_label` and `display_label_to_model_id`.
- Behavior: previously preserved `gemini-default` in `#19` and `tokscale submit` skipped those rows; now resolve via concrete `#19` > sibling sharing `#21` display label > verified display-label map; else keep `gemini-default`.
- Routing labels never occupy the `by_display` slot; recovered IDs still resolve through pricing aliases.

**Rollout**
- No config changes. Re-run `tokscale models` and `tokscale submit` to include previously excluded Gemini usage.

<sup>Written for commit 0ea0ea8134bca9c7c5d8e49d27a1db7ea1d8199a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1131?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/antigravity_cli.rs
+++ b/crates/tokscale-core/src/sessions/antigravity_cli.rs
@@ -146,6 +146,15 @@ impl SessionModels {
                 unresolved_labels.extend(label);
                 continue;
             };
+            if is_antigravity_routing_label(model) {
+                // A routing label names the router that served the request, never
+                // the model that answered it, so it is no evidence of a concrete
+                // model for this display label. Treat it like a row with no #19:
+                // it must not occupy the by_display slot, which would trip the
+                // ambiguity check against a real sibling id and drop the mapping.
+                unresolved_labels.extend(label);
+                continue;
+            }
             distinct.insert(model);
             if let Some(label) = label {
                 by_display
@@ -214,6 +223,31 @@ impl SessionModels {
     }
 }
 
+/// Antigravity CLI's generic routing label. It names which router served the
+/// request, never which model did, so a row carrying it carries no model
+/// identity of its own. Kept in sync with `is_generic_routing_label` in lib.rs,
+/// which excludes the label from submission when no concrete model is recovered
+/// here.
+fn is_antigravity_routing_label(model: &str) -> bool {
+    model.trim().eq_ignore_ascii_case("gemini-default")
+}
+
+/// Map an Antigravity `#21` display label to a concrete model id.
+///
+/// Display labels are server-supplied names (the app's `_getModelLabel` switch
+/// emits them from `MODEL_PLACEHOLDER_*` ids) and could be renamed or localized,
+/// so only labels verified against real user databases are mapped here; anything
+/// else returns `None` and the routing label is preserved for the submission-time
+/// exclusion rather than guessed at.
+fn display_label_to_model_id(label: &str) -> Option<&'static str> {
+    match label.trim() {
+        "Gemini 3.5 Flash (Low)" => Some("gemini-3.5-flash"),
+        "Gemini 3.5 Flash (Medium)" => Some("gemini-3.5-flash-medium"),
+        "Gemini 3.5 Flash (High)" => Some("gemini-3.5-flash-high"),
+        _ => None,
+    }
+}
+
 fn parse_gen_metadata(
     blob: &[u8],
     session_id: &str,
@@ -262,8 +296,18 @@ fn parse_gen_metadata(
         }
     }
 
-    let model_raw = non_empty_string_field(chat_model, 19)
+    let response_model = non_empty_string_field(chat_model, 19);
+    let model_raw = response_model
+        .filter(|m| !is_antigravity_routing_label(m))
         .or_else(|| session_models.recover(chat_model))
+        .or_else(|| {
+            // Routing labels carry no model identity of their own, but the
+            // sibling `#21` display label names the tier. Recover the concrete
+            // id from it; if the label is unknown, keep the routing label so
+            // the submission-time exclusion still applies.
+            non_empty_string_field(chat_model, 21).and_then(display_label_to_model_id)
+        })
+        .or(response_model)
         .unwrap_or("unknown");
     let model_id = pricing::aliases::resolve_alias(model_raw)
         .unwrap_or(model_raw)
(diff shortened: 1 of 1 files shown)


junhoyeo:
Not mergeable as-is, but only one line needs to change. The design is right and the tests are real.

`crates/tokscale-core/src/sessions/antigravity_cli.rs:244` maps "Gemini 3.5 Flash (Low)" to `gemini-3.5-flash`, but this crate already pins that exact display name to a different id. From `crates/tokscale-core/src/pricing/aliases.rs:73-76`, sourced from the pinned Antigravity Context Window Monitor SHA: "M187 is a distinct placeholder whose own `activeModelSpecs` entry has `modelId: 'gemini-3.5-flash-extra-low'` and `displayName: 'Gemini 3.5 Flash (Low)'` — the true Low tier." And `aliases.rs:85` maps `model_placeholder_m187` to `gemini-3.5-flash-extra-low`.

The result is that two tests pass at once on this branch and contradict each other. `pricing::aliases::tests::m187_and_m20_resolve_to_distinct_tiers_but_both_still_price` asserts the "(Low)" placeholder resolves to `gemini-3.5-flash-extra-low`, while the new `sessions::antigravity_cli::tests::routing_label_resolved_from_display_label_parsing` asserts at `antigravity_cli.rs:1340` that the same "(Low)" label resolves to `gemini-3.5-flash`. Same label, two ids, so a user on both the IDE and the CLI gets Low-tier usage split across two model rows.

To be fair about blast radius: this is not a cost error today. The dataset carries only the generic `google/gemini-3.5-flash` entry, and I confirmed `gemini-3.5-flash`, `-extra-low`, `-medium` and `-high` all land on it, so every tier prices identically right now. It is a model-identity error that becomes a cost error the day per-tier rows appear, which is the exact scenario `aliases.rs:76-79` was written to guard against. Changing line 244 to `gemini-3.5-flash-extra-low` fixes it, and the value passes through `resolve_alias` at `antigravity_cli.rs:312` unchanged, same as the other two arms. The "(Medium)" and "(High)" arms are correct as written; they match M20 and M133/M132.

Everything else holds up:

- Excluding routing-label rows from `by_display` at `antigravity_cli.rs:149-157` is the right fix rather than a workaround. A `gemini-default` row occupying the slot would read as a genuine conflict against a sibling id and clear the mapping entirely, and `routing_label_does_not_poison_by_display_ambiguity` pins that.
- The precedence in `parse_gen_metadata` at `antigravity_cli.rs:300-311` is ordered correctly: a concrete `#19` first, then same-file sibling evidence, then the hardcoded label table, so observed evidence always beats the table. The final `.or(response_model)` at line 310 deliberately restores the routing label so the exclusion at `lib.rs:3911` still fires, and `routing_label_without_any_evidence_is_kept` covers it. That is the case most easily got wrong.
- On ambiguity: many display labels sharing the routing label is harmless since resolution is per-row from that row's own `#21`, and the reverse case of one label with two concrete ids is still handled by the alias-canonical comparison at `antigravity_cli.rs:173-181`.
- The split is 134 test lines to 45 production lines, and the tests assert `model_id` on messages returned from `parse_antigravity_cli_file` rather than probing the helper in isolation. That is the right level.
- No interaction with f169a1f0. That commit is entirely within `crates/tokscale-cli/src/antigravity.rs` and concerns TLS/plaintext probing on the DesktopAgent transport; this PR is `crates/tokscale-core/src/sessions/antigravity_cli.rs` and concerns protobuf parsing of the local database. Different crates, no shared code, no conflict in intent.

One follow-up, not blocking: `is_antigravity_routing_label` at `antigravity_cli.rs:231` is now the third place that knows the string `gemini-default`, alongside `is_generic_routing_label` at `lib.rs:3838-3839` and, structurally, `is_routing_label` at `lookup.rs:2543` whose `ROUTING_LABELS` list does not include it. The doc comment saying it is kept in sync with `lib.rs` is honest, but nothing enforces it. Worth collapsing to one predicate eventually.

Nit: `is_antigravity_routing_label` compares case-insensitively while `display_label_to_model_id` is an exact match. That is probably deliberate given the labels are server-supplied proper nouns and an unrecognized one falls back safely, but the asymmetry is unexplained.

CI is a complete green run here. Locally, `cargo test -p tokscale-core --lib` on the PR head passes 1784/1784.

azestkingscrown:
Done — `display_label_to_model_id` now maps `"Gemini 3.5 Flash (Low)"` to `gemini-3.5-flash-extra-low`, matching `aliases.rs:85` (`model_placeholder_m187`). CI is green, both `routing_label_resolved_from_display_label_parsing` and `m187_and_m20_resolve_to_distinct_tiers_but_both_still_price` pass on this head.

junhoyeo:
Verified and merging. `display_label_to_model_id` now agrees with the SHA-pinned mapping in `aliases.rs:85`, and the two tests that previously asserted different ids for the same display label — `pricing::aliases::tests::m187_and_m20_resolve_to_distinct_tiers_but_both_still_price` and `sessions::antigravity_cli::tests::routing_label_resolved_from_display_label_parsing` — now pass against the same target. Thanks for the quick turnaround.

PR #1132, opened 2026-08-16
perf(cli): drop the image codecs and SVG font stacks nothing renders

junhoyeo:
## What

`tokscale`'s release binary was linking two largely unused subsystems: every image codec `image` ships, and resvg's font-shaping and raster-image stacks. Nothing in the tree can reach either. This PR pins the six dependencies responsible, in two atomic commits.

**Result: 17,061,280 -> 15,275,072 bytes on stripped aarch64-apple-darwin — 1.70 MiB, 10.5%.**

| commit | binary | step | cumulative |
|---|---:|---:|---:|
| baseline `f169a1f0` | 17,061,280 | — | — |
| link only the codecs wrapped uses | 16,333,392 | −711 KiB | −4.27% |
| stop linking resvg's font/raster stacks | 15,275,072 | −1,033 KiB | **−10.47%** |

For scale: `cargo bloat` puts the **entire 50-file session-parser layer at 824 KiB of `.text`**. This PR removes more than twice that, from dependency metadata alone.

## Commit 1 — image codecs

`wrapped` is the only consumer of `image`: it decodes the PNG/JPEG/WebP assets it downloads and encodes a PNG. A census of every image asset the CLI references is exactly 26 `.png`, 13 `.jpg`, 1 `.webp` and 1 `.svg`. The build was nonetheless linking AVIF, BMP, DDS, EXR, Farbfeld, GIF, HDR, ICO, PNM, QOI, TGA and TIFF.

Setting `default-features = false` on `image` alone produced a **byte-for-byte identical binary** — three sibling dependencies re-enable the codecs through Cargo feature unification:

| Dependency | Default feature | What it drags in |
|---|---|---|
| `imageproc` | `["rayon", "image/default"]` | re-enables the entire codec set; only `rayon` is wanted |
| `qrcode` | `["image", "svg", "pic"]` | three renderers, while `tokscale qr` uses `qrcode::render::unicode` from the crate core |
| `arboard` | `["image-data"]` | `image` (with `tiff` on macOS) plus CoreGraphics/Win32 GDI bitmap paths, while the TUI yank key only calls `Clipboard::set_text` |

`imageproc` is the subtle one: it declares `image` with `default-features = false` in its dependency table, then re-enables everything through its own `[features] default`.

## Commit 2 — resvg font and raster stacks

The only SVG the CLI ever rasterises is the wrapped footer logo, which is three `<path>` elements with no `<text>`, no `<image>` and no `font-family`.

- `raster-images` links resvg's own `gif`, `image-webp` and `zune-jpeg` for bitmaps embedded in SVG. They resolve to `image-webp 0.1.3` and `zune-jpeg 0.4.21`, which do **not** unify with the `0.2.4`/`0.5.12` that `image` already provides — so the binary carried two independent copies of the WebP and JPEG decoders.
- `text` pulls `fontdb`, `rustybuzz`, `unicode-bidi`, `unicode-script` and `unicode-vo` to shape SVG text runs. Wrapped draws its own text with `ab_glyph` and never asks usvg to lay out a glyph.

Removed outright: `rustybuzz`, `fontdb`, `gif`, `unicode-bidi`, `unicode-script`, `unicode-vo`, and the duplicate `zune-jpeg` / `image-webp` / `ttf-parser` versions.

## Verification

- `cargo test --release -p tokscale-core -p tokscale-cli` — 0 failures.
- `cargo clippy --all-targets -- -D warnings` — clean. `cargo fmt --check` — clean.
- `cargo tree -e features -i image` resolves to exactly `png`, `jpeg`, `webp`; `exr`, `tiff`, `ravif`, `gif`, `rustybuzz`, `fontdb` are gone from the graph.
- Duplicate crate versions drop from 16 pairs to 10.
- **Rendering checked against a cleared cache.** The logo fetch is wrapped in `if let Ok(..)`, so a rasterisation failure would silently omit the logo rather than error. I emptied `~/.config/tokscale/cache/images` to force a genuine re-fetch and re-rasterise, then inspected the output: the logo mark and the TOKSCALE wordmark render exactly as before, and `wrapped` still emits a 2400x2400 RGBA PNG with provider icons and fonts intact.

Note the rendered PNG is not byte-reproducible between runs because the underlying usage data is live; a same-binary control run establishes that, so comparison was done visually and structurally rather than by hash.

## Not done

`comfy-table` pulls a second `crossterm` (0.29 alongside ratatui's 0.28). Its `tty` feature is what provides terminal-width detection, and the CLI uses `ContentArrangement::Dynamic` / `DynamicFullWidth` in three places, so dropping it would silently stop tables filling the terminal. Left alone deliberately.

`arboard` and `qrcode` only compile-verify on macOS locally; the Linux/Windows/Android targets are covered by `build-native.yml` on this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,24 +57,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "aligned"
-version = "0.4.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ee4508988c62edf04abd8d92897fca0c2995d907ce1dfeaf369dac3716a40685"
-dependencies = [
- "as-slice",
-]
-
-[[package]]
-name = "aligned-vec"
-version = "0.6.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc890384c8602f339876ded803c97ad529f3842aba97f6392b3dba0dd171769b"
-dependencies = [
- "equator",
-]
-
-[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -155,40 +137,20 @@ dependencies = [
 ]
 
 [[package]]
-name = "arbitrary"
-version = "1.4.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c3d036a3c4ab069c7b410a2ce876bd74808d2d0888a82667669f8e783a898bf1"
-
-[[package]]
 name = "arboard"
 version = "3.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0348a1c054491f4bfe6ab86a7b6ab1e44e45d899005de92f58b3df180b36ddaf"
 dependencies = [
  "clipboard-win",
- "image",
  "log",
  "objc2",
  "objc2-app-kit",
- "objc2-core-foundation",
- "objc2-core-graphics",
  "objc2-foundation",
  "parking_lot",
  "percent-encoding",
  "windows-sys 0.60.2",
  "x11rb",
-]
-
-[[package]]
-name = "arg_enum_proc_macro"
-version = "0.3.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0ae92a5119aa49cdbcf6b9f893fe4e1d98b04ccbf82ee0584ad948a44a734dea"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


junhoyeo:
@codex review

Specifically: this changes dependency features only, but the risk is behavioural rather than syntactic. Please check whether any code path can still reach a codec or renderer that has been switched off — `image` is now `png`/`jpeg`/`webp` only, `imageproc` loses `image/default`, `qrcode` loses the `image`/`svg`/`pic` renderers, and `arboard` loses `image-data`.

junhoyeo:
## Asset-format audit against the live network

Since this PR narrows `image` to `png`/`jpeg`/`webp`, the risk is not a compile error but a runtime decode failure on some asset whose real format is not in that set. Extensions are not evidence, so I fetched every image URL the CLI can request and sniffed the actual magic bytes.

All 25 literal asset URLs in `wrapped.rs`:

| detected format | count |
|---|---:|
| PNG | 15 |
| JPEG | 7 |
| WebP | 1 |
| SVG (handled by usvg, not `image`) | 1 |
| non-200 | 1 |

Nothing outside the enabled set. Two findings worth recording:

**GitHub avatar URLs serve JPEG despite the `.png` extension.** `https://github.com/<user>.png` returns `content-type: image/jpeg` for every account I checked (`junhoyeo`, `PrimeIntellect-ai`, `torvalds`). That is fine here because `jpeg` is enabled, but it means the extension in these URLs is not a reliable signal, and anyone narrowing this feature list further should sniff rather than trust the suffix.

**`https://tokscale.ai/assets/logos/openclaw.png` returns 404** (`text/html`). This is pre-existing on `main` and unrelated to this PR — the fetch is fault-tolerant, so the OpenClaw logo is silently omitted from wrapped output today. Flagging it as a separate bug rather than fixing it here.

PR #1135, opened 2026-08-16
fix(frontend): make quickstart label scroll to commands

### LINKED ISSUE #1107: Make your landing page "QUICKSTART" label a button
I do understand each website has its own design, but I STILL CANNOT BELIEVE THIS BLUE ELEMENT IS A F**KING LABEL INSTEAD OF A CLICKABLE BUTTON.

Image

It might feels like I'm a little bit angry, and I shouldn't be in that way if my friend didn't tell me this is not clickable, after I criticized the BUTTON has a too large width. 

PLEASE, at least make it a link scrolling down to your two commands below, DO NOT LEAVE IT UNCLICKABLE.

xtianxx:
## Summary

- Make the QUICKSTART label on the landing page clickable.
- Link it to the Quick Start section.

Closes #1107

## Testing

- `make test/frontend`
- `make typecheck`
- Manual verification

---
## Summary by cubic
Makes the landing page Quickstart label a button that smoothly scrolls to the command cards to improve discoverability and keyboard navigation. Previously it was a non-interactive label; now it performs in-page scroll with proper focus and hover states.

- Converts `QuickstartLabel` from `div` to `button` and adds an onClick that calls `scrollIntoView({ behavior: "smooth", block: "start" })` on the cards wrapper.
- Adds `scroll-margin-top: 72px` to `QuickstartCardsWrapper` to prevent overlap with the fixed nav.
- Preserves styling and adds cursor, hover/active feedback, and a focus-visible outline; removes the external link behavior.

<sup>Written for commit deaa7b316be848674b429937ebfbfcfdea44bf98. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1135?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/src/components/landing/sections/QuickstartSection.tsx
+++ b/packages/frontend/src/components/landing/sections/QuickstartSection.tsx
@@ -1,5 +1,6 @@
 "use client";
 
+import { useRef } from "react";
 import Image from "next/image";
 import styled, { css, keyframes } from "styled-components";
 import { useCopy, useSquircleClip } from "../hooks";
@@ -8,12 +9,21 @@ import { SquircleBorder } from "../components";
 export function QuickstartSection() {
   const tui = useCopy("bunx tokscale@latest");
   const submit = useCopy("bunx tokscale@latest submit");
+  const cardsWrapperRef = useRef<HTMLDivElement>(null);
   const {
     setElementRef: setCardsRowRef,
     clipPath: cardsRowClipPath,
     svgDef: cardsRowSvgDef,
     borderDef: cardsRowBorderDef,
   } = useSquircleClip<HTMLDivElement>(32, 0.6, true, 1);
+
+  const scrollToCommands = () => {
+    const wrapper = cardsWrapperRef.current;
+    if (!wrapper) return;
+
+    // Scroll to the wrapper, not the row — the row's clip-path would clip the target.
+    wrapper.scrollIntoView({ behavior: "smooth", block: "start" });
+  };
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


junhoyeo:
Thanks @xtianxx. The mechanics here are right — `styled.div` to `styled.a`, `rel="noopener noreferrer"` alongside `target="_blank"`, and `text-decoration: none` to keep the visual — and `#installation` does exist in the README, so the link resolves.

Two things before this goes in.

**The destination does not match what #1107 asked for.** The issue is specific: "at least make it a link scrolling down to your two commands below". This navigates off-site to GitHub in a new tab instead. The two commands are rendered directly beneath the label in this same component — `QuickstartCardsWrapper` at `QuickstartSection.tsx:48`, holding `bunx tokscale@latest` and `bunx tokscale@latest submit`, each already with its own copy button. Sending someone to GitHub to read installation docs when the thing they want is 40px below, and already copyable, is a worse outcome than the current unclickable label for anyone who follows it.

An in-page anchor needs an `id` on the wrapper, since the file has none today (the only `id` is the SVG clip-path at `:30`).

**The affordance problem is the actual complaint, and it is unaddressed.** #1107 is not really "this has no href", it is "I could not tell this was interactive". `QuickstartLabel` at `:139` has no hover or focus-visible state, so at rest it looks exactly as it did before. `<a href>` does inherit `cursor: pointer` from the UA stylesheet, which helps once the pointer is already over it, but nothing changes on hover and there is no keyboard focus ring. Someone scanning the page still cannot tell it does anything.

Suggest pointing it at the commands below and adding a hover and `:focus-visible` treatment. If linking to the README is genuinely preferred over scrolling, that is your call — but then it is worth saying so on #1107, because it is not what was requested.

junhoyeo (approved):

PR #1136, opened 2026-08-16
feat: add MiniMax Code headless usage tracking

hetaoBackend:
## Summary

- register MiniMax Code as a first-class Tokscale client across the Rust core, CLI/TUI, frontend registry, and documentation
- add `tokscale headless mcode exec ...`, automatically selecting MCode's `stream-json` output
- parse authoritative assistant usage and pair it with the provider/model reported by the final `exec.result`
- keep MiniMax Code attribution scoped to Tokscale's headless capture directory instead of scanning the shared Desktop/Runtime session store

## Compatibility

MiniMax Code streams that do not contain `model.providerId` and `model.modelId` in the final `exec.result` are deliberately skipped. This prevents partial captures and older CLI releases from being priced against a guessed model.

## Validation

- `cargo test -p tokscale-core sessions::mcode -- --nocapture`
- `cargo test -p tokscale-core test_scan_all_clients_headless_paths -- --nocapture`
- `cargo test -p tokscale-cli mcode -- --nocapture`
- `cargo test -p tokscale-cli test_client_ -- --nocapture`
- `cargo clippy -p tokscale-core --lib -p tokscale-cli --bins -- -D warnings`
- `bun run --cwd packages/frontend test -- __tests__/lib/clientRegistry.test.ts`
- `bun run --cwd packages/frontend test -- __tests__/components/sourceLogo.test.tsx`
- `bun run --cwd packages/frontend typecheck`

`cargo clippy --all-targets` also reaches a pre-existing `unnecessary_sort_by` denial in `sessions/cherrystudio.rs` under Rust 1.97.1; the changed library and binary targets pass with warnings denied.

---
## Summary by cubic
Adds MiniMax Code headless usage tracking across `tokscale-core`, `tokscale-cli`, and the frontend. Previously headless capture supported only Codex and a bad UTF‑8 byte could drop an entire stream; now `tokscale` ingests `mcode exec` streams reliably and attributes usage to the exact provider/model.

- Registers `mcode` as a headless client (`~/.config/tokscale/headless/mcode/*.jsonl`; override with `TOKSCALE_HEADLESS_DIR`) and includes it in scanning, counts, TUI hotkeys, and frontend (name, logo, color).
- CLI: `tokscale headless mcode exec ...` is supported. It injects `--output-format stream-json` immediately after `exec` unless `--no-auto-flags` is set, requires the `exec` subcommand, and errors on non‑JSONL output. `tokscale clients` now reports headless support and message counts for both `codex` and `mcode`.
- Parser: buffers assistant `message` usage and emits it only when the final `exec.result` includes `model.providerId` and `model.modelId`; otherwise the stream is skipped to avoid mispricing. Uses lossy line reading to continue past undecodable bytes instead of truncating the capture.
- Docs: README documents MiniMax Code headless capture paths and examples.

<sup>Written for commit 7b01286cf273311ce895c6ff5e4052c435038783. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1136?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo:
This is close to mergeable — the integration checklist is done properly and the token accounting is correct, which is the part that usually goes wrong. One line should change before merge, and the docs are half-finished.

## The one thing to fix first

`crates/tokscale-core/src/sessions/mcode.rs:37` reads the capture with `BufReader::new(file).lines().map_while(Result::ok)`. That is the exact spelling `crates/tokscale-core/src/sessions/utils.rs:13` documents as the #1031 bug: `BufRead::lines()` yields `Err(InvalidData)` for any line that is not valid UTF-8, and `map_while(Result::ok)` turns that into end-of-iteration, so one stray byte silently drops every record after it.

It is worse in this parser than in the ones that were fixed. Usage is buffered in `pending_by_turn` and only flushed when `exec.result` arrives (`mcode.rs:115`), and nothing flushes at EOF. So a bad byte anywhere *before* the final `exec.result` does not truncate the tail — it drops the entire capture to zero messages. I confirmed the iterator behaviour on a fixture shaped like a real stream (assistant usage line, a line containing a lone `0xFF`, then `exec.result`): 3 lines on disk, 1 line yielded, `exec.result` never seen, 0 messages emitted. Reading the same file with the `read_until` + `from_utf8_lossy` approach yields all 3.

The fix is `utils::lossy_lines(BufReader::new(file))`, which is what `pi.rs`, `grok.rs`, `dsh.rs`, `prime_agent.rs`, `droid.rs`, `reasonix.rs`, and both copilot parsers already use. It also gets you BOM stripping for free.

## Worth confirming before merge

Any turn whose usage never meets a model-bearing `exec.result` is dropped at EOF with no diagnostic — `pending_by_turn` is simply abandoned when `parse_mcode_file` returns at `mcode.rs:146`. The PR frames this as skipping partial captures, and fail-closed is the right direction since it can only under-report. But it means the parser reports usage for exactly one `turnId` per `exec.result`. If a real `mcode exec` run with several tool-call rounds emits more than one `turnId` before its final result, everything except the matching turn becomes invisible. Have you checked a multi-round capture, not just a single-turn one? If `mcode exec` really is one turn per invocation this is fine as written.

## Docs are incomplete

`CONTRIBUTING.md` step 7 asks for the supported-client overview plus the data-location matrix, mirrored into the maintained translations. The headless section is updated well, but:

- `README.md:56-106` — the main client table every other client appears in has no MiniMax Code row.
- `README.md:1476-1519` — the Windows data-locations matrix has no row either; Codex CLI has one at `README.md:1481`.
- `README.ko.md`, `README.ja.md`, `README.zh-cn.md` are untouched. The existing MiniMax hits in those files are the summarizer and quota provider, not this client.

## Nits

- `crates/tokscale-cli/src/main.rs:6591` — `prepare_headless_args` handles `codex` explicitly and treats *everything else* as mcode. Correct today because `main.rs:6496` only admits two sources, but the next headless client silently inherits mcode's `exec` requirement and `stream-json` injection. A `match source` would fail loudly instead.
- `main.rs:6511` rejects a non-jsonl capture format, but that guard is on tokscale's own `--format`, not on mcode's `--output-format`. `prepare_headless_args` deliberately preserves an explicit `--output-format=json` (asserted at `main.rs:6649-6658`), and that capture parses to zero messages with no warning.
- `crates/tokscale-core/src/clients.rs:966` points the logo at `https://github.com/MiniMax-AI.png`. It resolves (200, `image/jpeg`), and external URLs are allowed per `CONTRIBUTING.md` step 6, so this is fine — just noting `test_hosted_logo_urls_have_a_committed_asset` only guards `tokscale.ai/assets/logos/` URLs, so nothing in CI will catch it if MiniMax renames the org.

## What is right

Worth saying plainly, since these are the parts that normally need a second round.

Token accounting is correct. `mcode.rs:149-157` maps the four disjoint buckets and hardcodes `reasoning: 0`, which is the right call for an additive `TokenBreakdown` — mapping a vendor reasoning count that is already inside output is the usual double-count here. `totalTokens` is correctly not mapped into any bucket, and your own fixture at `mcode.rs:181` shows why the buckets are safe to add: 20468 + 46 + 3651 = 24165 = the reported `totalTokens`, so `inputTokens` excludes cache reads. Negative values are clamped at `mcode.rs:159`.

`cost: 0.0` with the resulting `CostSource::Unknown` matches `dsh.rs:225`, so pricing resolves downstream rather than being pinned to a provider-reported zero.

No `parser_version` arm is needed. `message_cache.rs:1112` falls to `_ => 1`, and the namespace is `client.as_str()`, so `mcode` gets its own version-1 namespace with no collision against `micode`.

Registry wiring is complete and I checked each layer rather than trusting the compiler: `clients.rs:963-971`, scanner discovery at `scanner.rs:1427` and `scanner.rs:1719-1730` (via `headless_roots`, so `TOKSCALE_HEADLESS_DIR` is honoured), both dispatch paths at `lib.rs:1933` and `lib.rs:4781-4789`, the three `ClientFilter` mappings at `main.rs:1095`/`1153`/`1214`/`1271`, hotkey `M` at `client_ui.rs:55` with no collision, and all four frontend registries including `SUPPORTED_CLIENT_TYPES`. The capture filename written at `main.rs:6544` lands under `<root>/mcode/` and matches the `*.jsonl` glob.

Tests assert real values, not just absence of panic — `cli_tests.rs:2158` pins all four buckets across a cold and warm cache pass, and `cli_tests.rs:278` covers the arg injection end to end.

Generalizing `client == ClientId::Codex` to `client.supports_headless()` at `main.rs:4251` is the right cleanup to make alongside this.

CI is green now that the fork workflows were approved: 22/22 passing, including all nine native targets and tests on both ubuntu and windows.

junhoyeo:
@codex review

Focus on `crates/tokscale-core/src/sessions/mcode.rs`: the stream is buffered in `pending_by_turn` and only flushed when the trailing `exec.result` is seen, with no drain at EOF. Please check whether any input shape can strand usage that should have been counted, and whether the new `lossy_lines` read path changes how a malformed line is skipped.

junhoyeo:
Merging. I pushed two commits to your branch rather than sending you back and forth.

`mcode.rs` was reading with `lines().map_while(Result::ok)`, which ends the iteration at the first line that is not valid UTF-8 instead of skipping it — the #1031 failure. It mattered more here than it usually does: usage buffers in `pending_by_turn` and only flushes when the trailing `exec.result` arrives, and nothing drains at EOF, so one stray byte before that result discarded the whole capture rather than truncating a tail. It is now on `utils::lossy_lines`, with a regression test that returns 0 messages under the old spelling and 1 under the new one. That was the only blocking issue.

I also added the MiniMax Code rows to the client table and the data-locations table, which were the two places the docs were still missing it.

The rest held up well under review: the token buckets are disjoint and additive with `reasoning` pinned to 0 and `totalTokens` deliberately not mapped into a bucket, which is the part that usually goes wrong in a new client; the registry wiring is complete at every layer including both dispatch paths and all three filter mappings; and the tests assert real token values across cold and warm cache passes rather than just parsing.

Two things left as follow-ups rather than blockers. A `turnId` that never meets a model-bearing `exec.result` is abandoned at EOF — fail-closed, so it can only under-report, but if a real `mcode exec` can emit more than one turn per capture, everything but the last would be invisible; worth confirming against real output. And `prepare_headless_args` treats every non-Codex source as mcode through an else branch, which will silently mis-handle the next headless client that is added.

Note for the record: Codex could not review this one, the account is over its code-review limit. It merged on a full green CI run plus manual review.

PR #1137, opened 2026-08-16
refactor(core): share SQLite, usage-struct and JSONL drivers across the session parsers

junhoyeo:
Stacked on #1133 — this targets `refactor/opencode-schema-parsers`, so GitHub will retarget it to `main` once that merges. Review the four commits here on their own; the base branch's two commits belong to #1133.

This is a maintainability refactor. The size win is small and that was accepted going in: the whole 50-parser layer is 824 KiB of a 12.7 MiB `.text`, so there is no large lever here. The point is a smaller surface for the next parser to copy-paste from.

## Per-commit byte deltas

Measured with `cargo build --release -p tokscale-cli` in an isolated `CARGO_TARGET_DIR`, one build per commit. The base (#1133 tip) reproduces at 17,028,240 bytes.

| commit | binary | delta |
|---|---|---|
| base (#1133 tip, `765a46c4`) | 17,028,240 | — |
| `refactor(core): share one SQLite row driver...` | 16,995,088 | **-33,152** |
| `refactor(core): share the duplicated usage structs...` | 16,995,072 | **-16** |
| `fix(droid)` + `refactor(core): share one JSONL line driver...` | 16,962,096 | **-32,976** |
| **total** | **16,962,096** | **-66,144** (-0.39%) |

Source: 1,129 lines removed, 842 added across 24 parser files, net **-287 lines** with the two new shared drivers included.

## One SQLite row driver

`utils::sqlite_for_each_row` / `sqlite_for_each_row_on` own open, prepare, execute and per-row-decode failure for every SQLite-backed parser. Each parser keeps only its SQL string and its row-decode closure.

The callback is `&mut dyn FnMut(&Row) -> rusqlite::Result<()>`, not a generic `impl FnMut`: a type parameter would monomorphize the driver once per calling parser and grow the binary, which is the thing this exists to avoid.

The return type is `SqliteScan`, not `bool`, because two callers depend on the distinction. Hermes falls back to coarser session totals only when the per-model query did not run at all, and the OpenCode-schema and ZCode parsers retry an older query only when `prepare` rejected the newer one, never when it prepared and then failed to execute. Parsers that probe optional tables pass `None` for the log label and stay silent, so a missing table is still not reported as a fault.

This collapses the 27 `tracing` callsites in `sessions/` to 11, of which 5 are the driver itself. The remaining 6 are semantic, not boilerplate: Zed payload decode, the Hermes `session_model_usage` probe, Copilot Desktop timestamp parsing.

## Shared usage structs

Six `Deserialize` structs modelled three shapes: `ClaudeUsage` ≡ `AugmentTokenUsage`, `GjcUsage` ≡ `OpenClawUsage`, `GjcCost` ≡ `OpenClawCost`. They become `utils::AnthropicUsage`, `utils::CamelUsage` and `utils::CamelCost`.

Only exact duplicates are merged. The wider relatives of the Anthropic shape (`codex`, `jcode`, `amp`, `droid`, `devin`, `cline`) keep their own structs, and so does `pi::PiUsage`. Folding them in would require `#[serde(alias = ...)]`, and an alias makes a parser start counting a field it deliberately ignores today, which changes reported totals.

Verified zero-widening mechanically: after resolving `rename_all`, per-field `rename` and `alias`, the set of accepted JSON keys of each shared struct is identical to both structs it replaces, so no document can parse differently. An exhaustive pass over all 109 production `Deserialize` structs in `sessions/` found these three pairs and no others, so this is the whole safe surface for the idea.

This one is worth **-16 bytes**. LTO had already inlined those derives into their callers. Reporting it rather than padding it.

## One JSONL line driver, with `claudecode.rs` deliberately excluded

`utils::for_each_json_line` replaces the open/`BufReader`/unwrap/trim/skip-blank prologue in 15 files (16 call sites), handing the sink a trimmed line plus its physical line index. The line is borrowed out of one reused buffer instead of allocated per line, so the driver does strictly less work than `BufReader::lines()`.

**`claudecode.rs` is not migrated, on measurement.** It is the hottest parse path in the product. On a frozen 1.3 GB corpus of real Claude Code transcripts, with the message cache dropped before every run and run order alternated:

| variant | paired median vs base | pairs slower | Wilcoxon p |
|---|---|---|---|
| driver applied to `claudecode.rs` | **+4.5%** (+60 ms) | 29 / 30 | < 0.0001 |
| `claudecode.rs` on `utils::lossy_lines`, no closure | **+6.8%** (+90 ms) | 20 / 20 | 0.0001 |
| `claudecode.rs` left alone (**what this PR ships**) | -0.9% (-12 ms) | 6 / 16 | 0.56 |

The cost is the per-line lossy decode plus the closure boundary; neither alone explains it, and reverting the closure while keeping `lossy_lines` was worse, not better. For a migrated parser with real data, OpenClaw over 138 MB of transcripts measures +1.6% median with p = 0.42 — indistinguishable from noise.

### Correction to the premise behind this item

The audit that motivated this expected a correctness win: 17 JSONL parsers still on raw `BufReader::lines()`, each carrying the #1031 bug where one undecodable byte silently truncated ~98% of an 83 MB Grok log. That is not what those parsers do. #1031 came from `map_while(Result::ok)`, which turns a decode error into end-of-iteration. Every parser here spelled it `Err(_) => continue`, which drops only the offending line and keeps reading — `read_line` consumes the bytes before it validates them, so the iterator resumes on the next line.

Verified directly, on a transcript whose middle record carries a raw `0xFF` byte: before this change tokscale reports the two clean records (500 input tokens, 2 messages); after it reports all three (700 input tokens, 3 messages). So the driver recovers that one line and strips a leading BOM. It does not rescue the rest of a file, because the rest of the file was never lost.

The one genuine instance of the #1031 spelling was `droid.rs`, which wrote `line.ok()?` and therefore abandoned model extraction for the whole file at the first bad byte, leaving the session on a `<provider>-unknown` model id. That is fixed in its own commit.

## Gates

- `cargo test --release -p tokscale-core -p tokscale-cli` — **3,099 passed, 0 failed**, tests unmodified.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.

Behaviour was additionally checked end to end by replaying the report over a frozen copy of real local session data before and after: **590 sessions and 29 client/provider/model groups identical** on input, output, cacheRead, cacheWrite, reasoning, messageCount and cost. The `prime-agent` client is excluded from that comparison because it is live on this machine and grows between runs; a reversed-order replay showed zero differences at all.

## Not done

`sessions/pi.rs` and `sessions/prime_agent.rs` are untouched — #1134 is in flight against both. `pi.rs` therefore keeps its own `PiUsage` rather than joining `CamelUsage`, and neither file was moved onto either driver.

---
## Summary by cubic
Shares SQLite, JSONL, and usage-struct drivers across session parsers to reduce boilerplate and binary size while preserving reported metrics and the historical `claudecode::*` public API. Per-parser loops are replaced with shared helpers; behavior is unchanged except `droid` no longer aborts on a bad byte and JSONL readers strip a BOM.

- SQLite: replace per-parser prepare/execute loops with `sessions::utils::sqlite_for_each_row` and `sqlite_for_each_row_on`. Callbacks use `&mut dyn FnMut` to avoid monomorphization. Return `SqliteScan` so Hermes and OpenCode/ZCode can distinguish prepare vs execute failures. Optional table probes pass `None` to stay quiet.
- JSONL: adopt `sessions::utils::for_each_json_line` to reuse a buffer, strip BOM, and keep reading past undecodable bytes. Do not migrate `claudecode.rs` (measured regression on hot path). Fix `droid.rs` to iterate lossy lines so one bad byte no longer aborts model extraction.
- Usage structs: deduplicate identical shapes into `sessions::utils::AnthropicUsage`, `CamelUsage`, and `CamelCost` with no key widening. Re-export `sessions::utils::AnthropicUsage` as `claudecode::ClaudeUsage`; keep `ClaudeEntry`/`ClaudeMessage` public to preserve the crate’s API.
- Gates: tests unchanged and passing; end-to-end session counts, token totals, and costs match. Release binary: -66,144 bytes (-0.39%).

<sup>Written for commit a750bd25942f6eba21230c3e6f4058537c901aaf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1137?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/antigravity_cli.rs
+++ b/crates/tokscale-core/src/sessions/antigravity_cli.rs
@@ -41,7 +41,7 @@
 //! server-supplied name that gets renamed (`Gemini 3 Flash` → `Gemini 3.5 Flash
 //! (High)`) and could be localized.
 
-use super::utils::open_readonly_sqlite_opt;
+use super::utils::{open_readonly_sqlite_opt, sqlite_for_each_row_on};
 use super::{normalize_workspace_key, workspace_label_from_key, UnifiedMessage};
 use crate::{pricing, provider_identity, TokenBreakdown};
 use rusqlite::Connection;
@@ -61,20 +61,23 @@ pub fn parse_antigravity_cli_file(path: &Path) -> Vec<UnifiedMessage> {
 
     let (timestamp, workspace_key, workspace_label) = read_trajectory_meta(&conn, path);
 
-    let mut stmt = match conn.prepare("SELECT data FROM gen_metadata ORDER BY idx") {
-        Ok(stmt) => stmt,
-        // Not an Antigravity CLI database (table missing) — nothing to count.
-        Err(_) => return Vec::new(),
-    };
-    let rows = match stmt.query_map([], |row| row.get::<_, Vec<u8>>(0)) {
-        Ok(rows) => rows,
-        Err(_) => return Vec::new(),
-    };
-
     // Buffered rather than streamed so a row missing its own `#19` can borrow
     // attribution from anywhere in the conversation, not just from rows that
     // happen to precede it.
-    let blobs: Vec<Vec<u8>> = rows.flatten().collect();
+    //
+    // Quiet: a database without `gen_metadata` is not an Antigravity CLI
+    // database at all, so there is nothing to warn about.
+    let mut blobs: Vec<Vec<u8>> = Vec::new();
+    sqlite_for_each_row_on(
+        &conn,
+        path,
+        "SELECT data FROM gen_metadata ORDER BY idx",
+        None,
+        &mut |row| {
+            blobs.push(row.get::<_, Vec<u8>>(0)?);
+            Ok(())
+        },
+    );
     let session_models = SessionModels::from_blobs(&blobs);
 
     let mut messages = Vec::new();
--- a/crates/tokscale-core/src/sessions/augment.rs
+++ b/crates/tokscale-core/src/sessions/augment.rs
@@ -36,9 +36,11 @@
 //! always set. If a future format needs multiple messages per turn, revisit
 //! that flag before counting turns.
 
-use super::utils::{file_modified_timestamp_ms, parse_timestamp_str, read_file_or_none};
+use super::utils::{
+    file_modified_timestamp_ms, parse_timestamp_str, read_file_or_none, AnthropicUsage,
+};
 use super::UnifiedMessage;
-use crate::{provider_identity, TokenBreakdown};
+use crate::provider_identity;
 use serde::Deserialize;
 use std::path::Path;
 
@@ -83,15 +85,7 @@ struct AugmentExchange {
 /// do not fail the turn.
 #[derive(Debug, Deserialize)]
 struct AugmentResponseNode {
-    token_usage: Option<AugmentTokenUsage>,
-}
-
-#[derive(Debug, Deserialize)]
-struct AugmentTokenUsage {
-    input_tokens: Option<i64>,
-    output_tokens: Option<i64>,
-    cache_read_input_tokens: Option<i64>,
-    cache_creation_input_tokens: Option<i64>,
+    token_usage: Option<AnthropicUsage>,
 }
 
 fn model_id(model: Option<&str>) -> String {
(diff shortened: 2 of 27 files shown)


junhoyeo:
Good catch — fixed in 845d83b1.

`tokscale-core` is not published to crates.io (the sparse index returns 404; the release pipeline ships a compiled binary through npm, and `tokscale-cli` is the only consumer of the rlib), so nothing could actually break today. But the paths were public and dropping them was gratuitous, so they are back:

- `ClaudeEntry` and `ClaudeMessage` are `pub` again.
- `claudecode` re-exports the shared block as `pub use super::utils::AnthropicUsage as ClaudeUsage;`, so `tokscale_core::sessions::claudecode::ClaudeUsage` resolves to the same four fields.
- `sessions::utils` stays crate-private; only `AnthropicUsage` and its fields are `pub`, which is what the re-export requires.

Verified rather than assumed, by compiling a consumer against the crate:

```rust
fn _probe(entry: &tokscale_core::sessions::claudecode::ClaudeEntry) -> Option<i64> {
    let msg: &tokscale_core::sessions::claudecode::ClaudeMessage = entry.message.as_ref()?;
    let usage: &tokscale_core::sessions::claudecode::ClaudeUsage = msg.usage.as_ref()?;
    usage.input_tokens
}
```

That builds on this branch and does not build on the previous commit. Gates re-run on the new tip: 3,099 tests pass, clippy `-D warnings` clean, `cargo fmt --check` clean, release binary unchanged at 16,962,096 bytes.

PR #1138, opened 2026-08-16
perf: replace vendored OpenSSL with rustls

junhoyeo:
## What

`reqwest` was configured with `native-tls-vendored`, which compiles and statically links a full OpenSSL (`openssl-src v300.5.5+3.5.5`) into every **Linux** artifact. macOS and Windows never paid for it — `nm` finds zero `SSL_`/`EVP_` symbols in the darwin binary, because native-tls delegates to Security.framework there and schannel on Windows. The cost fell entirely on the platform most people install from:

| published package (v4.13.0) | unpacked |
|---|---:|
| `@tokscale/cli-linux-x64-gnu` | 25,847,376 |
| `@tokscale/cli-linux-x64-musl` | 24,241,010 |
| `@tokscale/cli-win32-x64-msvc` | 22,953,994 |
| `@tokscale/cli-linux-arm64-gnu` | 22,319,648 |
| `@tokscale/cli-darwin-arm64` | 17,005,619 |

Switching to `rustls-tls-native-roots` removes `openssl-src`, `openssl-sys`, `openssl`, `native-tls` and `tokio-native-tls` from the Linux graph, replacing them with `rustls` + `ring` + `rustls-native-certs`.

**`rustls-tls-native-roots`, not plain `rustls-tls`** — plain `rustls-tls` trusts only the webpki root set baked into the binary, which silently breaks anyone behind a corporate MITM proxy or using a private CA. `rustls-native-certs` reads the OS trust store, so certificate acceptance is unchanged.

## This is a platform trade, not a pure win

The direction reverses by platform. macOS grows, because rustls and ring are real Rust and assembly compiled in where native-tls was a thin shim over an OS framework:

| | before | after | delta |
|---|---:|---:|---:|
| stripped `aarch64-apple-darwin` | 17,061,280 | 18,052,848 | **+969 KiB** |

Linux is where the win is. Per-target numbers from this PR's own `build-native` run are in a comment below once CI finishes.

## The blocker I had to clear first

`antigravity_https_client` (`crates/tokscale-cli/src/antigravity.rs`) is the only place in the tree that configures TLS directly. It talks to a **loopback** language server over `https://127.0.0.1:<port>` with `danger_accept_invalid_certs(true)`. rustls is stricter than OpenSSL about loopback certificates — it rejects certificates with no `subjectAltName` outright — so this had to be proven rather than assumed.

Reproduced that exact client configuration against `openssl s_server`:

| server certificate | result |
|---|---|
| self-signed, IP SAN `127.0.0.1` (antigravity's shape) | **200 OK** |
| self-signed, **no SAN at all** (rustls' strictest case) | **200 OK** |
| nothing listening (control) | fails fast — `Connection refused (os error 61)`, no hang |

`danger_accept_invalid_certs` bypasses the verifier under rustls exactly as it did under native-tls. A grep for `danger_accept_invalid_hostnames`, `add_root_certificate`, `Identity::`, `native_tls`, `tls_built_in_root_certs` finds no other TLS configuration anywhere in the tree.

## Verification

- `cargo test --release -p tokscale-core -p tokscale-cli` — 0 failures, tests unmodified.
- `cargo clippy --all-targets -- -D warnings` — clean. `cargo fmt --check` — clean.
- `cargo tree --target x86_64-unknown-linux-gnu` before/after: `openssl-src v300.5.5+3.5.5`, `openssl-sys`, `openssl`, `native-tls`, `tokio-native-tls` → `rustls 0.23.43`, `ring 0.17.14`, `rustls-native-certs 0.8.4`, `hyper-rustls`, `tokio-rustls`.
- **Live HTTPS with the built binary**: `tokscale wrapped` fetches provider logos from tokscale.ai, raw.githubusercontent.com and github.com and renders them into the output PNG, and the models.dev pricing cache is refetched and rewritten to disk.

## Note

Conflicts with #1132 on `Cargo.lock` only (that PR trims `crates/tokscale-cli/Cargo.toml` dependency features; this one changes root `Cargo.toml`). Mechanically resolvable by regenerating the lockfile after whichever lands first.

---
## Summary by cubic
Replaces `reqwest`’s `native-tls-vendored` OpenSSL with `rustls-tls-native-roots` to drop vendored OpenSSL from Linux builds and shrink artifacts, while keeping OS trust store behavior. macOS binaries grow slightly because `rustls`/`ring` are compiled in.

- Dependency graph: removes `openssl-src`, `openssl-sys`, `openssl`, `native-tls`, `tokio-native-tls`, `hyper-tls`; adds `rustls`, `ring`, `rustls-native-certs`, `hyper-rustls`, `tokio-rustls`, `quinn`.
- Behavior: trusts the OS store; corporate MITM/private CAs continue to work. The loopback client with `danger_accept_invalid_certs(true)` still connects; no other TLS configuration exists in the tree.
- Size: Linux artifacts shrink; `aarch64-apple-darwin` grows by ~969 KiB.
- Rollout: no configuration or code changes required.

<sup>Written for commit c7b2fd6dd634f4192d294be2fdff65c556569729. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1138?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,7 +32,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -315,6 +315,23 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
+
+[[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -453,9 +470,9 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
-version = "0.9.4"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
 dependencies = [
  "core-foundation-sys",
  "libc",
@@ -472,6 +489,15 @@ name = "cpufeatures"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
 dependencies = [
  "libc",
 ]
@@ -805,21 +831,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
 
 [[package]]
-name = "foreign-types"
-version = "0.3.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
-dependencies = [
- "foreign-types-shared",
-]
-
-[[package]]
-name = "foreign-types-shared"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
-
-[[package]]
 name = "form_urlencoded"
 version = "1.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -924,8 +935,22 @@ checksum = "899def5c37c4fd7b2664648c28120ecec138e4d395b459e5ca34f9cce2dd77fd"
 dependencies = [
  "cfg-if",
  "libc",
- "r-efi",
+ "r-efi 5.3.0",
  "wasip2",
+]
+
+[[package]]
+name = "getrandom"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "300e883d756b2e4ec94e02791f39b04b522276138852cfc41d9fb7e904106099"
+dependencies = [
+ "cfg-if",
+ "js-sys",
+ "libc",
+ "r-efi 6.0.0",
+ "rand_core 0.10.1",
+ "wasm-bindgen",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


junhoyeo:
## Per-target measurements from CI

Sizes are the `build-native` artifact sizes (zip-compressed, so not raw binary bytes — but built by the same workflow on the same runners, so comparable run-to-run). Baseline is #1134's run, which changes only `sessions/*.rs` and so has main's exact dependency graph.

| target | baseline | with rustls | delta | % | with #1132 too | % |
|---|---:|---:|---:|---:|---:|---:|
| `aarch64-apple-darwin` | 7,931,054 | 8,526,969 | +595,915 | +7.51% | -238,560 | -3.01% |
| `aarch64-linux-android` | 10,366,239 | 8,677,144 | -1,689,095 | -16.29% | -2,560,304 | -24.70% |
| `aarch64-pc-windows-msvc` | 8,277,874 | 8,891,245 | +613,371 | +7.41% | -229,582 | -2.77% |
| `aarch64-unknown-linux-gnu` | 10,419,035 | 8,706,186 | -1,712,849 | -16.44% | -2,586,911 | -24.83% |
| `aarch64-unknown-linux-musl` | 10,323,177 | 8,681,904 | -1,641,273 | -15.90% | -2,512,101 | -24.33% |
| `x86_64-apple-darwin` | 8,342,763 | 8,975,548 | +632,785 | +7.58% | -228,406 | -2.74% |
| `x86_64-pc-windows-msvc` | 9,011,533 | 9,662,071 | +650,538 | +7.22% | -231,695 | -2.57% |
| `x86_64-unknown-linux-gnu` | 10,650,741 | 9,296,121 | -1,354,620 | -12.72% | -2,244,728 | -21.08% |
| `x86_64-unknown-linux-musl` | 10,533,722 | 9,263,806 | -1,269,916 | -12.06% | -2,158,312 | -20.49% |

## Read this together with #1132

**On its own, this PR is a trade, not a win.** Linux and Android drop 12–16%; macOS and Windows grow 7.2–7.6%, because rustls and ring are compiled into the binary where native-tls was a thin shim over Security.framework / schannel.

**Landed alongside #1132, every target improves** — the last two columns. Linux and Android fall 20–25%, and macOS and Windows still come out 2.6–3.0% ahead, because #1132's codec and font-stack trims are worth ~840–890 KiB on every platform and more than absorb the TLS regression.

So the recommendation is to land both or neither. Landing this one alone would ship a measurable regression to every macOS and Windows user to benefit Linux; landing both is a strict improvement everywhere, concentrated on the platform that is currently worst off.

Method note: the combined column adds the two measured deltas. The changes are disjoint (this PR touches root `Cargo.toml`'s `reqwest` entry, #1132 touches `crates/tokscale-cli/Cargo.toml` image/SVG features) so they should be very close to additive, but the exact combined figure will be confirmed once one of them lands and the other rebases.

junhoyeo:
@codex review

This is a one-line dependency change with a disproportionate blast radius: it swaps the TLS stack for every HTTPS call the CLI makes. Two things worth scrutinising in particular.

First, `antigravity_https_client` in `crates/tokscale-cli/src/antigravity.rs` talks to a loopback language server with `danger_accept_invalid_certs(true)`, and rustls is stricter than OpenSSL about loopback certificates. I tested that path against `openssl s_server` and it returns 200 OK for both an IP-SAN and a no-SAN self-signed certificate, but please check whether any other call site makes assumptions about the TLS backend that I missed.

Second, `rustls-tls-native-roots` was chosen over plain `rustls-tls` so the OS trust store is still consulted. Please confirm that reading is right for reqwest 0.12 and that nothing else in the tree re-enables `native-tls` transitively.

junhoyeo:
## Rebased onto `main` after #1132 landed

The `Cargo.lock` conflict is resolved: I took main's lockfile (which carries #1132's codec and font-stack trims) and let cargo re-resolve it against the rustls manifest, so both changes are present rather than one overwriting the other. Verified mechanically — `openssl-src` 0 entries, `ravif` 0 entries, `rustls` present.

**This also settles the platform question with a real combined build rather than an estimate.** Measured locally on stripped aarch64-apple-darwin:

| | binary | vs baseline |
|---|---:|---:|
| baseline `f169a1f0` | 17,061,280 | — |
| rustls alone (pre-rebase) | 18,052,848 | +5.81% |
| **#1132 + rustls (this tip)** | **16,233,600** | **−4.85%** |

So the macOS regression this PR showed in isolation is more than absorbed now that #1132 is in `main`: macOS ends up 4.85% smaller than before either change, while Linux and Android keep the 12–16% drop from dropping vendored OpenSSL. Every platform improves.

Gates re-run on the rebased tip: `cargo test --release -p tokscale-core -p tokscale-cli` 0 failures, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

PR #1140, opened 2026-08-17
fix(pricing): resolve OpenAI Daybreak Blue alias

### LINKED ISSUE #1114: `openai/gpt-daybreak-blue-latest` usage is excluded from submission as unpriced
## Summary

`tokscale submit` detects Codex usage recorded under `openai/gpt-daybreak-blue-latest`, but excludes it because no authoritative model-to-price mapping is available.

## Reproduction

```sh
bunx tokscale@latest submit


Tokscale - Submit Usage Data

Warning: excluded 000 unpriced openai/gpt-daybreak-blue-latest message(s) (000 tokens): no authoritative model-to-price mapping. Remaining priced usage will be submitted.

Submitting to server...

Successfully submitted!
```

## Related

- #1035 — previous reports of unpriced model IDs during submission
- #1053 — changed submission behavior to exclude unpriced usage and submit the priced remainder

aravhawk:
## Summary
- map OpenAI’s Daybreak Blue aliases to the documented `gpt-5.6-sol` target
- support API, Codex, and provider-qualified model IDs
- add regression coverage for pricing lookup, token-bucket coverage, and cost calculation

## Testing
- `cargo test -p tokscale-core pricing::aliases::tests`
- `cargo fmt --all --check`

Closes #1114.

---
## Summary by cubic
Resolves all OpenAI Daybreak Blue model IDs to `gpt-5.6-sol` so pricing lookup, token-bucket rates, and cost calculation consistently use the GPT-5.6 Sol row. Previously, provider-qualified `openai/daybreak-blue-latest` (and some aliases) could resolve to none and be unpriced; now all covered.

**Review notes**
- Adds aliases for `daybreak-blue-latest`, `gpt-daybreak-blue-latest`, `openai/daybreak-blue-latest`, and `openai/gpt-daybreak-blue-latest`; resolution is case-insensitive.
- Pins all aliases to `gpt-5.6-sol`; provider-qualified forms are explicit because prefix stripping does not re-run alias resolution.
- Extends `tokscale-core` tests to verify pricing lookup, token-bucket coverage, and cost for all four spellings via `lookup_with_source_and_provider` and `calculate_cost_with_provider`.

<sup>Written for commit f3296218f060f3deebabea48ad74f021f5a624b7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1140?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/pricing/aliases.rs
+++ b/crates/tokscale-core/src/pricing/aliases.rs
@@ -133,13 +133,16 @@ static MODEL_ALIASES: Lazy<HashMap<&'static str, &'static str>> = Lazy::new(|| {
     m.insert("grok-composer-2.5-fast", "composer-2.5-fast");
     // OpenAI documents the API spelling below as a moving alias for
     // `gpt-5.6-sol`; Codex records the same alias with its `gpt-` prefix.
-    // Keep both spellings pinned to the currently documented target so the
-    // upstream GPT-5.6 Sol row supplies all token-bucket rates.
+    // Keep the API, Codex, and provider-qualified spellings pinned to the
+    // currently documented target so the upstream GPT-5.6 Sol row supplies
+    // all token-bucket rates. The qualified form must be explicit because
+    // provider-prefix stripping does not run alias resolution a second time.
     // Sources (accessed 2026-08-17):
     // https://developers.openai.com/api/docs/guides/safety-checks/cybersecurity
     // https://developers.openai.com/api/docs/pricing
     m.insert("daybreak-blue-latest", "gpt-5.6-sol");
     m.insert("gpt-daybreak-blue-latest", "gpt-5.6-sol");
+    m.insert("openai/gpt-daybreak-blue-latest", "gpt-5.6-sol");
 
     // Synthetic model variants (only where resolver needs help)
     m.insert("kimi-k2.5-nvfp4", "kimi-k2.5"); // Quantization variant → base model pricing
@@ -229,6 +232,10 @@ mod tests {
             resolve_alias("GPT-DAYBREAK-BLUE-LATEST"),
             Some("gpt-5.6-sol")
         );
+        assert_eq!(
+            resolve_alias("openai/gpt-daybreak-blue-latest"),
+            Some("gpt-5.6-sol")
+        );
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


junhoyeo:
Mergeable. The target is correct, and the test asserts resolved cost rather than the alias string, which is the right bar for this table. One asymmetry is worth adding first since it is a one-line change.

I checked the target against both pages you cite, not just the PR description. `developers.openai.com/api/docs/pricing` says "`daybreak-blue-latest` and `daybreak-red-latest` are aliases that currently point to `gpt-5.6-sol` and `gpt-5.6-cyber`, respectively", and the cybersecurity guide says "For approved API projects, `gpt-daybreak-blue-latest` resolves to `gpt-5.6-sol`". So `crates/tokscale-core/src/pricing/aliases.rs:143-145` names the correct row and not a same-family neighbour, which was the main thing to get wrong here.

The reasoning in your comment at `aliases.rs:138-139` is accurate. `resolve_alias` runs exactly once, at `crates/tokscale-core/src/pricing/lookup.rs:522`, before the generic prefix fallback at `lookup.rs:619` and the unknown-prefix stripping at `lookup.rs:650`, so the provider-qualified spelling really does need its own key. Case-insensitivity comes free from the `to_lowercase` at `aliases.rs:154`, consistent with the neighbouring entries.

I also checked the part that actually moves money for this specific model. `gpt-5.6-sol` is in `is_openai_full_request_272k_model` at `lookup.rs:1605`, and both the tiered-source preference at `lookup.rs:722` and `uses_openai_full_request_272k_pricing` at `lookup.rs:1661` read the alias-resolved id rather than the raw one. Against a LiteLLM row carrying above-272k tiers and a flat OpenRouter row, all three new spellings elect the LiteLLM tiered row and produce the same above-272k cost as a direct `gpt-5.6-sol` lookup. No silent tier loss through the alias.

The gap is that the bare API spelling is not covered in provider-qualified form:

```
daybreak-blue-latest             -> Some("gpt-5.6-sol")
gpt-daybreak-blue-latest         -> Some("gpt-5.6-sol")
openai/gpt-daybreak-blue-latest  -> Some("gpt-5.6-sol")
openai/daybreak-blue-latest      -> None
```

`aliases.rs:143` adds the bare API id and `aliases.rs:145` adds the qualified `gpt-` form, but not the qualified bare form, and by exactly the reasoning your own comment gives it cannot resolve on its own. `m.insert("openai/daybreak-blue-latest", "gpt-5.6-sol");` makes the four spellings symmetric.

Two smaller notes, neither blocking:

- `daybreak-red-latest` and `gpt-daybreak-red-latest` resolve to `gpt-5.6-cyber` in the same documented paragraph and are unpriced today. Out of scope for a PR titled for Blue, but worth a follow-up.
- These are moving aliases by design; the pricing page states they will be repointed as new Daybreak models ship, with pricing adjusted to match. The `accessed 2026-08-17` note is the right instinct, but nothing fails when OpenAI repoints them, and unlike Anthropic's `-0` aliases at `aliases.rs:108-109`, which stay within a model line, this one can jump across generations. Not something to solve in this PR.

On size: of the +71, three lines are table entries, nine are the sourced comment, and 59 are tests. Nothing is duplicating an existing table.

On CI: the workflows had been sitting on `action_required` because this is a fork PR, so nothing had actually been compiled or tested until now. They have since run for real and are green, including all nine cross-compile builds and both test jobs. Locally on the PR head, `cargo test -p tokscale-core --lib` passes 1782/1782.

junhoyeo:
Merging. Good change, and the sourcing is right — I checked both OpenAI pages you cited rather than taking the mapping on trust, and `daybreak-blue-latest` does currently resolve to `gpt-5.6-sol`. The tiered pricing also survives the alias: `gpt-5.6-sol` is in the above-272k set, and both the tiered-source preference and the 272k check read the alias-resolved id, so a resolved spelling elects the same tiered row and produces the same cost as a direct lookup.

I pushed one commit to your branch. The table had `daybreak-blue-latest`, `gpt-daybreak-blue-latest` and `openai/gpt-daybreak-blue-latest`, but not `openai/daybreak-blue-latest` — and by the reasoning in your own comment, that prefix stripping does not re-run alias resolution, that spelling could not fall back to its bare form, so it resolved to `None`. I also widened the cost test to iterate all four spellings through `calculate_cost_with_provider` instead of two, so each is asserted against real rates rather than as a string mapping.

`daybreak-red-latest` and `gpt-daybreak-red-latest` point at `gpt-5.6-cyber` in the same documented paragraph and are still unpriced. Out of scope for a Blue-titled PR; I will pick those up separately.

Worth noting these are moving aliases that OpenAI says will be repointed, and unlike the Anthropic `-0` aliases these can jump generations rather than staying in-family. Nothing breaks when it happens, it just needs revisiting.

PR #1143, opened 2026-08-18
refactor(core): put the Pi-format parsers on the shared session drivers

junhoyeo:
`sessions/pi.rs` and `sessions/prime_agent.rs` were the two parsers left out of #1137, so they were still carrying their own line plumbing. This puts what actually fits onto the shared drivers in `sessions/utils.rs` and documents, with evidence, the two things that do not fit.

## What moved

- **One line driver for the Pi family.** Pi had a two-variant `PiLines`/`PiLine` iterator wrapping either `BufRead::lines()` (Pi, Senpi, Kimchi) or `utils::lossy_lines_with_bytes` (Prime Agent), each with its own BOM handling, and Prime Agent's accounting walk opened and iterated the same file a third way. All three now go through `utils::for_each_json_line_with_bytes`, which shares `read_line_payload` with `for_each_json_line`, so terminator and BOM handling exist once. `lossy_lines_with_bytes`, `LossyLinesWithBytes` and `LossyLine` had no other callers and are deleted.
- **One JSON line decode helper.** `prime_agent::parse_pi_json_line` becomes `utils::parse_json_line`, replacing the three open-coded copies of "clear the buffer, copy the line in, hand it to simd-json" in `pi.rs`.
- **One token breakdown.** `PiUsage::to_breakdown` replaces the inline breakdown at the Pi emit site and `prime_agent::usage_breakdown`, matching `AnthropicUsage::to_breakdown` and `CamelUsage::to_breakdown`.

The driver hands the sink borrowed slices out of its own buffer, so a scan no longer allocates a `String` per line for Pi, or a `String` plus a `Vec<u8>` per line for Prime Agent.

## What did not move, and why

**`for_each_json_line` cannot serve either Pi reader mode.** It hands out only a lossily decoded `&str`, and both halves of the family need more than that, which is why the new driver is a sibling rather than a migration onto it.

- Prime Agent hashes the exact source bytes of a damaged record into `prime-agent:damaged:<sha256>`, and detects a lineage key mangled by invalid UTF-8 by testing those bytes with `str::from_utf8`. Both are gone after a lossy decode: distinct invalid sequences have already collapsed onto one U+FFFD. `rejects_only_headers_with_damaged_lineage_keys` pins exactly this, requiring a raw `0xFF` appended to a complete `parentSession` key to be rejected while the valid-UTF-8 spelling `"parentSession\u{FFFD}"` is accepted.
- Pi, Senpi and Kimchi drop a record whose bytes are not valid UTF-8 rather than reading it through replacement characters. Decoding lossily makes them emit messages they do not emit today. Measured on a record whose model carries one `0xFF` byte, on `main` and on this branch alike:

  ```
  strict=[]  lossy=[("unknown", 28)]
  ```

  Pi skips it; Prime Agent emits it as model `unknown`. Opting the first three clients into the second behaviour needs a parser-version bump for each, which `PiParseOptions::standard` already says it is deferring, so the driver reports `valid_utf8` and they keep skipping.

**`PiUsage` does not fold into `CamelUsage`.** The accepted key sets are not identical: `PiUsage` reads `reasoning` and routes every other key into the flattened extras map `has_damaged_key` needs, while `CamelUsage` reads `cost`. A struct holding the union of both rejects documents that each accepts today (simd-json, the real decode path):

  | document | `PiUsage` | `CamelUsage` | union |
  | --- | --- | --- | --- |
  | real Pi usage block | ok | ok | ok |
  | `"cost": 5` | **ok** | fail | **fail** |
  | `"cost": "free"` | **ok** | fail | **fail** |
  | `"cost": {"total": "x"}` | **ok** | fail | **fail** |
  | `"reasoning": "a lot"` | fail | **ok** | **fail** |
  | `"reasoning": 5` | ok | ok | ok |

  A failed usage block is not a lost cost field, it is a lost message: `pi_emitted_record` requires `message.usage`. And `cost` is on every Pi usage block in the local corpus (18,641 of 18,641 sampled), so the narrowing would sit on the common path. The two stay separate and the note on `CamelUsage` now says so with the reason.

## Verification

Behaviour was replayed over a frozen 1.5 GB copy of the real local corpus, 1,049 transcripts from `~/.prime/agent/sessions`, `~/.prime/agent/session-artifacts` and `~/.senpi`, each parsed as Pi, Senpi, Kimchi and Prime Agent: 371,860 emitted messages compared field by field (session id, model, provider, timestamp, all five token buckets, dedup key, agent, workspace) plus every file's accounting record.

Output is byte-identical to `main` on 372,879 of 372,909 lines. The 30 that differ are accounting records whose attribution order comes from `HashMap` iteration; running the unmodified `main` binary twice permutes the same 30 lines, and their contents are identical as multisets. That nondeterminism predates this branch.

No `parser_version` changed.

## Gates

```
cargo test --release -p tokscale-core -p tokscale-cli
     Running unittests src/bin/fake_codex.rs (target/release/deps/fake_codex-d06643f8e460009c)
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running unittests src/main.rs (target/release/deps/tokscale-e9b2f4323747cf84)
test result: ok. 1111 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 9.00s
     Running tests/cli_tests.rs (target/release/deps/cli_tests-fb621575f1c08b58)
test result: ok. 167 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 11.34s
     Running unittests src/lib.rs (target/release/deps/tokscale_core-b2cf759b19043090)
test result: ok. 1784 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 8.41s
     Running tests/anthropic_catalog.rs (target/release/deps/anthropic_catalog-7bd6e4a8705be7bd)
test result: ok. 0 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/codebuff.rs (target/release/deps/codebuff-7fc36a2c56b4939e)
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
     Running tests/freebuff.rs (target/release/deps/freebuff-acbb540750cb289b)
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
     Running tests/gjc.rs (target/release/deps/gjc-b0279e7f0f2696a4)
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
     Running tests/hermes.rs (target/release/deps/hermes-a7a19f91710f2c18)
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
     Running tests/jcode.rs (target/release/deps/jcode-48e393f87d4e2dcc)
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
     Running tests/junie.rs (target/release/deps/junie-efa40852b91f3542)
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
   Doc-tests tokscale_core
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

3,104 passed, 0 failed, 5 ignored across 12 suites. Every existing test is unmodified. The known `trae::sync` contender flake did not appear.

cargo clippy --all-targets -- -D warnings
Checking tokscale-core v4.13.0 (/Users/junhoyeo/tokscale-worktrees/pi-drivers/crates/tokscale-core)
    Checking tokscale-cli v4.13.0 (/Users/junhoyeo/tokscale-worktrees/pi-drivers/crates/tokscale-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 7.08s
(exit 0, no warnings)

cargo fmt --check
(exit 0, no output)
```

## Binary size

Stripped release binary, `cargo build --release -p tokscale-cli`, measured per commit:

| commit | bytes | delta |
| --- | --- | --- |
| `main` (1be9f19f) | 16,167,488 | baseline |
| share one token breakdown | 16,167,488 | 0 |
| decode JSON lines through one helper | 16,167,488 | 0 |
| read transcripts through a shared line driver | 16,167,504 | **+16** |
| document why the usage blocks stay separate | 16,167,504 | 0 |

Net +16 bytes. LTO had already merged the duplicated breakdowns and decode sites, so the source dedup buys nothing there, and the driver costs 16 bytes rather than saving any. This is a maintainability and allocation change, not a size win; the earlier round's equivalent item was worth -16 bytes and this one is worth +16.

---
## Summary by cubic
Unifies Pi-format transcript parsing on shared drivers to remove duplication and per-line allocations without changing behavior. Pi and Prime Agent now use one line driver and one JSON decode helper; outputs match `main`.

- Review notes:
  - New `sessions/utils.rs::for_each_json_line_with_bytes` drives both Pi and Prime Agent. It exposes raw bytes, trimmed text, and `valid_utf8`, and allows early stop via `ControlFlow`. BOM and terminator handling live here.
  - New `sessions/utils.rs::parse_json_line` replaces open-coded simd-json calls.
  - `PiUsage::to_breakdown` centralizes token breakdown. Prime accounting now uses it.
  - Kept behavior: Pi/Senpi/Kimchi still drop invalid-UTF-8 records; Prime Agent still inspects and hashes raw bytes. `for_each_json_line` cannot serve these needs and remains separate.
  - Deleted `lossy_lines_with_bytes`, `LossyLinesWithBytes`, and `LossyLine` (no remaining callers).
  - Verified on a 1.5 GB corpus: messages identical; only nondeterministic `HashMap` attribution order differs in 30 accounting lines. No `parser_version` change. No migrations required.

<sup>Written for commit ae638d547777dfd52a0c31e0f0188e1853380303. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1143?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/sessions/pi.rs
+++ b/crates/tokscale-core/src/sessions/pi.rs
@@ -235,8 +235,8 @@ pub struct PiUsage {
     pub cache_write: Option<i64>,
     #[allow(dead_code)]
     pub total_tokens: Option<i64>,
-    /// Parsed so the omission below is a real decision rather than an accident
-    /// of the schema, but never summed: see the note at the emit site.
+    /// Parsed so the omission in [`PiUsage::to_breakdown`] is a real decision
+    /// rather than an accident of the schema, but never summed.
     #[allow(dead_code)]
     pub reasoning: Option<i64>,
     #[serde(flatten)]
@@ -244,6 +244,24 @@ pub struct PiUsage {
 }
 
 impl PiUsage {
+    /// Token breakdown with every field clamped at zero, in the spelling
+    /// `utils::CamelUsage` uses for the same wire shape.
+    ///
+    /// `reasoning` is read but deliberately not mapped onto
+    /// `TokenBreakdown::reasoning`. In the Pi format reasoning tokens are a
+    /// subset of `output` (Pi's own `totalTokens` excludes them), whereas
+    /// tokscale totals `reasoning` as its own additive bucket. Mapping it
+    /// through would double count.
+    pub(crate) fn to_breakdown(&self) -> TokenBreakdown {
+        TokenBreakdown {
+            input: self.input.unwrap_or(0).max(0),
+            output: self.output.unwrap_or(0).max(0),
+            cache_read: self.cache_read.unwrap_or(0).max(0),
+            cache_write: self.cache_write.unwrap_or(0).max(0),
+            reasoning: 0,
+        }
+    }
+
     pub(crate) fn has_damaged_key(&self) -> bool {
         const TOKEN_COUNTER_KEYS: &[&str] = &[
             "input",
@@ -726,24 +744,13 @@ fn parse_pi_format_file_inner(
             .map(|timestamp| timestamp.timestamp_millis());
         let timestamp = recorded_timestamp.unwrap_or(fallback_timestamp);
 
-        // `usage.reasoning` is read but deliberately not mapped onto
-        // `TokenBreakdown::reasoning`. In the Pi format reasoning tokens are a
-        // subset of `output` (Pi's own `totalTokens` excludes them), whereas
-        // tokscale totals `reasoning` as its own additive bucket. Mapping it
-        // through would double count.
         let mut unified = UnifiedMessage::new_with_agent(
             client,
             model,
             provider.as_str(),
             session_id.clone().unwrap_or_else(|| "unknown".to_string()),
             timestamp,
-            TokenBreakdown {
-                input: usage.input.unwrap_or(0).max(0),
-                output: usage.output.unwrap_or(0).max(0),
-                cache_read: usage.cache_read.unwrap_or(0).max(0),
-                cache_write: usage.cache_write.unwrap_or(0).max(0),
-                reasoning: 0,
-            },
+            usage.to_breakdown(),
             0.0,
             agent.clone(),
         );
--- a/crates/tokscale-core/src/sessions/pi.rs
+++ b/crates/tokscale-core/src/sessions/pi.rs
@@ -8,7 +8,9 @@
 //! Pi descendants reuse this record layout verbatim, so [`parse_pi_format_file`]
 //! is shared: see `sessions::senpi` for Senpi (OmO Native).
 
-use super::utils::{file_modified_timestamp_ms, lossy_lines_with_bytes, LossyLine};
+use super::utils::{
+    file_modified_timestamp_ms, lossy_lines_with_bytes, parse_json_line, LossyLine,
+};
 use super::{normalize_workspace_key, workspace_label_from_key, UnifiedMessage};
 use crate::provider_identity::inferred_provider_from_model;
 use crate::TokenBreakdown;
@@ -624,16 +626,14 @@ fn parse_pi_format_file_inner(
         }
 
         if session_id.is_none() {
-            buffer.clear();
-            buffer.extend_from_slice(trimmed.as_bytes());
-            let entry_type = match simd_json::from_slice::<PiEntryTypeProbe>(&mut buffer) {
-                Ok(probe) => probe.entry_type,
-                Err(_)
-                    if options.lossy_line_reader && pre_header_line_is_skippable(trimmed, None) =>
+            let entry_type = match parse_json_line::<PiEntryTypeProbe>(trimmed, &mut buffer) {
+                Some(probe) => probe.entry_type,
+                None if options.lossy_line_reader
+                    && pre_header_line_is_skippable(trimmed, None) =>
                 {
                     continue;
                 }
-                Err(_) => return Vec::new(),
+                None => return Vec::new(),
             };
 
             if entry_type != "session" {
@@ -643,11 +643,8 @@ fn parse_pi_format_file_inner(
                 return Vec::new();
             }
 
-            buffer.clear();
-            buffer.extend_from_slice(trimmed.as_bytes());
-            let header = match simd_json::from_slice::<PiSessionHeader>(&mut buffer) {
-                Ok(h) => h,
-                Err(_) => return Vec::new(),
+            let Some(header) = parse_json_line::<PiSessionHeader>(trimmed, &mut buffer) else {
+                return Vec::new();
             };
             let has_raw_damaged_lineage_key = line
                 .raw_bytes()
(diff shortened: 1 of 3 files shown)


junhoyeo:
## Independent verification

I re-checked the three load-bearing claims rather than taking the numbers on trust.

**Binary delta reproduced exactly.** Built both revisions locally, stripped release: `main` 72a8ea0f = **16,167,488**, this branch = **16,167,504**. The +16 bytes is real and this is not a size win — correctly reported as such rather than padded.

**Behaviour is unchanged, verified by A/B against a controlled fixture home.** I built transcripts covering the exact edge cases this PR touches — plain, BOM-prefixed header, a record whose `model` carries a raw `0xFF`, and `cost` as both a number and a string — laid out under `.pi`, `.senpi`, `.config/kimchi` and `.prime`, then ran both binaries with isolated `TOKSCALE_CONFIG_DIR` so no shared cache could mask a difference. Order-normalised, the parsed output is **identical**.

| observation | confirms |
|---|---|
| `prime-agent` has `model=unknown in=28`; pi/senpi/kimchi do not | the lossy path still emits the invalid-UTF-8 record and the strict path still skips it |
| pi/senpi `claude-opus-5 msgs=5` across 5 files | BOM-prefixed headers still recovered |
| kimchi/prime `msgs=1` for the same ids | dedup still collapses repeated session/message ids |
| `gpt-5 in=16 out=4 msgs=2` | both `cost: 5` and `cost: "free"` still parse — precisely why folding `PiUsage` into `CamelUsage` was refused |

**The ordering nondeterminism predates this branch.** The raw diff shows client entries permuted, which could look like a regression. Running the **unmodified main binary against itself** twice produces the same permutation and 12 differing lines, so it is pre-existing HashMap iteration order, not something this PR introduces.

Also confirmed: only `sessions/{pi,prime_agent,utils}.rs` are touched (the branch predates #1141, which is why a raw `origin/main..HEAD` diff misleadingly also lists `clients.rs`/`wrapped.rs`), no `parser_version` or `CACHE_FORMAT_VERSION` is modified, and the deleted `LossyLinesWithBytes`/`LossyLine` are gone while `LossyLines` correctly remains for its other callers.

Note: Codex could not review this PR — the account is over its code-review usage limit.

PR #1144, opened 2026-08-18
fix(pricing): correct MiniMax-M3 first-party rates in alias regression fixture

octo-patch:
Reason: The `minimax-m3` alias regression fixture stated stale first-party rates and no cache-read rate, and the `minimax` summarizer backend was missing from the backend tables in all four READMEs.

## Pricing fixture

`test_apply_pricing_if_available_prices_minimax_m3_bare_id_via_alias` builds a fixture dataset where the first-party row competes with a hosted reseller row and a zero-cost third-party row, then asserts the bare `minimax-m3` id resolves to the first-party key. The row standing in for "real first-party rates" was priced at $0.30 input / $1.20 output per million tokens and carried no cache-read rate.

The elected key was still correct, so the test passed — but the numbers a reader takes away from a row commented "Real first-party rates" were wrong, and the missing cache-read rate meant cached input in this fixture priced at $0 with nothing pointing that out.

This updates the row to the current published first-party rates — $0.60 input, $2.40 output, $0.12 cache read per million tokens — and gives the fixture message 500k cached-input tokens so the cache-read rate is covered by the cost assertion rather than sitting in the row unexercised. The asserted cost moves from 0.42 to 0.9.

The two competing rows are untouched: the zero-cost row still yields 0.0 and the reseller row still yields 42.0, so both remain far enough from the first-party result that electing either could not be mistaken for it.

## Summarizer backend tables

`--summarizer minimax` has been dispatchable in `report.rs` since the backend was added, and it appears in the command examples in each README, but the **Summarizer backends** table — the place a reader actually goes to choose a backend — listed only `apple-fm`, `claude`, `codex`, `gemini`, and `kiro` in all four translations. None of the environment variables that configure it were documented there.

This adds a `minimax` row to the table in `README.md`, `README.ja.md`, `README.ko.md`, and `README.zh-cn.md`, documenting the values the code actually reads: `MINIMAX_API_KEY` / `MINIMAX_API_TOKEN` for the key, `MINIMAX_API_REGION=cn` to select `https://api.minimaxi.com/v1` instead of the default `https://api.minimax.io/v1`, and `MINIMAX_MODEL` to override the default `MiniMax-M3`. The backend is also added to the task-attributed-reports feature bullet in each README, which listed the other five backends.

## Checks

- `cargo test -p tokscale-core --lib` — 1785 passed, 1 pre-existing failure unrelated to this change: `test_empty_parse_results_are_not_cached_for_optional_file_sources` sets a fixture file to mode `0o000` and expects the parse to yield nothing, but the test ran as root, which bypasses the permission check and reads the file anyway. It fails the same way on an unmodified checkout in this environment.
- `cargo test -p tokscale-core --lib minimax` — 5 passed, covering the updated fixture and the alias resolution tests.
- Removing `cache_read_input_token_cost` from the updated row was verified to fail the assertion (0.84 against an expected 0.9), confirming the new rate is actually exercised.

`cargo clippy` was not run — the component was not available in this environment.

---
## Summary by cubic
Fixes stale `MiniMax-M3` first‑party rates in the alias regression fixture and adds the `minimax` summarizer backend to all README backend tables. The test now asserts current first‑party pricing, including cache reads, and the docs show how to configure the backend.

- Pricing fixture: Old row used $0.30 input / $1.20 output per million with no cache‑read; new row uses $0.60 input / $2.40 output / $0.12 cache‑read. The message now includes 500k cached‑input tokens, moving the expected cost from 0.42 to 0.9. Competing zero‑cost and reseller rows are unchanged; alias still resolves to the first‑party key.
- Docs: Add `minimax` to backend tables and feature bullets in `README.md`, `README.ja.md`, `README.ko.md`, and `README.zh-cn.md`. Document `MINIMAX_API_KEY`/`MINIMAX_API_TOKEN`, default model `MiniMax-M3`, default endpoint `https://api.minimax.io/v1`, `MINIMAX_API_REGION=cn` to use `https://api.minimaxi.com/v1`, and `MINIMAX_MODEL` to override.

<sup>Written for commit 9a8deda0fab681ced6715b7eb2785e3f2164ab52. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1144?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -184,7 +184,7 @@ AI支援開発の時代において、**トークンは新しいエネルギー*
 - **ネイティブRustコア** - 10倍高速な処理のため、すべての解析と集計をRustで実行
 - **Web可視化** - 2Dと3Dビューのインタラクティブ貢献グラフ
 - **柔軟なフィルタリング** - プラットフォーム、日付範囲、年別フィルタリング
-- **タスク別レポート** - マルチバックエンド対応（Apple FM、Claude、Codex、Gemini、Kiro）の LLM によるセッション要約とタスクグルーピング
+- **タスク別レポート** - マルチバックエンド対応（Apple FM、Claude、Codex、Gemini、Kiro、MiniMax）の LLM によるセッション要約とタスクグルーピング
 - **JSONエクスポート** - 外部可視化ツール用のデータ生成
 - **ソーシャルプラットフォーム** - 使用量の共有、リーダーボード競争、公開プロフィール閲覧
 
@@ -765,6 +765,7 @@ LLM 要約は**デフォルトで有効**になっています（`--no-summarize
 | `codex` | `codex --quiet` | Codex CLI がインストールされ認証済みである必要があります。 |
 | `gemini` | `gemini -p` | Gemini CLI がインストールされ認証済みである必要があります。 |
 | `kiro` | `kiro --non-interactive` | Kiro CLI がインストールされ認証済みである必要があります。 |
+| `minimax` | （HTTP API） | OpenAI 互換の chat-completions API を使用するため、CLI は不要です。`MINIMAX_API_KEY` または `MINIMAX_API_TOKEN` を設定してください。既定ではグローバルエンドポイント（`https://api.minimax.io/v1`）で `MiniMax-M3` を使用します。`MINIMAX_API_REGION=cn` を設定すると `https://api.minimaxi.com/v1` を使用し、`MINIMAX_MODEL` で別のモデル（例: `MiniMax-M2.7`）を選択できます。 |
 
 **仕組み:**
 
--- a/README.ko.md
+++ b/README.ko.md
@@ -182,7 +182,7 @@ AI 지원 개발 시대에 **토큰은 새로운 에너지**입니다. 토큰은
 - **네이티브 Rust 코어** - 모든 파싱과 집계를 Rust로 처리해 최대 10배 빠른 성능
 - **웹 시각화** - 2D 및 3D 뷰의 인터랙티브 기여 그래프
 - **유연한 필터링** - 플랫폼, 날짜 범위 또는 연도별 필터링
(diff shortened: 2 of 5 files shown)


junhoyeo:
Merging. Verified the parts that can be checked locally rather than taking them on trust.

The fixture arithmetic is exact: the old row gave `1_000_000 * 3e-7 + 100_000 * 1.2e-6 = 0.42` and the new one gives `1_000_000 * 6e-7 + 100_000 * 2.4e-6 + 500_000 * 1.2e-7 = 0.90`, matching the asserted value, and the per-token constants convert cleanly to the $0.60 / $2.40 / $0.12 per-million figures in the comment. Adding cached input to the fixture message is the right call — a cache-read rate sitting in a row that no assertion exercises is how that kind of drift goes unnoticed in the first place.

Every environment variable the new table row documents matches what the code actually reads: `MINIMAX_API_KEY` with the `MINIMAX_API_TOKEN` fallback at `report.rs:950`, `MINIMAX_API_REGION` selecting `https://api.minimaxi.com/v1` at `:916`, `MINIMAX_MODEL` overriding the `MiniMax-M3` default at `:908`, and the global default at `:913`.

`cargo test -p tokscale-core --lib minimax` passes all five here.

One caveat on the record: I could not independently confirm those are MiniMax's current published first-party rates, since web access is not configured in this environment. It is a test fixture rather than the pricing dataset, so production pricing is unaffected either way, but the numbers are only as good as the source you took them from.

The pre-existing failure you noted is real and unrelated — `test_empty_parse_results_are_not_cached_for_optional_file_sources` assumes a `0o000` file is unreadable, which does not hold when the suite runs as root.

PR #1154, opened 2026-08-19
fix(cursor): preserve exported token buckets and costs

carlosflorencio:
## Summary

- Map Cursor's exported input and cache-write columns directly to their independent token buckets.
- Preserve finite CSV costs, including explicit zero, as provider-reported so the pricing pass cannot overwrite Cursor's authoritative amount.
- Leave missing and sentinel costs such as `Included` and `-` eligible for model-catalog estimation.
- Bump the Cursor parser cache version so previously cached token arithmetic and cost provenance are rebuilt.

## Why

Cursor's `Input (w/ Cache Write)` and `Input (w/o Cache Write)` columns are separate buckets, but Tokscale subtracted the latter from the former. Tokscale also discarded the provenance of Cursor's numeric `Cost` column and then replaced it with catalog pricing. That under-counted tokens and dropped the Team/Enterprise Cursor Token Rate already included in the export.

## Validation

- `cargo test -p tokscale-core cursor` — 47 passed.
- `cargo test -p tokscale-core -- --skip test_empty_parse_results_are_not_cached_for_optional_file_sources` — 1,847 passed, 4 ignored; the skipped Unix permissions test cannot simulate unreadability when run as root.
- `cargo check -p tokscale-core --all-targets`

## Scope

This change trusts Cursor's numeric exported cost. It does not add membership persistence or a membership-based surcharge fallback for rows without a usable cost.

## Impact

Cursor's export lists `Input (w/ Cache Write)` and `Input (w/o Cache Write)` as separate buckets. The old parser computed `cache_write = input_with_cache_write - input_without_cache_write`, which undercounts cache-write tokens whenever the two columns are not a strict superset/subset. It then overwrote Cursor's numeric `Cost` with a catalog estimate, dropping any surcharge already baked into the export (e.g. the Team/Enterprise Cursor Token Rate).

Example from a real GPT-5.6 Luna export row set:

| Field | Before | After |
|---|---:|---:|
| Cache-write tokens | 600,152 | 920,460 |
| Reported cost | ~$2.78 (catalog estimate) | $14.65 (Cursor's exported cost) |

The `$14.65` matches what Cursor's own dashboard shows for that usage, because the export's `Cost` column already includes Cursor's per-token rate for the plan.

## Behavior change for existing users

Bumping the Cursor parser cache version to `2` rebuilds previously cached rows on the next run. As a result:

- Cache-write token counts for past Cursor usage may increase.
- Cursor spend that had a numeric exported cost now reports that exported amount instead of a catalog estimate, so historical Cursor totals can shift (usually upward).

No manual migration is required; the cache rebuilds automatically.

## Cost provenance rules

- Numeric `Cost` (including `$0.00`) -> retained and marked provider-reported; never repriced.
- `Included`, `-`, empty, and non-finite values -> left as unknown so catalog pricing can estimate them.

## Validation

In addition to the unit tests, this was verified end-to-end against a real Cursor usage export: after the cache rebuild, every token bucket in the `models --client cursor` report matched the export's columns exactly, and the reported total equaled the sum of the export's numeric `Cost` column plus catalog estimates for the sentinel (`Included`/`-`) rows.

---
## Summary by cubic
Preserves Cursor’s exported token buckets and provider‑reported costs, and rejects negative exported costs. Previously we subtracted “Input (w/o Cache Write)” from “Input (w/ Cache Write)” and repriced numeric “Cost”; now we map buckets directly, keep Cursor’s numeric cost (including $0.00), and treat negative values as unknown.

- Map “Input (w/ Cache Write)” -> cache_write and “Input (w/o Cache Write)” -> input; no subtraction.
- Parse costs as finite, non‑negative numbers; retain numeric values as provider‑reported and never reprice. Treat “Included”/“-”/missing and negative costs as unknown so catalog pricing can estimate.
- Bump the Cursor parser cache version to 2 so `tokscale-core` rebuilds cached token arithmetic and cost provenance.
- Side effects: cache_write tokens may increase; post‑rebuild spend aligns with Cursor exports, including plan‑specific rates. No manual migration required.

<sup>Written for commit 3f27d290facc1ef9b359967a51f21edb8a532a54. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1154?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


carlosflorencio:
Pushed `1dbe59e` with a small parser hardening fix: negative or otherwise invalid Cursor costs are now left eligible for catalog estimation instead of being marked as authoritative `$0`, while explicit `$0.00` remains provider-reported. Added a regression test; the full core test suite passes (1,848 passed, 4 ignored), with `cargo check --all-targets` and formatting clean.

junhoyeo (approved):
Thanks!

PR #1160, opened 2026-08-20
feat(clients): add fx (vercel-labs) session parsing

thedavidweng:
## Summary

- Register [fx](https://github.com/vercel-labs/fx) (vercel-labs), the Zig-based Unix-like coding agent, as a first-class Tokscale client across the Rust core, CLI/TUI, frontend registry, and documentation
- Parse per-session usage snapshots from `~/.fx/sessions/<sessionId>/usage-v2.json`, paired with the sibling `session.json` (workspace root + timestamps) and the shared `sessions/index.json` (session titles); the deprecated global `usage.jsonl` stream is intentionally ignored
- Emit one `UnifiedMessage` per (session × model) with `message_count = request_count` and the authoritative provider-reported cost; a snapshot that carries only the top-level aggregates (empty `models`) is attributed to a synthetic `fx-unknown` model so the session totals are never silently dropped
- Tolerate JSON `null` for the nullable `reasoning_tokens`/`request_count` fields (the wire schema serializes absent values as null), which previously failed the whole file parse
- Wire the scanner pattern, CLI `ClientFilter`, TUI hotkey (`X`, avoiding the global `s` sources-picker binding), `parser_version(Fx)=1`, and the client-count / display-name registry invariants
- Point the core logo at the checked-in `.github/assets/client-fx.png` (matching the frontend) and document fx in the client tables across all four READMEs (en/ko/ja/zh-cn)
- Frontend: register `fx` in `SUPPORTED_CLIENT_TYPES`, `SOURCE_DISPLAY_NAMES`, `SOURCE_LOGOS`, and `SOURCE_COLORS`

## Compatibility

A session whose snapshot has an empty `models` array and no top-level aggregates is skipped (nothing to report). Otherwise the top-level totals are attributed to the synthetic `fx-unknown` model rather than dropped.

## Validation

- `cargo test -p tokscale-core --lib` — 1808 passed
- `cargo test -p tokscale-cli --bin tokscale` — 1113 passed
- `bun run --cwd packages/frontend test` — 910 passed (incl. `clientRegistry.test.ts` fx submit + logo/asset checks)
- `bun run --cwd packages/frontend typecheck`

`cargo clippy --all-targets` also reaches a pre-existing `unnecessary_sort_by` denial in `sessions/cherrystudio.rs` under Rust 1.97.1; the changed library and binary targets pass with warnings denied.


junhoyeo (approved):
Great!

PR #1161, opened 2026-08-20
feat(sessions): add Oh My Pi (omp) as a local session source

bn852j6nky-code:
## What

Adds Oh My Pi (`omp`) as a local session source, so its usage is counted alongside Pi rather than silently missing.

## How

omp is a pi-mono descendant and writes the same session JSONL record format, so parsing delegates to `pi::parse_pi_format_file` — the same one-line pattern `senpi.rs`, `gjc.rs` and `prime_agent.rs` already use. Only the scan root and client id differ.

The layout sits one level deeper than Pi's:

```
~/.omp/agent/sessions/<encoded-cwd>/<timestamp>_<uuid>.jsonl                    # session
~/.omp/agent/sessions/<encoded-cwd>/<timestamp>_<uuid>/<AgentName>.jsonl        # subagents
~/.omp/agent/sessions/<encoded-cwd>/<timestamp>_<uuid>/__advisor.<name>.jsonl   # advisor runs
```

Each subagent file carries its own session header, so the recursive `*.jsonl` scan picks them up and parses each independently.

## One deliberate decision worth reviewing

**The root is a fixed `PathRoot::Home`, not a `PathRoot::EnvVar`.** omp honours `PI_CODING_AGENT_DIR` for its session directory — but that is the *same* variable Pi uses. Wiring it as an env-var root would mean a single override points both clients at one tree and double-counts every message in it. Pi is registered with a fixed home-relative root for exactly this reason, so omp follows it. `test_omp_root_ignores_pi_coding_agent_dir` pins the behaviour so it does not get "fixed" into a regression later.

Happy to switch to an env-var root if you would rather have the override and handle the collision another way.

## Verification

- `cargo test -p tokscale-core` — 1815 pass. The only failure is the pre-existing `test_discover_micode_dbs_in_dirs_collapses_same_file_via_symlink`, which calls `std::os::windows::fs::symlink_dir(..).unwrap()` and needs Developer Mode/admin on Windows; unrelated to this change.
- `cargo fmt --all -- --check` clean; `cargo clippy` adds no new warnings.
- Against a real install, `tokscale clients` reports the two as separate sources with no double-counting:

```
Pi
sessions: ~\.pi\agent\sessions
messages: 25.1K

Oh My Pi
sessions: ~\.omp\agent\sessions
messages: 5.6K
```

Pi's 25.1K is unchanged from before the adapter existed.

## Notes

- New client discriminant `Omp = 49`; the `ClientId::COUNT` test is updated 49 -> 50.
- TUI hotkey is `Y` — every case of `o`, `m` and `p` was already taken (OpenCode/OpenClaw, MiMo Code/Mcode, Pi/Prime Agent) and `s` is the global sources binding, so `Y` stands in for the Y in "oh mY pi". Happy to change it.
- No logo URL set, matching other recently-added clients.

---
## Summary by cubic
Adds Oh My Pi (`omp`) as its own local session source and gives it sole ownership of `~/.omp/agent/sessions`. Previously `omp` files were ignored or counted under Pi, causing double counting when both were enabled; now `omp` is parsed and attributed to `omp` only.

- Scanner: drop the `.omp` special-case under Pi; register `ClientId::Omp` with `PathRoot::Home`, `relative` `.omp/agent/sessions`, and `pattern` `*.jsonl`. New tests pin that Pi scans only `~/.pi/...`, `omp` scans only `~/.omp/...`.
- Root policy: ignore `PI_CODING_AGENT_DIR` for `omp` to avoid collisions with Pi; test `test_omp_root_ignores_pi_coding_agent_dir` pins this.
- Parser: add `sessions::omp` delegating to `pi::parse_pi_format_file`; recursive scan picks up subagent/advisor `*.jsonl`. Cache version arm for `omp` notes shared Pi-format parser; Pi parser bumps may invalidate `omp` caches.
- Core wiring: include `omp` in cached and local parsing lanes; update counts; `ClientId::COUNT` is now 50.
- CLI/TUI: add `Omp` filter (`omp`) and hotkey `Y`; update registry tests.
- Frontend/Docs: register display name, logo, and color; update data-location and platform-path tables across all READMEs; add `omp` to `packages/cli` and `packages/tokscale` keywords and to `packages/frontend` client registries.

<sup>Written for commit bccd63ab3dc2a0c6289aea7d484185d3e4b0e2f1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1161?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -69,7 +69,8 @@
 | <img width="48px" src=".github/assets/client-codebuff.png" alt="Codebuff" /> | [Codebuff](https://codebuff.com/) | `~/.config/manicode/` (+ `manicode-dev`、`manicode-staging`; `CODEBUFF_DATA_DIR` でオーバーライド可能) |
 | <img width="48px" src=".github/assets/client-freebuff.png" alt="Freebuff" /> | [Freebuff](https://github.com/CodebuffAI/freebuff) | Codebuff と同じ `~/.config/manicode/` を共有（同一ランタイム）；トークン使用量はトランスクリプトから推定（ローカル使用量なし；`FREEBUFF_DATA_DIR` でオーバーライド可能） |
 | <img width="48px" src=".github/assets/client-droid.png" alt="Droid" /> | [Droid (Factory Droid)](https://factory.ai/) | `~/.factory/sessions/` |
-| <img width="48px" src=".github/assets/client-pi.png" alt="Pi" /> | [Pi](https://github.com/badlogic/pi-mono) | `~/.pi/agent/sessions/` and `~/.omp/agent/sessions/` ([Oh My Pi](https://github.com/can1357/oh-my-pi)) |
+| <img width="48px" src=".github/assets/client-pi.png" alt="Pi" /> | [Pi](https://github.com/badlogic/pi-mono) | `~/.pi/agent/sessions/` |
+| <img width="48px" src="https://github.com/can1357.png" alt="Oh My Pi" /> | [Oh My Pi](https://github.com/can1357/oh-my-pi) | `~/.omp/agent/sessions/**/*.jsonl` |
 | <img width="48px" src=".github/assets/client-senpi.png" alt="Senpi" /> | [Senpi (OmO Native)](https://github.com/code-yeongyu/senpi) | `~/.senpi/agent/sessions/` (`SENPI_CODING_AGENT_DIR` でオーバーライド可能) |
 | <img width="48px" src="https://github.com/getkimchi.png" alt="Kimchi" /> | [Kimchi Coding](https://kimchi.dev/) | `~/.config/kimchi/harness/sessions/`（`KIMCHI_CODING_AGENT_DIR` でオーバーライド可能） |
 | <img width="48px" src=".github/assets/client-synthetic.png" alt="Reasonix" /> | [Reasonix](https://github.com/esengine/DeepSeek-Reasonix) | `~/.reasonix/stats/*.jsonl`（`REASONIX_STATE_HOME` または `REASONIX_HOME` でオーバーライド可能） |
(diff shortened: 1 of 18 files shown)


junhoyeo:
Thanks for this — the parser work is careful, and the comment explaining why the root is fixed rather than env-derived shows you were already thinking about double-counting. That instinct was right, but it needs to go one step further: Pi already scans this exact root today.

**Blocking: every Oh My Pi session is counted twice.**

`scanner.rs:1838-1842` on main pushes `~/.omp/agent/sessions` as a `ClientId::Pi` scan task unconditionally:

```rust
// Oh My Pi fork (https://github.com/can1357/oh-my-pi) — same JSONL format, different root
if enabled.contains(&ClientId::Pi) {
    let omp_path = join_native(home_dir, ".omp/agent/sessions");
    push_unique_scan_task(&mut tasks, &mut seen_scan_roots, ClientId::Pi, omp_path);
}
```

The new client declares `relative: ".omp/agent/sessions"`, the same path, and this PR does not touch `scanner.rs`. `push_unique_scan_task` dedupes by `(ClientId, path)`, so a second client on the same path is not deduped — it is a different key.

Built the branch and ran it against a fixture holding one omp session with 1000 input / 500 output:

```
-c pi    -> [('pi',  1000, 500)]
-c omp   -> [('omp', 1000, 500)]
default  -> [('omp', 1000, 500, 1), ('pi', 1000, 500, 1)]   TOTAL input = 2000
```

With the default client set — which is what users get — that session's tokens and cost are doubled. Two existing tests pin the current behaviour and will need to move with whatever you decide: `test_scan_all_clients_omp_scanned_as_pi` and `test_scan_all_clients_pi_from_both_paths`.

Worth deciding deliberately which way to resolve it. Splitting omp out is defensible — it is a distinct product and attribution is genuinely more useful — but then the Pi extra-root at `scanner.rs:1838` has to go, and existing users' historical omp usage silently moves from the `pi` row to a new `omp` row. Keeping it folded into Pi costs nothing and changes no one's numbers. Either is fine; both cannot ship together.

**Two registry gaps, which are what CI is failing on.**

`tui::data::tests::test_client_display_names_match_canonical_registry` fails locally on this branch — the TUI display-name map has no `omp` entry.

The frontend registries are also missing it, which is the red Frontend Vitest. `packages/frontend/src/lib/constants.ts` has zero occurrences of `omp`, where a comparable recent client has three:

```
constants.ts:79:   dsh: "DeepSeek Harness",
constants.ts:140:  dsh: "https://github.com/deepseek-ai.png",
constants.ts:194:  dsh: "#4D6BFE",
```

`logo: None` in the client entry is also unusual — every other client carries one, and `https://github.com/can1357.png` would work if there is no dedicated mark.

The rest of the wiring I checked by hand looks complete: `clients.rs`, `sessions/mod.rs`, both dispatch paths in `lib.rs`, and `main.rs`/`client_ui.rs` filters are all updated, and `cargo test -p tokscale-core --lib` is otherwise green at 1817 passed.

PR #1164, opened 2026-08-21
fix(scanner): remove nested par_bridge to prevent futex deadlock on large directories

### LINKED ISSUE #1153: Submit scan intermittently hangs on Linux (futex_wait_queue, ~0% CPU) when walking ~/.config/Code/logs with tens of thousands of entries
## Summary
`tokscale submit` intermittently hangs during 'Scanning local session data...' on a Linux host (kyber). The process parks 97+ threads on `futex_wait_queue` at ~0% CPU (1.8s CPU consumed in 3.5min), making no progress, and never reaches submission. Reproduced repeatedly; on retry the same scan sometimes completes fine (flaky, contention-dependent).

Environment: CLI 4.13.0 (cli-linux-x64-gnu), Linux x86_64.

## What the stuck process shows
- `/proc/<pid>/wchan` = `futex_wait_queue`, 97 threads, VmRSS ~22MB
- Open fds include `~/.config/Code/logs/...` (VSCode extension session logs)
- `tokscale clients` also hangs with the same profile

## Suspected trigger
The scanner walks `~/.config/Code/logs` which on this box has **34,598 entries in a single directory** (roo-cline, kilocode, Cline, etc. session logs accreting hourly). Combined with the large opencode SQLite DB (1.19 GB) and codex sessions (2.1 GB), the concurrent scan thread-pool rolls up into a futex deadlock under load.

## Mitigation applied (workaround, not fix)
In our automation we now cap the submit with `timeout 240` so a wedged run self-terminates. See https://github.com/shunkakinoki/dotfiles/pull/2412

## Suggestion
Consider: (a) bounding/parallelizing the directory walk more carefully so a huge dir cannot wedge the pool, (b) skipping or streaming the opencode SQLite DB instead of loading it wholesale, or (c) documenting scan-path pruning.

aruru10313:
## Summary
    Refs #1153 where `tokscale submit` intermittently hangs (futex_wait_queue, ~0% CPU) when scanning large directories like `~/.config/Code/logs`.

    ### Description
    The directory scanner (`scan_directory`) used `.par_bridge()` on the `WalkDir` iterator. Because `scan_directory` is already called from within a parallel iterator (`.into_par_iter()`)
  mapping across multiple scan roots, this created nested parallelism.

    `par_bridge()` on a sequential iterator uses a mutex to synchronize thread access. When tens of thousands of files are discovered in a single directory, the Rayon thread pool becomes
  starved and deadlocks waiting on this mutex, hanging the scan entirely.

    Removing `.par_bridge()` from `scan_directory` resolves the deadlock by making the file tree walk sequential per root, while the roots themselves are still processed concurrently. Since
  `WalkDir` reads entry types without triggering a `stat()` syscall on Linux/Windows, the sequential filter is extremely fast and the overhead of `.par_bridge()` was unnecessary.

---
## Summary by cubic
Removes `.par_bridge()` from the directory scanner to stop futex deadlocks on large directory trees. Before: `WalkDir` used `.par_bridge()` inside a parallel root iterator; after: each root walks sequentially while roots still run in parallel.

- Fixes #1153 causing intermittent hangs in `tokscale submit` on very large trees (e.g., `~/.config/Code/logs`).
- Scope: only `scan_directory`; no API or CLI changes.
- Impact: eliminates deadlocks; performance impact is expected to be negligible.

<sup>Written for commit 8076c7e210713e9c2ce7ed4bab45c267d72c22a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1164?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/scanner.rs
+++ b/crates/tokscale-core/src/scanner.rs
@@ -382,7 +382,6 @@ pub fn scan_directory(root: &str, pattern: &str) -> Vec<PathBuf> {
 
     let mut paths: Vec<PathBuf> = WalkDir::new(root)
         .into_iter()
-        .par_bridge()
         .filter_map(|e| e.ok())
         .filter(|e| {
             let path = e.path();


junhoyeo:
Thanks for this — merging it. I merged `main` into the branch (the `Lint` red was ours: your base `ab5b5095` carried a rustfmt drift that `ee26e61d` has since fixed on main, so it was never your code).

Two notes, one of which changed the PR body.

**The change is good, and it is also a speedup.** `scan_directory` ends in `paths.sort_unstable()`, so removing `.par_bridge()` is provably behaviour-preserving, and benchmarked over a synthetic 31.5k-file tree the sequential walk beats the bridged one in every configuration (~130ms vs ~250ms for a single large root). `par_bridge` serialises `WalkDir::next()` under its own mutex and only fans out the cheap filter closure, so it was a net pessimisation. Root-level parallelism via `into_par_iter` is untouched.

**But the stated mechanism does not hold, so I changed `Fixes #1153` to `Refs #1153`.** I went and read the pinned rayon (1.11.0) `par_bridge.rs`: it carries a `threads_started` recursion guard with the comment *"in case `Iter::next()` calls rayon internally, so we don't deadlock our mutex"*, and `drop(iter)` precedes `folder.consume(it)`. The filter closure in `scan_directory` never re-enters rayon, so there is no lock-ordering cycle to deadlock on. What nested `par_bridge` produces here is a convoy stall, not a deadlock — same remedy, different cause.

That matters only because #1153 reports a 1.19 GB opencode SQLite DB and 2.1 GB of codex sessions alongside the 34k-entry directory, and names the wholesale SQLite load as a co-suspect. Auto-closing it on merge would retire the only tracking issue for a hang whose root cause we have not actually established. Everything else about the PR stands — this is purely about not losing the issue.

PR #1166, opened 2026-08-21
fix(cli): do not wait on stdout pump thread if headless capture times out

### LINKED ISSUE #1049: headless capture can hang past its timeout when the child leaves a descendant holding stdout
## Summary

`run_capture_command`'s timeout bounds how long it waits for the *direct child*, but not how long the call takes. If the child leaves behind a descendant that inherited its stdout pipe, the function blocks after the kill — potentially forever — and `TOKSCALE_NATIVE_TIMEOUT_MS` silently stops meaning what it says.

Reasoned from the code, **not reproduced**. Filing it because the reasoning is short and the failure mode is a hang rather than a wrong number.

## Mechanism

`crates/tokscale-cli/src/main.rs` (around `run_capture_command`):

```rust
let output_handle = thread::spawn(move || -> Result<()> {
    let mut reader = std::io::BufReader::new(stdout);
    loop {
        match reader.read(&mut buffer) {
            Ok(0) => return Ok(()),     // <- only exit
            ...
        }
    }
});

// ... on deadline:
timed_out = true;
let _ = child.kill();
break child.wait()?;

// ...
let output_result = output_handle.join()?;   // <- unbounded
```

The pump thread's only non-error exit is `Ok(0)` — EOF on the child's stdout. A pipe reports EOF when **every** handle to its write end is closed, and a spawned child passes its stdout handle to anything it spawns.

`child.kill()` terminates one process. It is `TerminateProcess` on Windows and `SIGKILL` on Unix; neither reaches descendants, and neither closes a handle a descendant already holds. So after a timeout:

1. the direct child dies and `child.wait()` reaps it,
2. a surviving grandchild still holds the write end,
3. `read` never returns 0,
4. `output_handle.join()` blocks with no deadline.

`std::process::exit(124)` at the call site is never reached, because control does not return from `run_capture_command`.

## Why it hasn't been seen

The test fixture (`crates/tokscale-cli/src/bin/fake_codex.rs`) sleeps and spawns nothing, so its pipe closes the instant it is killed. `headless_capture_slow_command_times_out` therefore exercises the kill path but never the surviving-descendant case.

Real `codex` is a different matter — an agent CLI that can spawn MCP servers, shells and language servers. Any of those outliving the parent kill reproduces this.

## Scope

Affects `tokscale headless <source>` on every platform. Not a data-correctness bug: nothing is miscounted, and partial output already on disk is intact. The symptom is that the command hangs instead of exiting 124, and the user's configured timeout appears not to work.

## Possible directions

Not proposing one — this wants a decision from whoever owns the capture path.

- **Join with a deadline.** Give the pump thread a grace period after the kill and proceed without it on expiry. Simple, but detaching a thread that holds a file handle needs care.
- **Kill the process group.** `setsid` + `killpg` on Unix, a Job Object on Windows. Correct, and matches what a user means by "time out this subprocess", but it is a real behaviour change and platform-specific.
- **Drop the read end.** Closing the reader makes a pending `read` fail rather than block. Needs checking per platform — it does not obviously interrupt a blocked read already in flight.

The narrow fix (bounded join) restores the timeout's meaning. The broad fix (process group) also stops orphaned grandchildren accumulating, which is a separate problem this code has.

## Related

- #1046 — widened the timing band in `headless_capture_slow_command_times_out`; that band's tightness is what led me to read this path.

aruru10313:
### Summary
    Fixes #1049 where `tokscale headless` hangs past its timeout if the child process leaves a descendant holding the stdout pipe.

    ### Description
    In `run_capture_command`, if a timeout occurs, the direct child is killed. However, if a grandchild process inherited the stdout pipe and survives the parent's death, the pipe remains
  open. 

    Previously, the main thread unconditionally called `output_handle.join()`, causing it to block indefinitely on the pump thread's `read()` until the grandchild eventually exited. This
  defeated the purpose of the timeout.

    This fix changes the logic so that if the deadline has expired (`timed_out` is true), we do not wait for the pump thread to join. The function immediately returns the timeout outcome,
  which results in a `std::process::exit(124)`, tearing down the blocked background thread and correctly terminating the command.

---
## Summary by cubic
Bounds stdout draining on timeout to prevent `tokscale headless` from hanging when a descendant holds the pipe, while preserving already-written output. Previously we joined the stdout pump unbounded and could hang; now we wait up to 2s and then exit 124.

- On timeout, wait up to 2s (`STDOUT_DRAIN_GRACE`) for the stdout pump via a channel; normal completion still waits and propagates errors.
- Avoids hangs from descendants holding stdout and reduces dropped output compared to skipping the drain entirely; any remaining stdout after the grace period is discarded.
- Adds `descendant` mode to the `fake_codex` stand-in and a Unix-only test covering #1049; suppresses `clippy::zombie_processes` in `fake_codex` to allow the deliberately unreaped grandchild.

<sup>Written for commit 3583a4feae5c278a9fb045bf81a60dbb0c549c21. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1166?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/bin/fake_codex.rs
+++ b/crates/tokscale-cli/src/bin/fake_codex.rs
@@ -64,6 +64,23 @@ fn main() {
         // so the parent's timeout always wins by a margin the runner cannot eat
         // into. See `FAKE_CODEX_SLOW_SLEEP_SECS`.
         "slow" => std::thread::sleep(std::time::Duration::from_secs(FAKE_CODEX_SLOW_SLEEP_SECS)),
+        // #1049's shape. A grandchild inherits the stdout pipe and outlives the
+        // direct child, so killing the child does not close the write end and
+        // the parent's pump never reaches EOF. The grandchild must outlast the
+        // test's own upper bound, not merely the parent's deadline: a parent
+        // that waits for EOF unboundedly has to look *slower than the bound*,
+        // or the test cannot tell it apart from a parent that drained promptly.
+        "descendant" => {
+            let exe = std::env::current_exe().expect("current_exe");
+            std::process::Command::new(exe)
+                .env("TOKSCALE_FAKE_CODEX_MODE", "slow")
+                .stdout(std::process::Stdio::inherit())
+                .stdin(std::process::Stdio::null())
+                .stderr(std::process::Stdio::null())
+                .spawn()
+                .expect("failed to spawn descendant");
+            std::thread::sleep(std::time::Duration::from_secs(FAKE_CODEX_SLOW_SLEEP_SECS));
+        }
         _ => {
             eprintln!("unknown TOKSCALE_FAKE_CODEX_MODE");
             std::process::exit(2);
--- a/crates/tokscale-cli/src/bin/fake_codex.rs
+++ b/crates/tokscale-cli/src/bin/fake_codex.rs
@@ -72,7 +72,12 @@ fn main() {
         // or the test cannot tell it apart from a parent that drained promptly.
         "descendant" => {
             let exe = std::env::current_exe().expect("current_exe");
-            std::process::Command::new(exe)
+            // Deliberately never waited on: outliving this process, still
+            // holding the inherited stdout pipe, is the entire point of the
+            // fixture. Reaping it here would close the write end and destroy
+            // the condition under test.
+            #[allow(clippy::zombie_processes)]
+            let _descendant = std::process::Command::new(exe)
                 .env("TOKSCALE_FAKE_CODEX_MODE", "slow")
                 .stdout(std::process::Stdio::inherit())
                 .stdin(std::process::Stdio::null())
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -6401,6 +6401,15 @@ struct CaptureCommandOutcome {
     timed_out: bool,
 }
 
+/// How long the stdout pump gets to finish draining a killed child's pipe.
+///
+/// Bounded because a *descendant* of the child may still hold the write end, in
+/// which case the pump never reaches EOF and an unbounded wait hangs the whole
+/// timeout (#1049). Two seconds because the ordinary case -- the pipe closing
+/// with the child -- only has to move at most one pipe buffer, so anything past
+/// a few milliseconds already means a descendant is holding it open.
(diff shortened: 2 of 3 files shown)


junhoyeo:
Thanks — the hang you found is real, and I traced it to the same place you did: on timeout we `child.kill()` then `child.wait()`, but the pump blocks in `read()` until the *pipe* EOFs, and a descendant that inherited stdout keeps the write end open, so `join()` never returns and the timeout is defeated.

I pushed a revision that keeps your fix but bounds the wait rather than removing it.

The reason: `run_headless_command` prints `"Partial output saved"` and then calls `std::process::exit(124)` (`main.rs:6579`). `process::exit` does not wait for threads, so with no join at all, anything the child had already written but the pump had not yet copied — up to a full pipe buffer — is dropped, and a `write_all` in flight can be torn. That is the common case (no descendant), so the fix traded a rare hang for routine output loss in the file we advertise as saved.

Now the pump reports completion over a channel; the timeout path waits `STDOUT_DRAIN_GRACE` (2s) and the normal path still waits unconditionally. A descendant can no longer hang us, and a child that simply died still gets drained.

I also added the coverage this needed. `cli_tests.rs:449` already admitted the gap — *"this test cannot catch #1049: the stand-in spawns nothing"* — so a full revert of your 3-line diff left the suite green. `fake_codex` now has a `descendant` mode that spawns a grandchild inheriting stdout and outliving the direct child, which is #1049's actual shape. Verified it discriminates: **fails at 122.6s against an unbounded wait** (past the 60s bound) and **passes at 14.2s** with the bounded drain. It is `#[cfg(unix)]` — the grandchild is the stand-in binary and Windows locks a running image, which would strand the fixture's TempDir for the two minutes it lives; #1049 is a Linux report and this is ordinary `std::process` code, so ubuntu + macOS coverage carries it.

I also merged `main` in — the `Lint` red was ours, not yours (base `ab5b5095` carried a rustfmt drift, fixed on main by `ee26e61d`).

PR #1167, opened 2026-08-21
docs(readme): document submit exclusion warnings and the custom-pricing fix path

thedavidweng:
## Summary

Adds a short subsection under **Social** (all four README languages) documenting behavior that currently exists only in code and issue threads:

- `tokscale submit` excludes messages that cannot be priced, and prints the three `Warning: excluded N unpriced ...` reasons:
  - `no authoritative model-to-price mapping`
  - `generic routing label has no authoritative model-to-price mapping`
  - `pricing does not cover every populated token bucket`
- Excluded usage never submits invented/guessed spend; remaining priced usage still submits.
- The supported fix path is an exact-match entry in `custom-pricing.json` (explicit `0` = free model), verified via `tokscale pricing <model-id>` and `submit --dry-run`.

## Motivation

#1021 and #1035 are both users confused by exactly these warnings with no user-facing documentation of what they mean or how to resolve them. I hit the same wall locally (26 excluded provider/model pairs across hermes/nous/devin/cursor/tencent labels) and had to read the core sources to learn that:

- custom entries are consulted before the routing-label refusal (the documented escape hatch for labels like `auto`),
- all-zero entries declare genuinely free models,
- coverage requires a rate for every populated bucket, so cache rates matter even for cheap models.

This PR distills that into the README so the next user can self-serve. No code changes.

## Verification

- Local `custom-pricing.json` built per this doc cleared all 26 exclusions (`submit --dry-run`: zero warnings, +1.68B tokens recorded).
- Section added to EN/zh-cn/ja/ko at the same location (after the CLI Submit screenshot, inside Social).

---
## Summary by cubic
Documents why `tokscale submit` excludes unpriced messages and how to resolve it with `custom-pricing.json`, across all README locales. Expands the docs to cover all six exclusion reasons and clarifies that `custom-pricing.json` keys on the model ID only. Unknown models never submit guessed spend; priced usage still submits.

- Adds a short subsection after the CLI Submit screenshot in `README.md`, `README.zh-cn.md`, `README.ja.md`, and `README.ko.md`.
- Lists all six exclusion warnings, including provider/model ambiguity cases, and reiterates that pricing must cover all populated token buckets.
- Documents the fix: add an exact-match model ID entry to `~/.config/tokscale/custom-pricing.json` (explicit `0` means a free model), verify with `tokscale pricing <model-id>` and `tokscale submit --dry-run`. The file is keyed by the model ID alone; do not paste the `provider/model` pair from the warning.
- Docs-only; merged `main`; no code or runtime behavior changes.

<sup>Written for commit 3e9dadbadb341b09f1635504b8b210acf20fa73d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1167?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo:
Thanks — merging this, with one correction pushed on top.

The section said "The three exclusion reasons", but `exclude_unpriced_submission_messages` selects among **six** constants (`lib.rs:3837-3847`). The three undocumented ones are the `SubmissionSafetyGap` arms — `does not establish the requested provider`, `does not exactly name the requested model`, and `ambiguous across non-equivalent candidates` — and they are evaluated *before* the three you documented, so for provider-prefix and model-part matches they are the common case. A reader hitting one of those would have found the docs claiming it could not happen.

I also added the key format. The warning prints `provider/model`, but `CustomPricing::lookup` keys on the model id alone (`custom.rs:253-257`), so someone copying the warning verbatim writes an entry that silently never matches. I made the same correction to the CLI hint in #1168 so the two describe one procedure.

Kept all four locales in sync, and added the blank line before `### Autosubmit` that README.md was missing (ja/ko/zh-cn all had it).

PR #1168, opened 2026-08-21
feat(submit): print a fix hint after unpriced-exclusion warnings

thedavidweng:
## Summary

When `tokscale submit` excludes unpriced usage, it prints the per-model warnings and stops there. This PR follows the warning list with a homebrew-style hint naming the remedy:

```
  Warning: excluded 1372 unpriced nous/meituan/longcat-2.0:free message(s) (87,391,743 tokens): no authoritative model-to-price mapping. Remaining priced usage will be submitted.
  Hint: excluded models stay unsubmitted until priced. Add exact-match entries to
          /Users/me/.config/tokscale/custom-pricing.json
          (an explicit 0 declares a free model or a known routing-label rate), then re-check with
          `tokscale submit --dry-run` and `tokscale pricing <model-id>`.
```

- Path is resolved via the same `get_config_dir()` the loader uses, so `TOKSCALE_CONFIG_DIR` overrides are reflected.
- Printed only when exclusions exist; styled like the other auxiliary output (`bright_black`).

## Motivation

#1021 and #1035 are users who hit these warnings and could not tell what to do next; the README section documenting the fix only landed in #1167. Even with docs, the moment of failure is the right place to teach the escape hatch — the custom-pricing mechanism (exact match, explicit-0 free models, routing-label override) is otherwise discoverable only by reading core sources.

## Verification

- New assertions in `test_submit_excludes_unpriced_usage_and_keeps_the_rest` (hint present, names `custom-pricing.json`, points at `submit --dry-run`).
- `cargo test -p tokscale-cli --test cli_tests test_submit_excludes_unpriced_usage_and_keeps_the_rest test_submit_offline_without_pricing_cache_fails` passes; `cargo fmt --check` clean.
- Manually exercised against real local data: hint renders under 26 exclusion rows.

---
## Summary by cubic
Prints a follow-up fix hint after unpriced-usage exclusion warnings in `tokscale submit`, so users know how to price and re-verify; previously we only printed per-model warnings. The hint now also states that custom-pricing entries are keyed by the model id alone, preventing silent mismatches. Submission behavior is unchanged.

- Prints the hint only when exclusions exist; styling matches other auxiliary output.
- Resolves the path via `get_config_dir()` and honors `TOKSCALE_CONFIG_DIR`, naming `custom-pricing.json`.
- Instructs users to add exact-match entries keyed by the model id (explicit 0 for free models or routing labels) and to verify with `tokscale submit --dry-run` and `tokscale pricing <model-id>`.
- Extends `test_submit_excludes_unpriced_usage_and_keeps_the_rest` to assert the hint, including the key format; no changes to priced submission logic.
- Chore: applies `cargo fmt` to OpenCode session tests only.

<sup>Written for commit 0c0512774ff5809c620f7803d7283f02718e0070. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1168?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -5689,6 +5689,23 @@ fn report_unpriced_submission_exclusions(
             .yellow()
         );
     }
+
+    // Homebrew-style follow-up: the warnings above name the gap, but nothing
+    // told the user the fix is one file away. #1021/#1035 reporters (and the
+    // custom-pricing docs added after them) all had to read core sources to
+    // discover that an exact-match entry in custom-pricing.json — including
+    // explicit 0 rates for free models and routing labels — is the supported fix.
+    if !excluded.is_empty() {
+        let pricing_path = crate::paths::get_config_dir().join("custom-pricing.json");
+        println!(
+            "{}",
+            format!(
+                "  Hint: excluded models stay unsubmitted until priced. Add exact-match entries to\n          {}\n          (an explicit 0 declares a free model or a known routing-label rate), then re-check with\n          `tokscale submit --dry-run` and `tokscale pricing <model-id>`.",
+                pricing_path.display(),
+            )
+            .bright_black()
+        );
+    }
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -5700,7 +5700,7 @@ fn report_unpriced_submission_exclusions(
         println!(
             "{}",
             format!(
-                "  Hint: excluded models stay unsubmitted until priced. Add exact-match entries to\n          {}\n          (an explicit 0 declares a free model or a known routing-label rate), then re-check with\n          `tokscale submit --dry-run` and `tokscale pricing <model-id>`.",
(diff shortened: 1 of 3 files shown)


junhoyeo:
Thanks — merging this. The review found nothing wrong with the hint itself (five candidate findings were all refuted on inspection).

I pushed one clarification. The warning printed directly above your hint uses `provider/model`, but `CustomPricing::lookup` keys on the model id alone (`custom.rs:253-257`) — so a user following the hint by copying the warning verbatim writes an entry that silently never matches and sees the same exclusion next run. The hint now says "keyed by the model id alone (the `model` half of the `provider/model` above)". Your existing assertions matched on a prefix so they still hold; I added one covering the key format.

Same correction went into the READMEs in #1167 so the CLI and docs describe one procedure rather than two.

Also merged `main` in — the `Lint` red was ours (base `ab5b5095` rustfmt drift, fixed by `ee26e61d`).

One thing worth knowing: your branch and #1162 both carried a byte-identical `opencode.rs` rustfmt hunk, because you both independently worked around that same base breakage. It is already on main, so it resolves to a no-op on merge — no conflict, nothing to do.

Issue #1145, opened 2026-08-18
pricing(kimi): add Kimi Work model aliases and fix stale kimi-for-coding mapping


REMIXPLAYER:
## Summary

Kimi Work (the Kimi desktop app's agent mode) runs the same embedded `kimi-code` kernel as the standalone Kimi Code CLI and writes the identical `wire.jsonl` session protocol, but uses its own model ids (`k2d6-agent`, `k3-agent`, `k3-agent-swarm`) that tokscale's pricing aliases do not know.

Verified against `tokscale 4.13.0` (`darwin-arm64`) using the exact model ids that appear in Kimi CLI and Kimi Work wire logs.

Prices below are **USD** as served by the OpenRouter catalog (snapshot 2026-08-18), which is what tokscale resolves after aliasing. Official Moonshot pricing is CNY; there is no public cache-write price anywhere.

### 1. Kimi Work models — add aliases

| wire `model` / `modelAlias` | current tokscale (USD) | expected after alias (USD) |
|---|---|---|
| `k2d6-agent` | mis-matches to `xai/grok-…` ($1.25 / $2.5 / $0.2) | `kimi-k2.6` ($0.58 / $2.44 / $0.10) |
| `k3-agent` | `0` | `kimi-k3` ($3 / $15 / $0.30) |
| `k3-agent-swarm` (UI key `k3-agent-ultra`) | `0` | `kimi-k3` ($3 / $15 / $0.30) |

Proposed aliases (same pattern as the existing `k3 -> kimi-k3`):

```text
k2d6-agent     -> kimi-k2.6
k3-agent       -> kimi-k3
k3-agent-swarm -> kimi-k3
```

### 2. `kimi-for-coding` alias appears stale

`kimi-for-coding` is **Kimi K2.7 Code** (per Kimi CLI config `display_name=K2.7 Coding` and opencode metadata: `kimi-for-coding -> family kimi-k2`, canonical `moonshotai/kimi-k2.7-code`). tokscale currently prices it as `kimi-k2.5` (USD ~$0.57 / $2.85 / $0.10); the correct `kimi-k2.7-code` is USD $0.71 / $3.50 / $0.15.

```text
kimi-for-coding -> kimi-k2.7-code
```

(`kimi-for-coding-highspeed -> kimi-k2.7-code-highspeed` already matches; `kimi-k2.7-code-highspeed` has no OpenRouter USD listing and is only available in official CNY pricing.)

[...]

PR #1170, opened 2026-08-21
feat(kimi): discover Kimi Work sessions natively

### LINKED ISSUE #1171: feat(kimi): discover Kimi Work sessions natively
## Summary

Issue #1145 introduced Kimi Work's model IDs and documented the native collection roots as a follow-up opportunity. Kimi Work uses the same embedded Kimi Code `wire.jsonl` protocol, but its sessions are currently only discoverable through a manually configured extra scan path.

This issue tracks the complete native discovery feature, rather than a single Windows installation error.

## Scope

- Discover Kimi Work sessions automatically on macOS and Windows.
- Reuse the existing Kimi client and Kimi Code parser; do not add a second client ID or parser.
- Discover both `conv-*` and `ctitle-*` sessions under the documented `agents/main/wire.jsonl` layout.
- On Windows, honor Kimi Desktop's optional relocated `shareDir` configuration when present, with the platform-default path as fallback.
- Keep explicit `--home` scans isolated from the current user's environment roots.
- Do not scan a Kimi Work root on Linux, where Kimi Work has no build.

## Acceptance criteria

Native Kimi Work discovery should produce the same parsed messages and pricing results as an equivalent `TOKSCALE_EXTRA_DIRS=kimi:<root>` scan, without requiring private session data in the repository. Platform-specific path and environment-root behavior should be covered by sanitized regression fixtures.

## Relation

Follow-up to the Kimi Work discovery appendix in #1145. The Windows relocated `shareDir` path is one compatibility case within this broader native-discovery feature, not a separate issue.

REMIXPLAYER:
## Summary

- Discover Kimi Work sessions natively on macOS and Windows, without requiring `TOKSCALE_EXTRA_DIRS`.
- On Windows, honor Kimi Desktop's optional `shareDir` in `kimi-desktop/daimon-storage.json` when the app has relocated its data directory.
- Fall back to the platform default Work directory when the configuration is absent or invalid, while keeping explicit `--home` scans isolated from the current user's environment roots.
- Reuse the existing Kimi Code wire parser and pricing path; no new client ID or private session fixture is introduced.

This implements the native Kimi Work discovery follow-up identified in #1145. The model aliases themselves were already landed in #1147; this branch adds the broader discovery feature, with Windows relocated-share handling as one compatibility case for real Kimi Desktop installations.

### Windows root selection

On Windows, the home-relative Work root is always retained. For normal scans, a valid `shareDir` is used as the environment-derived root; the legacy `%APPDATA%` root is used only when the configuration is missing, unreadable, invalid, or blank. A valid `shareDir` intentionally replaces that fallback rather than merging both. Explicit `--home` scans disable environment roots.

## Implementation

- Registers Kimi Work roots inside the existing Kimi scanner branch.
- Reads the optional Windows `shareDir` configuration and appends `daimon/runtime/kimi-code/home/sessions` using native path components.
- Keeps the home-relative Work root and selects the configured `shareDir` or legacy `%APPDATA%` as the environment-root candidate.
- Preserves canonical root/file de-duplication.
- Adds regression coverage for `conv-*`, `ctitle-*`, environment-root isolation, configured `shareDir`, and Linux exclusion.
- Keeps the four maintained README translations synchronized.

## Validation

Windows validation was performed against real local Kimi Work data without including private paths, session IDs, prompts, or logs in this PR:

### Windows before/after

The relocated-share case was reproduced with real local Kimi Work data; only aggregate counts and behavior are reported here:

- Before `aa7c3ad0`, the default scan found 0 Kimi Work files, while an explicit extra scan root found 4.
- After `aa7c3ad0`, the default scan followed the configured `shareDir`, found all 4 files, and produced non-zero priced usage through the existing Kimi Code parser and pricing path.

- `cargo test -p tokscale-core kimi_work --lib -- --nocapture`: 4 passed
- `cargo test -p tokscale-core kimi --lib -- --nocapture`: 53 passed
- The Windows full core suite had one unrelated pre-existing symlink test failure because the account lacked the privilege required to create directory symlinks (OS error 1314); excluding only that test, 1816 passed and 0 failed.

Additional local validation:

- `cargo test --locked --workspace --all-features --quiet`: passed
- `cargo clippy --locked --workspace --all-features -- -D warnings`: passed
- Changed-file rustfmt and `git diff --check`: passed

The repository-wide format check still reports pre-existing formatting in `crates/tokscale-core/src/sessions/opencode.rs`, which is outside this PR.

Closes #1171
Related to #1145

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -73,7 +73,7 @@
 | <img width="48px" src=".github/assets/client-senpi.png" alt="Senpi" /> | [Senpi (OmO Native)](https://github.com/code-yeongyu/senpi) | `~/.senpi/agent/sessions/` (`SENPI_CODING_AGENT_DIR` でオーバーライド可能) |
 | <img width="48px" src="https://github.com/getkimchi.png" alt="Kimchi" /> | [Kimchi Coding](https://kimchi.dev/) | `~/.config/kimchi/harness/sessions/`（`KIMCHI_CODING_AGENT_DIR` でオーバーライド可能） |
 | <img width="48px" src=".github/assets/client-synthetic.png" alt="Reasonix" /> | [Reasonix](https://github.com/esengine/DeepSeek-Reasonix) | `~/.reasonix/stats/*.jsonl`（`REASONIX_STATE_HOME` または `REASONIX_HOME` でオーバーライド可能） |
-| <img width="48px" src=".github/assets/client-kimi.png" alt="Kimi" /> | [Kimi CLI](https://github.com/MoonshotAI/kimi-cli) / [Kimi Code](https://github.com/MoonshotAI/kimi-code) | kimi-cli: `~/.kimi/sessions/` kimi-code: `~/.kimi-code/sessions/` (`KIMI_CODE_HOME` でオーバーライド可能) |
+| <img width="48px" src=".github/assets/client-kimi.png" alt="Kimi" /> | [Kimi CLI](https://github.com/MoonshotAI/kimi-cli) / [Kimi Code](https://github.com/MoonshotAI/kimi-code) | kimi-cli: `~/.kimi/sessions/` kimi-code: `~/.kimi-code/sessions/` (`KIMI_CODE_HOME` でオーバーライド可能) kimi-work: デスクトップ app-data ルート（自動検出） |
 | <img width="48px" src=".github/assets/client-qwen.png" alt="Qwen" /> | [Qwen CLI](https://github.com/QwenLM/qwen-cli) | `~/.qwen/projects/` |
 | <img width="48px" src=".github/assets/client-roocode.png" alt="Roo Code" /> | [Roo Code](https://github.com/RooCodeInc/Roo-Code) | `~/.config/Code/User/globalStorage/rooveterinaryinc.roo-cline/tasks/` (+ server: `~/.vscode-server/data/User/globalStorage/rooveterinaryinc.roo-cline/tasks/`) |
 | <img width="48px" src=".github/assets/client-kilocode.png" alt="Kilo" /> | [Kilo](https://github.com/Kilo-Org/kilocode) | `~/.config/Code/User/globalStorage/kilocode.kilo-code/tasks/` (+ server: `~/.vscode-server/data/User/globalStorage/kilocode.kilo-code/tasks/`) |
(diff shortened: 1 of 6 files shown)


junhoyeo:
Thanks — this is in good shape, and thanks for merging `main` yourself (the `Lint` red was our bug, not yours: base `ab5b5095` carried a rustfmt drift now fixed by `ee26e61d`).

I verified the parts that usually bite this repo and they are all clean: you reuse `ClientId::Kimi` rather than registering a new client, so the registry/`COUNT` invariant is untouched; the Work layout ends in `.../agents/main/wire.jsonl` so `is_kimi_code_path` routes it to `parse_kimi_code_file` correctly; canonical-path dedup prevents double counting against `TOKSCALE_EXTRA_DIRS`; `--home` isolation holds; and `kimi_work_share_dir_root` parses untrusted JSON entirely through `.ok()?`/`?` with no unwraps. No `parser_version` bump is needed since parsing is unchanged.

I pushed one small commit: `kimi_work_wire_from_root` had no `#[cfg]` gate while both of its callers do, so it compiles with zero callers on linux and warns in every ubuntu test run. Clippy would not have caught it — our lint step runs without `--all-targets`, so `#[cfg(test)]` code is never linted.

**One design question I did not want to decide for you.** In `kimi_work_roots`, the Windows branch is:

```rust
roots.push(
    kimi_work_share_dir_root(&app_data)
        .unwrap_or_else(|| join_native_path(&app_data, KIMI_WORK_SUFFIX)),
);
```

That is either/or: a configured `shareDir` *replaces* the `%APPDATA%` default rather than supplementing it. On a redirected roaming profile with a relocated share, every session written before the relocation stops being scanned — which reads to me as contradicting the "History preservation" section of your PR description. A stale-but-parseable `shareDir` has the same effect, whereas the description says fallback happens when the config is "absent or invalid" (only an unreadable file, malformed JSON, a missing key, or a blank string reach the fallback).

Pushing both and letting the existing canonical-path dedup collapse the overlap would fix it:

```rust
if let Some(configured) = kimi_work_share_dir_root(&app_data) {
    roots.push(configured);
}
roots.push(join_native_path(&app_data, KIMI_WORK_SUFFIX));
```

I left it alone because `test_scan_all_clients_kimi_work_respects_env_roots` deliberately asserts the current behaviour (`assert!(!kimi_files.contains(&conflict_wire))`), so you clearly chose it on purpose. If that is intentional, could you drop the "History preservation" wording and note on `kimi_work_roots` that a configured `shareDir` is authoritative and abandons `%APPDATA%` history? Either resolution is fine — I just want the code and the description to agree.

REMIXPLAYER:
Thanks for the careful review and for catching the mismatch between the implementation and the PR description.I also appreciate the focused follow-up fix in `3ea21b06` and that you left the history-preservation trade-off for me to decide.


I have verified the Windows behavior with a valid `shareDir`: the home-relative root remains the default Work location, while a valid `shareDir` replaces the legacy `%APPDATA%` fallback for the environment-derived root. Explicit `--home` scans continue to ignore environment roots. `shareDir` is Kimi Desktop’s optional relocated data directory, not its application installation directory.

I have kept this behavior for this PR because scanning both roots could include stale or pre-migration sessions. This intentionally does not guarantee history preservation across data-directory migrations. I have removed that claim from the PR description and updated the scanner comments accordingly; the scanning implementation remains unchanged.

PR #1180, opened 2026-08-23
fix(cli): work around an Antigravity RPC sync hang on Windows via curl.exe fallback

### LINKED ISSUE #1129: Antigravity sync on Windows: #1127 could not have fixed it, and every RPC re-pays a doomed plaintext leg
Two findings, one of which invalidates a merged fix.

## 1. #1127's stated mechanism is impossible in this build

#1127 explained a Windows `antigravity sync` hang as: reqwest negotiates HTTP/2, the DesktopAgent advertises h2 but freezes when multiplexed, reqwest hangs to the 10s timeout. It added `.http1_only()` to the RPC client.

That cannot happen here. `reqwest` is declared `default-features = false` with only `json` and `native-tls-vendored`, so `http2` is off:

```
hyper::proto::h2 symbols:  0
hyper::proto::h1 symbols:  188
h2 in Cargo.lock:          absent
```

Neither `reqwest` 0.12.28 nor `hyper` 1.8.1 pulls `h2` in this tree, and hyper's HTTP/2 protocol module is not in the linked binary. The client never offers h2 over ALPN and has no HTTP/2 implementation. **`.http1_only()` is a no-op**, so #1127 did not fix what it claimed, and the reported hang is presumably still present.

Two signals pointed at this before merge and were not weighed: the branch was named `fix-windows-tls-renegotiation` while the body argued h2 multiplexing — two incompatible theories, neither verified — and CI could only ever show the change was harmless, which is indistinguishable from a no-op.

The `.http1_only()` line is harmless and stays as a guard if `http2` is ever enabled, but its comment asserted the false mechanism and has been corrected.

## 2. A real amplifier: the transport is re-derived on every RPC

`rpc_request` tries plaintext first and falls back to TLS. Nothing remembered the result — `AntigravityConnection` had no transport field, and `probe_heartbeat` learned the answer during discovery and discarded it.

Against a TLS-only DesktopAgent, every RPC re-pays the doomed plaintext leg. `try_fetch_session_artifact` issues **one RPC per session**, so the waste scales with the user's whole history. Timeouts are 10s read on the plaintext leg and 10s on the HTTPS client, so the worst case is ~20s per session — tens of sessions reads as "hangs indefinitely".

A regression test makes the cost concrete: with the fix reverted, two `rpc_request` calls produce two plaintext connection attempts against a port already known to speak TLS.

## Still open

Whether a plaintext write to the DesktopAgent's TLS listener on Windows is **reset immediately** or **stalls until the 10s read timeout**. Most TLS stacks reject a malformed first record at once, which would make the plaintext leg cheap and mean the amplifier above is not the whole story. This is the one measurement that decides it, and it needs a Windows box with Antigravity running — neither of which is available here.

## Asks

- @marshallcodes: on current `main`, does `tokscale antigravity sync` still hang, and roughly how long does a single RPC take?
- Treat #1127 as unverified rather than fixed.

marshallcodes:
Closes #1129

## Problem

On Windows, querying local Antigravity language server RPC endpoints
over HTTPS hangs indefinitely. The Windows SChannel / tokio-native-tls
stack cannot complete the TLS handshake against the local server,
causing commands like `antigravity sync` to time out.

## Solution

On Windows targets, `https_rpc_request` now routes HTTPS requests through
`curl.exe` instead of using `reqwest` / native TLS.

The command uses:

- `-k` (`--insecure`): acceptable here because the endpoint is always
  `https://127.0.0.1:<port>/...` and the local Antigravity language server
  presents a self-signed certificate that Windows SChannel cannot validate.
  This path never targets remote hosts.
- `--http1.1`: avoids HTTP/2 negotiation issues with the local server.
- `--max-time 10`: prevents the previous indefinite hang.

If `curl.exe` is unavailable, the function returns a clear error instead
of falling back to the old hanging path.

Also adds `#[cfg(not(target_os = "windows"))]` guards to unused
`reqwest` / `tokio` helpers that are only referenced on non-Windows paths.

## Testing

Tested manually on Windows with active Antigravity language server
connections.

Before:
```
    HTTPS RPC failed... Operation timed out after 10008 milliseconds with 0 bytes received
```
After:
```
    Antigravity sync
    detected connections: 2
    cached sessions after sync: 158
```

`cargo build --target x86_64-pc-windows-msvc` produces no warnings.

No automated test exercises the live Antigravity RPC TLS handshake yet.
The curl argument construction is isolated and could be unit-tested later.

---
## Summary by cubic
Fixes Windows hangs in Antigravity CLI HTTPS RPC by invoking the in-box `curl.exe` for local calls. Old behavior: TLS handshake stalled and could treat error JSON as success. New behavior: Windows shells out to `curl.exe` over HTTP/1.1 with a 10s cap, validates HTTP status, and keeps the CSRF token off argv; non-Windows is unchanged.

- Windows-only path uses `%SystemRoot%\System32\curl.exe`; returns a clear error if missing.
- Stays local to `127.0.0.1`, accepts the self-signed cert with `-k`, pins `--http1.1`, and sets `--max-time 10`.
- Sends the CSRF header and body via stdin using `-K -` (escaped) to avoid argv leaks; caps stdout at `MAX_RPC_BODY_BYTES`; fails on non-2xx with method-aware errors.
- Adds `#[cfg(not(target_os = "windows"))]` around unused `reqwest`/`tokio` helpers and tests to silence Windows build warnings.
- No migration required; ensure `curl.exe` is present on Windows at runtime.

<sup>Written for commit 06525a04813b6ad0bf90cd99710f47b093fc1278. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1180?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/antigravity.rs
+++ b/crates/tokscale-cli/src/antigravity.rs
@@ -1938,6 +1938,40 @@ fn https_rpc_request(
     method: &str,
     body: &Value,
 ) -> Result<Value> {
+    // The idea is to bypass tokio-native-tls SChannel renegotiation deadlocks on Windows
+    // by giving the task to the native OS curl.exe
+    #[cfg(target_os = "windows")] {
+        let url = format!(
+            "https://127.0.0.1:{}/exa.language_server_pb.LanguageServerService/{}",
+            connection.port, method
+        );
+        let body_str = serde_json::to_string(body)?;
+
+        let output = std::process::Command::new("curl.exe")
+            .args([
+                "-k", "-sS", "--http1.1", "--max-time", "10",
+                "-X", "POST", &url,
+                "-H", "Content-Type: application/json",
+                "-H", "Connect-Protocol-Version: 1",
+                "-H", &format!("X-Codeium-Csrf-Token: {}", connection.csrf_token),
+                "-d", &body_str,
+            ])
+            .output()
+            .with_context(|| "Failed to execute curl.exe for Windows RPC fallback")?;
+
+        if !output.status.success() {
+            anyhow::bail!(
+                "Windows curl.exe RPC fallback failed (exit code {}): {}",
+                output.status.code().unwrap_or(0),
+                String::from_utf8_lossy(&output.stderr)
+            );
+        }
+
+        let response_body = String::from_utf8_lossy(&output.stdout);
+        Ok(serde_json::from_str(&response_body)?)
+    }
+
+    #[cfg(not(target_os = "windows"))]
     antigravity_https_runtime().block_on(async {
         let url = format!(
             "https://127.0.0.1:{}/exa.language_server_pb.LanguageServerService/{}",
--- a/crates/tokscale-cli/src/antigravity.rs
+++ b/crates/tokscale-cli/src/antigravity.rs
@@ -1948,17 +1948,43 @@ fn https_rpc_request(
     method: &str,
     body: &Value,
 ) -> Result<Value> {
-    // The idea is to bypass tokio-native-tls SChannel renegotiation deadlocks on Windows
-    // by giving the task to the native OS curl.exe
+    // On Windows, the reqwest (rustls) request to the local DesktopAgent
+    // stalls until the timeout, while the in-box curl.exe (SChannel) completes
+    // the same request instantly (#1129). The mechanism is unverified, so the
+    // request is handed to curl.exe instead of guessing at a rustls-side fix.
+    // `--http1.1` is a defensive pin: unlike this workspace's reqwest build
+    // (no `http2` feature), curl can negotiate h2 via ALPN.
     #[cfg(target_os = "windows")]
     {
+        // curl config-file values are enclosed in double quotes, inside which
+        // backslashes and double quotes must be backslash-escaped.
+        fn curl_config_quote(value: &str) -> String {
+            let mut quoted = String::with_capacity(value.len() + 2);
+            quoted.push('"');
+            for ch in value.chars() {
+                if matches!(ch, '\\' | '"') {
+                    quoted.push('\\');
+                }
+                quoted.push(ch);
+            }
+            quoted.push('"');
+            quoted
+        }
+
         let url = format!(
             "https://127.0.0.1:{}/exa.language_server_pb.LanguageServerService/{}",
             connection.port, method
         );
         let body_str = serde_json::to_string(body)?;
 
-        let output = std::process::Command::new("curl.exe")
+        // Resolve curl.exe by absolute path: a PATH lookup can be shadowed by
+        // a user-writable directory, handing the CSRF token to an impostor.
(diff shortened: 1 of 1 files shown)


junhoyeo:
Thanks for chasing this — #1129 has been open a while and a real Windows repro is valuable. Before this goes in, the stated root cause needs re-checking, because it describes a stack that is no longer in the tree.

The problem statement says the Windows SChannel / `tokio-native-tls` stack cannot complete the handshake. `native-tls` was replaced with rustls in #1138, and your branch already contains that commit — `Cargo.toml:65` on your own branch reads `features = ["json", "rustls-tls-native-roots"]`. Checking the dependency graph for the Windows target on current main:

```
$ cargo tree -p tokscale-cli --target x86_64-pc-windows-msvc -i native-tls
error: package ID specification `native-tls` did not match any packages

$ cargo tree -p tokscale-cli --target x86_64-pc-windows-msvc -i schannel
schannel v0.1.28
└── rustls-native-certs v0.8.4
```

So the handshake is performed by `rustls` + `ring`, and `schannel` survives only underneath `rustls-native-certs`, which reads the OS trust store rather than doing any TLS. On this specific client that is moot anyway: `danger_accept_invalid_certs(true)` installs a no-op verifier under rustls, so the trust store is never consulted for these requests.

That matters because it makes the described failure mode impossible on current main, which means either the measurement predates the rustls migration, or the hang has a different cause and the fix would be treating a symptom.

Worth noting the specific case you describe does work under rustls. When I migrated in #1138 the loopback self-signed certificate was the blocker I had to clear, since rustls is stricter than OpenSSL and rejects certificates with no SAN outright. Reproducing `antigravity_https_client`'s exact configuration — `http1_only()`, `danger_accept_invalid_certs(true)`, `no_proxy()`, 10s timeout — against `openssl s_server`: 200 OK against a self-signed cert with an IP SAN for 127.0.0.1, 200 OK against a self-signed cert with no SAN at all, and a fast `Connection refused` rather than a hang against an unused port. That was on macOS, but rustls is pure Rust with no platform TLS backend, so there is no Windows-specific handshake path left to differ.

Could you re-test on a current main build before we take this? If it still hangs, that is a genuinely new finding and worth capturing properly — the timeout wording in your "before" output (`Operation timed out after 10008 milliseconds with 0 bytes received`) reads like curl's own message rather than reqwest's, so it would help to see the reqwest-side error too.

If rustls does resolve it, I would rather not add the `curl.exe` path: it introduces a runtime dependency on an external binary, moves TLS out of the process, and `-k` disables verification in a way that is harder to audit than the in-process verifier it replaces.

PR #1183, opened 2026-08-23
fix(cursor): extend timeout to 120s for explicit sync of large CSV exports

### LINKED ISSUE #1175: cursor sync times out at 15s on large Cursor CSV exports
`tokscale cursor sync` still fails on a valid Cursor session when the usage CSV is larger than the 15s HTTP timeout. The error is `error decoding response body`, which looks like a parse bug but is reqwest aborting while reading the body.

This is a follow-up to #1024 / #1025. Bumping 8s → 15s was not enough for a full `export-usage-events-csv?strategy=tokens` dump.

## Problem

[`CURSOR_HTTP_TIMEOUT`](https://github.com/junhoyeo/tokscale/blob/main/crates/tokscale-cli/src/cursor.rs) is hardcoded at 15s and is shared by:

- implicit pre-report auto-sync (where a short bound makes sense)
- explicit `tokscale cursor sync` (where the user is waiting for a full export)

There is no CLI flag or env override. `TOKSCALE_NATIVE_TIMEOUT_MS` does not apply here.

## Repro

1. `tokscale cursor login` with a valid `WorkosCursorSessionToken` (`tokscale cursor status` reports Session: Valid).
2. `tokscale cursor sync --json`

```json
{"synced":false,"rows":0,"error":"<account>: error decoding response body"}
```

The process returns in ~15.4s every time. Same on 4.8.1 and 4.13.0.

The same cookie against `https://cursor.com/api/dashboard/export-usage-events-csv?strategy=tokens` returns `200 text/csv` and a body that starts with `Date,`. Streaming that response to `~/.config/tokscale/cursor-cache/usage.csv` succeeded:

- 9,210,683 bytes
- 80,598 rows
- ~38s at ~240 KB/s

`tokscale submit -c cursor` then works from the cache (`Cursor sync failed; using cached data`).

## Expected

Explicit `tokscale cursor sync` should finish a multi-tens-of-seconds CSV download, or fail with a timeout message that says so. A 15s cap should not apply to a command whose only job is to fetch the full export.

## Suggested fix

Any of:

- Use a longer timeout (or no body timeout) on the explicit `cursor sync` path, keep 15s for auto-sync
- Stream the CSV to disk instead of `response.text().await` into a String
- Honor something like `TOKSCALE_CURSOR_TIMEOUT_MS`

Happy to test a build.

aruru10313:
Resolves #1175.

## Background
Previously, `tokscale cursor sync` used a hardcoded 15-second timeout for the Cursor CSV export endpoint. While this tight timeout makes sense for pre-report auto-syncs (to prevent the TUI from stalling indefinitely on a hung network), it causes explicit `tokscale cursor sync` commands to fail prematurely with an `error decoding response body` when downloading large usage histories.

## Changes
This PR distinguishes between explicit manual syncs and implicit auto-syncs by passing an `explicit: bool` flag down to `sync_cursor_cache`. 
- Manual `tokscale cursor sync` commands now use a **120-second timeout**, allowing large CSV exports to download completely.
- Background auto-syncs retain the fast **15-second cap**.

> Hope this is helpful for users with larger Cursor histories! Let me know if there's anything else needed here. 😊

### DIFF
--- a/crates/tokscale-cli/src/commands/wrapped.rs
+++ b/crates/tokscale-cli/src/commands/wrapped.rs
@@ -182,7 +182,7 @@ async fn load_wrapped_data(options: &WrappedOptions) -> Result<WrappedData> {
     let mut cursor_sync_result: Option<cursor::SyncCursorResult> = None;
 
     if include_cursor && cursor_logged_in {
-        cursor_sync_result = Some(cursor::sync_cursor_cache().await);
+        cursor_sync_result = Some(cursor::sync_cursor_cache(false).await);
     }
 
     if let Some(sync) = cursor_sync_result.as_ref() {
--- a/crates/tokscale-cli/src/cursor.rs
+++ b/crates/tokscale-cli/src/cursor.rs
@@ -11,7 +11,15 @@ use std::time::{Duration, SystemTime};
 /// auto-sync (which runs synchronously before local reports and the TUI) while
 /// still tolerating routine API latency. If the network is hung, the report
 /// proceeds against cached data after this timeout instead of stalling forever.
+/// Explicit `tokscale cursor sync` overrides this for the usage-CSV download
+/// with [`CURSOR_EXPLICIT_SYNC_TIMEOUT`].
 const CURSOR_HTTP_TIMEOUT: Duration = Duration::from_secs(15);
+
+/// Per-request timeout for the usage-CSV download during an explicit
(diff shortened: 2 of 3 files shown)


aruru10313:
Hope this is helpful for users with larger Cursor histories! Let me know if there's anything else needed here. 😊

PR #1214, opened 2026-08-27
perf(cache): compress source-message shard payloads

### LINKED ISSUE #1209: Excessive memory usage
xref #1100 

```
npx -y tokscale@4.14 submit
```

After running for a long time, it still OOMed, and this time v3 has the same issue.

Image

junhoyeo:
Shard payloads were stored as bare bincode. On a real 14 GB OpenCode database the shard holding 340,216 messages is 83 MB, and that same payload compresses to 15.6 MB with zstd at level 3 — 5.6x. zstd is already a dependency of this crate.

The reason this matters is not disk. `MAX_CACHE_SHARD_BYTES` gates both reading and writing a shard, so a source whose payload crosses 256 MiB stops being cacheable at all: `save_if_dirty` logs "future scans may remain cold" exactly once and the source is re-parsed on every subsequent scan, forever, with no further warning. Compression moves that cliff roughly 5.6x further out.

The two limits are now separate, because compression made them different quantities. `MAX_CACHE_SHARD_BYTES` continues to bound what lands on disk. `MAX_CACHE_SHARD_PAYLOAD_BYTES` bounds the uncompressed payload before it is compressed; holding the serialize limit at the on-disk cap would have refused payloads that fit on disk several times over, which is the failure this is meant to remove.

Old caches stay readable. The reader sniffs the zstd frame magic instead of relying on a format-version bump, so a pre-compression shard is passed through unchanged and rewritten compressed on its next save. Bumping the format version would have invalidated every existing shard and forced precisely the full re-parse this change exists to avoid — worst for the heavy users it is meant to help.

Measured on the real 83 MB shard: compression 0.08s, decompression 0.05s, against the tens of seconds a re-parse of the same source costs.

One existing test needed its fixture changed rather than its assertion. `test_aggregate_cache_can_exceed_individual_shard_limit` built its payload from `"a".repeat(20 * 1024)`, which compresses to almost nothing, so both shards fell under the cap and the test could no longer observe the property it exists to check. It now uses high-entropy content of the same length, so each shard stays a real fraction of the limit and the aggregate still exceeds it.

Scope note: this does not fix #1209. Peak RSS during a scan is dominated by live in-memory messages (measured at 1.37 GB live in 12.6 million blocks for a single OpenCode scan), not by cache size. This removes the cliff that makes a heavy source cold on every run, which is an amplifier of that cost rather than its cause.

Refs #1209

---
## Summary by cubic
Compresses source-message shard payloads with zstd so heavy sources stay cacheable instead of falling off a hard size cliff and being re-parsed on every scan.

- Compression keeps old caches readable: the reader sniffs the zstd frame magic and passes pre-compression shards through unchanged.
- `MAX_CACHE_SHARD_PAYLOAD_BYTES` now bounds the uncompressed payload separately from `MAX_CACHE_SHARD_BYTES`, which still bounds the on-disk size.
- Does not fix #1209; peak RSS during scans is dominated by live in-memory messages, not cache size.

<sup>Written for commit 59a1d8d7c64170526f3266b79f05c2f4b42e0f39. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1214?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -41,6 +41,41 @@ const CACHE_LOCK_FILENAME: &str = "source-message-cache.lock";
 const LEGACY_MONOLITH_CACHE_FILENAME: &str = "source-message-cache.bin";
 const CACHE_SHARD_COUNT: usize = 256;
 const MAX_CACHE_SHARD_BYTES: u64 = 256 * 1024 * 1024;
+
+/// Ceiling on the *uncompressed* entry payload inside a shard.
+///
+/// [`MAX_CACHE_SHARD_BYTES`] bounds what lands on disk. Since the payload is
+/// zstd-compressed the two are no longer the same number: real shards compress
+/// about 5.6x, so holding the serialize limit at the on-disk cap would refuse
+/// payloads that fit on disk several times over and send a heavy source back to
+/// re-parsing every scan. This bounds the pre-compression side on its own.
+const MAX_CACHE_SHARD_PAYLOAD_BYTES: u64 = 4 * MAX_CACHE_SHARD_BYTES;
+
+/// zstd frame magic (little-endian `0xFD2FB528`).
+///
+/// Shards written before compression hold a bare bincode payload, so the reader
+/// sniffs this rather than relying on a format bump: an old cache stays readable
+/// and is rewritten compressed on the next save, instead of being invalidated
+/// and forcing exactly the full re-parse this is meant to avoid.
+const ZSTD_MAGIC: [u8; 4] = [0x28, 0xB5, 0x2F, 0xFD];
+
+/// Compression level for shard payloads. Level 3 is zstd's default: on a real
+/// 83 MB shard it takes ~0.08s to write and ~0.05s to read back, against the
+/// tens of seconds a re-parse of the same source costs.
+const SHARD_COMPRESSION_LEVEL: i32 = 3;
+
+fn compress_shard_payload(payload: &[u8]) -> std::io::Result<Vec<u8>> {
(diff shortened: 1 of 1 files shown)


junhoyeo:
Closing this one.

The cliff it describes is real: `MAX_CACHE_SHARD_BYTES` gates reading *and* writing, so a source whose payload crosses 256 MiB stops being cacheable and gets re-parsed on every scan forever. That is worth fixing.

But this fixes it in the direction that hurts the profile it cites. `MAX_CACHE_SHARD_PAYLOAD_BYTES` is `4 * MAX_CACHE_SHARD_BYTES` = 1 GiB, and `decompress_shard_payload` materialises the whole payload into a second `Vec<u8>` before bincode ever sees it. So on a heavy profile the warm path gains a transient buffer of up to 1 GiB, held alongside the compressed bytes it was decoded from — on exactly the scan that #1209 reports as OOM-killed. Raising the ceiling on a source that is already dying at the ceiling is the wrong trade, even though the disk-side reasoning is sound.

The cold-forever problem is better attacked by not re-reading the source at all, which is what #1212 does for OpenCode: an incremental mark means a changed database costs what an unchanged one costs, so the shard never has to grow to stay useful. If a size cliff still bites after that, the fix should shrink the resident payload, not enlarge the buffer that holds it.

Not deleting the branch — the zstd-magic sniffing for backward compatibility is the good part and worth lifting if this comes back in a form that lowers peak instead of raising it.

PR #1232, opened 2026-08-29
fix(opencode): track incremental rows by provenance

### LINKED ISSUE #1230: OpenCode incremental scan infers safety instead of checking it
## What

The incremental rescan's safety is entirely **inferred**. A cached message
records nothing about the SQLite row that produced it, so deletions,
insertions, key changes and parser rejections are all deduced from table
statistics and a growing set of targeted probes. Every time one of those
deductions leaks, the answer has been another probe — and the set is not
closing.

Findings raised against this one subsystem across four rounds of adversarial
review:

| Round | Finding | Status |
|---|---|---|
| 1 | A renamed session leaves cached titles and workspaces stale | fixed in #1229 |
| 2 | v1 and v2 overwrite each other's metadata for a shared session id | fixed in #1229 |
| 2 | `prepared()` treats an execute failure as success | fixed in #1229 |
| 3 | `ran()` is true even after a step error truncated the scan | fixed in #1229 |
| 3 | A row that gains an embedded `$.id` is counted twice | fixed in #1229 |
| 4 | An embedded id changing **A → B**, or being lost | **open** |
| 4 | The SQL qualification predicate does not match the parser's | **open** |
| 4 | A non-monotonic `time_created` bypasses the deletion guard | **open** |

## Why patching them individually is the wrong move

Each of the three open findings could be answered with another probe, and the
next round would surface a fourth. The evidence for that is the table above:
every round fixed what it found, and every round found more.

## The design that converges

Storing a **row_id → dedup_key** map in the incremental state turns all four
into decisions rather than deductions:

- **Deletion** — a stored row id absent from the current set; remove exactly
  that message instead of refusing the whole scan
- **Insertion** — a row id not previously stored
- **Key change** — the same row id now yields a different key; replace rather
  than append
- **Parser rejection** — a changed row produces no message; drop the one it
  used to back

It also *replaces* three existing probes (the deletion guard, the
disqualification probe, the re-key probe) rather than adding a fourth.

**Measured cost** against the real 14.9 GB database (434,955 rows):

- Reading every row id: **0.76 s** — cheap because no JSON is parsed
- Total id bytes: **13 MB**, about 5% of the 256 MiB cache shard ceiling

For comparison, the alternative of counting qualifying rows inside the stats
query measured **+14.6 s** per rescan.

## Current exposure

All three open findings have **zero instances** on that same database:

```
rows carrying an embedded $.id                       0 / 434,955
assistant rows with tokens but no modelID            0
assistant rows with tokens but no time               0
assistant rows with tokens but no cache              0
```

So they are real but unobserved, which is why the incremental path ships
enabled in v4.15.0. If anyone reports a warm scan disagreeing with a cold one,
this issue is the first place to look.

## Failure mode

Silent double counting or permanent under-counting: the leaderboard number is
wrong and nothing signals it. That is the worst shape of defect this project
can have, which is why low exposure is a reason to schedule the work rather
than a reason to skip it.

IvGolovach:
## Summary

- Persist explicit physical-row provenance for every incrementally parsed OpenCode SQLite message.
- Reconcile the previous and current row inventories in one linear pass so deletions, parser rejections, and dedup-key changes remove the exact cached contribution they replaced.
- Preserve conservative full-scan fallbacks for ambiguous multi-row merges, schema/query changes, non-monotonic deltas, and concurrent inventory drift.
- Migrate the source-message cache from format 6 to 7 without discarding cached messages; OpenCode rebuilds only its provenance map on the next changed-database scan.

Fixes #1230.

## Why

The previous incremental path inferred safety from aggregate row counts and timestamp high-water marks. That could detect several unsafe cases, but it could not prove which physical SQLite row backed a cached usage message. A deletion hidden by an insertion, a row that stopped qualifying as assistant usage, or a row whose embedded message ID changed could therefore require a full parse merely to avoid stale or duplicated accounting.

This change makes the invariant explicit: every cached OpenCode contribution is tied to the source row that produced it. Warm scans can now update or remove exact contributions when the row inventory proves the operation safe, while uncertain cases still fail closed to the existing full-scan path.

## Diff scope

- `crates/tokscale-core/src/sessions/opencode_schema.rs`: adds ordered row inventories, row-to-message provenance, exact incremental reconciliation, and conservative ambiguity guards.
- `crates/tokscale-core/src/sessions/opencode.rs` and `crates/tokscale-core/src/lib.rs`: adds cold-versus-warm regression coverage for deletion, replacement, rejection, rekeying, backward timestamps, and metadata refresh behavior.
- `crates/tokscale-core/src/message_cache.rs`: bumps cache format 6 to 7 and migrates version-6 entries while retaining their parsed messages.
- No CLI surface, frontend behavior, database schema, workflow, dependency, or release configuration changed.

## Branch integrity

- Base: `main`
- Head SHA: `1dd7a64dc81b1c28ffd29f60fc1497a6bc1809c8`
- Ahead/behind: `0 behind / 1 ahead`
- `origin/main` is an ancestor of the head; diff proof was computed against the freshly fetched base.

## Commit integrity

- One logical commit: `1dd7a64dc81b1c28ffd29f60fc1497a6bc1809c8 fix(opencode): track incremental rows by provenance`
- The final diff contains only the four intended OpenCode/cache source and test files listed above.
- Version: not applicable at PR stage — the repository publish workflow synchronizes runtime package versions.

## Diff hygiene

## Validation mode and proof

Validation mode: Mode 3 — this changes persisted-cache interpretation and incremental usage-accounting correctness.

- `cargo check -p tokscale-core`: PASS.
- `cargo test -p tokscale-core --lib`: PASS — 1,941 passed, 2 ignored, 0 failed.
- Workspace Clippy was not run successfully: the local Rust 1.86 toolchain is below the Rust 1.88 MSRV of all-target development dependencies, and the narrower core all-target run reaches a pre-existing `needless_lifetimes` warning in unmodified `sessions/pi.rs`.
- Required remote CI: pending until the PR is published. This PR is ready for review, not merge-ready until required checks and review gates pass on the final SHA.

## Cache and migration notes

- This is an application cache-format migration, not a database migration.
- Version-6 cache shards retain their parsed messages and discard only the OpenCode incremental mark that lacks trustworthy row provenance.
- The next changed OpenCode database scan performs one full reconstruction of the provenance map and rewrites the shard in format 7.
- DB downgrade: not applicable.
- Data repair: not applicable; the cache can be safely regenerated from source data.

## CI context confirmation

- No workflow files or CI context names changed.
- Pending — required repository CI can run only after the PR is published.

## Runtime safety

- The row inventories are sorted and reconciled in one linear pass; the implementation avoids a second database-sized row hash map and avoids quadratic lookups.
- No new blocking locks, background workers, or unbounded queues were introduced.
- Ambiguous multi-row provenance, query/schema drift, missed non-monotonic updates, and concurrent inventory changes continue to fall back to a full scan.

## Documentation integrity

Not applicable — no user-facing commands, public API, operational procedure, or runbook changed.

## Rollback plan

- Rollback: revert this PR.
- DB downgrade: not applicable.
- Operational caveats: reverting causes format-7 cache shards to be treated as stale and rebuilt from source data.

## Known residual risks

- The first changed OpenCode scan after upgrading a version-6 shard intentionally performs a full parse to establish trustworthy row provenance.

---
## Summary by cubic
Fixes #1230 by persisting exact row provenance for every incrementally parsed OpenCode SQLite message, so warm scans now handle deletions, parser rejections, and dedup-key changes without falling back to full parses. Previously, incremental safety relied on aggregate row counts and high-water marks, which couldn't prove which physical row produced a cached message.

- Reconciles previous and current row inventories in one linear pass, removing cached messages that no longer have a backing row.
- Conservative full-scan fallback remains for ambiguous multi-row merges, schema/query changes, non-monotonic deltas, and concurrent inventory drift.
- Bumps the source-message cache from format 6 to 7; version-6 shards keep their parsed messages and rebuild the provenance map on the next changed-database scan.
- Adds randomized mutation tests that verify warm rescans match full parses across inserts, deletes, rewrites, rekeys, and disqualifications.
- Applies a Clippy style fix with no behavior change.

<sup>Written for commit 1d9fdb65d6aa5de40359a1a89828ec7a3b36d27a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1232?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -9889,7 +9889,7 @@ mod tests {
 
     #[test]
     #[serial_test::serial]
-    fn test_opencode_warm_scan_re_parses_after_a_row_is_deleted() {
+    fn test_opencode_warm_scan_removes_a_deleted_row_incrementally() {
         let cache_home = tempfile::TempDir::new().unwrap();
         let source_home = tempfile::TempDir::new().unwrap();
         let _cache_env = redirect_cache_home(cache_home.path());
@@ -9908,9 +9908,8 @@ mod tests {
         );
         assert_eq!(cold.len(), 2);
 
-        // OpenCode cascades a session delete onto its messages. An incremental
-        // scan cannot see the row that went away, so the guard has to force a
-        // full re-parse rather than keep counting it.
+        // OpenCode cascades a session delete onto its messages. The persisted
+        // row provenance lets the warm scan remove exactly that cached source.
         conn.execute("DELETE FROM message WHERE id = 'msg-zzz'", [])
             .unwrap();
 
@@ -9924,8 +9923,8 @@ mod tests {
         assert_eq!(
             sessions::opencode_schema::INCREMENTAL_RESCANS
                 .load(std::sync::atomic::Ordering::Relaxed),
-            rescans_before,
-            "a deletion must not be served by an incremental scan"
+            rescans_before + 1,
+            "an ordinary deletion should stay incremental"
         );
         assert_eq!(warm.len(), 1);
         assert_eq!(warm[0].dedup_key.as_deref(), Some("msg-aaa"));
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -32,9 +32,13 @@ use std::time::UNIX_EPOCH;
 // Both older layouts have wire migrations below, so nothing is discarded --
 // which matters most for Claude, whose entries are the only copy of compacted
 // history.
-const CACHE_FORMAT_VERSION: u32 = 6;
+// 7: OpenCode incremental marks record row provenance. Version-6 shards have a
+// wire migration below: unrelated clients retain their cache, while OpenCode
+// keeps its parsed messages and takes one full scan to acquire the new map.
+const CACHE_FORMAT_VERSION: u32 = 7;
 const LEGACY_CACHE_FORMAT_VERSION_V4: u32 = 4;
 const LEGACY_CACHE_FORMAT_VERSION_V5: u32 = 5;
+const LEGACY_CACHE_FORMAT_VERSION_V6: u32 = 6;
 // V2 intentionally starts cold and leaves source-message-cache.bin untouched:
 // the monolith did not record a trustworthy parser owner for migration.
 const CACHE_SHARD_DIRNAME: &str = "source-message-cache-v2";
@@ -1429,6 +1433,54 @@ impl From<LegacyCachedSourceEntryV5> for CachedSourceEntry {
     }
 }
 
+/// Exact version-6 OpenCode mark layout. The aggregate/probe-based state has
+/// no trustworthy row provenance to synthesize during migration, so OpenCode
+/// entries keep their parsed messages but drop only this mark and rebuild it
+/// on the next changed-database scan.
+#[derive(Debug, Serialize, Deserialize)]
+struct LegacyOpenCodeGroupMarkV6 {
+    query_digest: u64,
+    row_count: i64,
+    created_high_water: i64,
+    updated_high_water: i64,
+    metadata_high_water: i64,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct LegacyOpenCodeIncrementalStateV6 {
+    groups: Vec<Option<LegacyOpenCodeGroupMarkV6>>,
+    merged_dedup_keys: Vec<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct LegacyCachedSourceEntryV6 {
+    parser_namespace: String,
+    parser_version: u32,
+    path: CachedPath,
+    fingerprint: SourceFingerprint,
+    messages: Vec<UnifiedMessage>,
+    fallback_timestamp_indices: Vec<usize>,
+    codex_incremental: Option<CodexIncrementalCache>,
+    prime_accounting: Option<crate::sessions::prime_agent::PrimeFileAccounting>,
+    opencode_incremental: Option<LegacyOpenCodeIncrementalStateV6>,
+}
+
+impl From<LegacyCachedSourceEntryV6> for CachedSourceEntry {
+    fn from(entry: LegacyCachedSourceEntryV6) -> Self {
+        Self {
+            parser_namespace: entry.parser_namespace,
+            parser_version: entry.parser_version,
+            path: entry.path,
+            fingerprint: entry.fingerprint,
+            messages: entry.messages,
+            fallback_timestamp_indices: entry.fallback_timestamp_indices,
+            codex_incremental: entry.codex_incremental,
+            prime_accounting: entry.prime_accounting,
+            opencode_incremental: None,
+        }
+    }
+}
+
 impl CachedSourceEntry {
     pub(crate) fn new(
         identity: CacheIdentity,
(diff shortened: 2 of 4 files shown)


junhoyeo:
Reviewed and verified locally. This is the right shape, and it does something I
had not realised was needed.

**Full gate green** on the branch: `cargo fmt --check`, `cargo test --workspace`,
and clippy both ways (bare, which catches `dead_code`, and `--all-targets`,
which catches lints in test code).

**It replaces rather than accumulates.** The four probes I had layered on over
successive review rounds — `disqualified_row_backs_cached_message`,
`rekeyed_row_backs_cached_message` and their two queries — are gone, not
supplemented. That was the point of #1230: each probe approximated "did the
cold-scan result change in a way I cannot see", and each round found another
leak. Row provenance makes it a decision instead of an approximation.

**The v6 → v7 migration is conservative in the right direction.** It mirrors the
exact legacy layout, keeps the cached messages, and discards only the OpenCode
incremental state — which has no provenance to synthesise — costing one full
scan rather than a cold cache.

## The measurement that convinced me

Named scenarios pin the cases someone thought of, so I wrote a randomised
warm-versus-cold equivalence check instead: 48 seeds, each applying four random
mutations drawn from insert / delete / rewrite / re-key / disqualify, then
asserting a warm rescan equals a cold parse of the same state.

Both branches pass it. But counting *how often the warm path actually resumes*
rather than falling back tells a different story:

| | resumed incrementally | fell back to a full scan |
|---|---|---|
| `main` | **0 / 48** | 48 |
| this PR | **43 / 48** | 5 |

On `main` the incremental path never fires. My probes were safe but so
conservative that any real mutation refused, which left #1212's optimisation
effectively inert — and I never noticed, because every check I wrote asked
"does it refuse when it must?" and none asked "how often does it refuse?".

This PR resumes in 43 of 48 and agrees with a cold parse in all 43.

## One suggestion

Worth adding that randomised check to the suite. The named tests here are good
and cover the three findings precisely, but they cannot catch a mutation
sequence nobody enumerated — which is exactly how the last four rounds of
findings arrived. Happy to open it as a follow-up if you would rather keep this
PR focused.

Approving on the strength of the gate, the probe removal, the migration
handling, and the resume-rate measurement above.

junhoyeo:
Pushed the two `clippy::question_mark` fixes to this branch — both were `match` arms returning `None`, replaced with `?`. No behaviour change, and `cargo test --workspace` plus clippy in both forms are green against CI's toolchain.

Worth knowing why they got through: the repo pins no toolchain, so a local checkout runs whatever `rustup` last installed while CI takes `dtolnay/rust-toolchain@stable`. Here that was **1.93 locally against 1.98 in CI** — the *same* clippy command, different lint sets. I hit this reviewing your PR and my own gate had the same blind spot all session.

Filed #1237 to pin it.

junhoyeo (approved):
Reviewed via subagent: replaces #1212 inference-based incremental safety with explicit per-row provenance; does not revert #1085 Claude retention (namespace-gated, preserved in V6->V7 migration) or #624; handles A->B rekey exactly and fails closed to full scan on any ambiguity. No double-count reintroduction.

PR #1233, opened 2026-08-29
perf(copilot): stream VS Code messages into submit fold

IvGolovach:
## Summary

- Stream projected Copilot VS Code messages directly into the submit fold instead of materializing a second `Vec<UnifiedMessage>` for the entire lane.
- Release already parsed Copilot OTEL/desktop messages after projecting the deduplication state needed by the VS Code lane.
- Reduce the submission pricing batch from 65,536 to 4,096 full messages so this buffer has a smaller, corpus-independent ceiling.
- Preserve the collecting parser wrapper for callers that genuinely need an owned result.

Related to #1209. This PR intentionally does not close that issue: the latest reporter evidence identifies the Copilot CLI OTEL parser under `~/.copilot/otel` as the primary OOM path, while this change removes a separate retained-message peak in the VS Code and submit aggregation path.

## Why

The streaming submit fold removed the final whole-corpus aggregation buffer, but the Copilot path could still retain the complete OTEL/desktop lane while building a second vector containing every VS Code request message. On request-heavy histories, each `UnifiedMessage` also owns several separately allocated strings, so keeping both collections alive extends a significant fixed per-request cost through the scan.

The submit sink also buffered 65,536 complete messages before applying the existing pricing validators. That preserved accounting behavior, but allowed the supposedly bounded buffer itself to grow to hundreds of megabytes on Copilot-heavy profiles.

This change moves the streaming boundary to the VS Code parser consumer, flushes the earlier Copilot rows once only their deduplication projections remain necessary, and retains the existing filtering, timezone, pricing, validation, sessionization, and exclusion-merging logic.

## Diff scope

- `crates/tokscale-core/src/sessions/copilot_vscode.rs`: adds a callback-based parser path that emits messages in stable file/request order, while retaining the existing collecting API as a wrapper.
- `crates/tokscale-core/src/lib.rs`: flushes the prior Copilot lane before VS Code parsing, streams accepted VS Code messages through the shared parse tail, stores session timestamps without cloning `(String, timestamp)` pairs per lookup, and lowers the `GraphSink` pricing batch to 4,096.
- Adds regression tests for emission order, bounded batch retention, and submission-exclusion aggregation across batch boundaries.
- No CLI surface, frontend behavior, database schema, cache format, workflow, dependency, or release configuration changed.

## Branch integrity

- Base: `main`
- Validated base SHA: `7ad8827a25cc8086f91a2ec4aa385f2d664085cc`
[...]

PR #1234, opened 2026-08-29
feat(submit): retain unpriced tokens with incomplete cost

### LINKED ISSUE #1044: Submission protocol cannot express unknown cost, so degraded pricing would silently lower recorded spend

IvGolovach:
## Summary

- Retain token-bearing usage that cannot be priced authoritatively instead of dropping it from submissions.
- Submit affected usage at zero cost and mark only the affected daily totals with `costIsComplete: false`.
- Preserve exact payload semantics for fully priced days and keep a total pricing outage fatal.
- Report every zero-priced provider/model group with a direct custom-pricing recovery path.
- Reconcile the implementation with the bounded streaming sink now on `main`, including merged diagnostics across batches.

Fixes #1044.

This also provides the safe client-side resolution for #1013, #1021, and #1035: unpriced rows no longer abort or disappear from a submission, while the server-side invariant from #1058 prevents an incomplete cost from lowering previously recorded spend.

## Why

The previous client behavior excluded every token-bearing row whose cost could not be established from authoritative provider-specific pricing. That avoided inventing spend, but it also made submitted token totals incomplete and caused all-unpriced histories to look like no usage. Submitting those rows at `$0.00` without an explicit completeness signal would be unsafe because the server could interpret an unknown cost as complete and overwrite a higher daily total.

PR #1058 added the server half of the protocol: a day can declare its cost incomplete, and the server preserves the appropriate cost floor. This PR completes the client half. It keeps the tokens, contributes zero unverified cost, and explicitly marks the affected day so the server can distinguish unknown cost from a real zero.

## Diff scope

- `crates/tokscale-core/src/lib.rs`: replace submission-only exclusions with zero-priced usage preparation; retain unpriced messages in aggregation; record affected dates; validate only the authoritatively priced subset; preserve the fatal guard for a complete pricing-data outage; reconcile the bounded `GraphSink` path and its cross-batch test with the new diagnostics.
- `crates/tokscale-core/src/aggregator.rs`: initialize the new submission metadata for normal graph generation.
- `crates/tokscale-core/src/pricing/lookup.rs`: align resolver documentation and tests with the incomplete-cost submission contract.
- `crates/tokscale-cli/src/main.rs`: serialize `totals.costIsComplete: false` only for affected days and replace exclusion output with explicit zero-cost warnings and recovery guidance.
- `crates/tokscale-cli/tests/cli_tests.rs`: verify that the CLI retains unpriced token totals, explains the cost-incomplete protection, and still fails when no pricing data can be loaded.
- `crates/tokscale-core/src/sessions/opencode.rs`: serialize the randomized incremental-rescan test because it mutates the same global diagnostic counter asserted exactly by existing serial tests; this is test isolation only and changes no runtime behavior.
- No database migration, dependency, CI workflow, release workflow, or parser behavior changed.

## Branch integrity

- Base: `main`
- Validated base SHA: `cf9f0b3453249649be8c0634bd74080c062ac3bc`
- Head SHA: `8287d9345b90d7c8aa26554f44cffda061bbe3bf`
- Ahead/behind: 0 behind / 2 ahead
- Merge base: `cf9f0b3453249649be8c0634bd74080c062ac3bc`
- `origin/main` is an ancestor of the branch; all final diff proof was computed against the freshly fetched base.

## Commit integrity

- `50df5deb23613cfeae39effd2dc50361c5d6d003 feat(submit): retain unpriced tokens with incomplete cost`
- `8287d9345b90d7c8aa26554f44cffda061bbe3bf test(opencode): serialize incremental rescan counter checks`
- The first commit is the submission protocol change. The second is a narrow test-only correction required after rebasing onto the current `main`; it prevents a shared test counter race exposed by the newly merged randomized test.
- Both commits are authored and committed by Ivan Golovach using the verified GitHub noreply identity.
- Ledger: not applicable — not required for these change families.
- Version: release manifests are unchanged; the repository publish workflow owns the synchronized package version bump.

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: six intended modified files listed above.
- No environment files, secrets, caches, coverage artifacts, build output, migration metadata, or unrelated generated files are included.

## Validation mode and proof

Validation mode: Mode 3 for the submission accounting and server protocol contract, plus Mode 2R for the narrow post-rebase test-isolation correction.

- `cargo test -p tokscale-core submission_merges_unpriced_usage_across_bounded_batches --lib -- --nocapture`: PASS, 1 passed.
- `cargo test -p tokscale-core test_opencode_warm_scan_reads_only_changed_rows_and_agrees_with_a_cold_parse --lib -- --nocapture`: PASS, 1 passed.
- `cargo test -p tokscale-core --lib`: PASS, 1962 passed and 2 ignored. Run outside the filesystem sandbox so loopback HTTP fixture tests could bind ephemeral ports.
- `cargo test -p tokscale-cli --bin tokscale`: PASS, 1200 passed and 1 ignored.
- `cargo test -p tokscale-cli --test cli_tests`: PASS, 175 passed.
- `bun run --cwd packages/frontend test -- __tests__/api/submit.test.ts -t costIsComplete`: PASS, 16 passed and 51 filtered.
- `rustup run stable cargo clippy --locked --workspace --all-features -- -D warnings`: PASS on Rust 1.95.0.
- `rustup run stable cargo clippy --locked --workspace --all-features --all-targets -- -D warnings`: PASS on Rust 1.95.0.
- Full frontend suite: not run — no frontend file changed; the focused server-contract suite passed, and mandatory GitHub CI is the final-SHA full proof.

## Protocol and deployment notes

- Server support was merged in #1058 on 2026-08-08.
- Live deployment of that server support has not been verified from this worktree.
- Deployment order is mandatory: confirm the live server accepts and enforces `costIsComplete` before releasing this CLI change. An older server may ignore the unknown field and lose the cost-floor protection.
- Existing released clients remain compatible because an absent flag continues to mean complete.

## Migration notes

Not applicable — this client PR changes no database migration. The server-side schema and migration work belong to #1058.

## CI context confirmation

- CI context names are unchanged.
- Required GitHub CI passed on final SHA `8287d9345b90d7c8aa26554f44cffda061bbe3bf`: strict lint, Ubuntu and Windows tests, launcher smoke, TUI container, and all eight native build targets are green.
- `Can I merge this PR?`, WIP, Vercel, and Vercel Preview Comments are green. Coverage Report is intentionally skipped by the workflow for this event, and the Cubic reviewer concluded neutral/skipped; neither is a failed context.
- GitHub reports `CLEAN` and `MERGEABLE`. No formal owner review has been submitted on the final SHA, so this is ready for owner review rather than represented as owner-approved.

## Runtime safety

- Only token-bearing rows without authoritative pricing are zeroed; already authoritative costs remain unchanged.
- `validate_priced_messages` still receives only the authoritatively priced subset, so zero-priced fallback rows are not misrepresented as verified pricing.
- A complete pricing-data outage still fails the submission rather than turning an entire local history into ungrounded zero cost.
- Complete days omit the new field and retain exact overwrite semantics; only affected days opt into the server-side floor.
- Batch aggregation keeps the existing bounded message buffer and merges only compact provider/model diagnostics and affected-date metadata.
- The OpenCode follow-up changes only test scheduling; no production annotation or code path changed.
- No new blocking lock, unbounded queue, or invariant regression is introduced.

## Documentation integrity

Not applicable — no command syntax, configuration format, data location, or release procedure changed. The CLI warning and remediation hint are updated in the same diff to describe the new behavior.

## Rollback plan

- Rollback: `git revert <post_merge_commit_sha>` after merge. Replace the placeholder with the final merge commit SHA.
- Operational caveat: reverting restores exclusion of unpriced token-bearing rows from submissions.

## Known residual risks

- Releasing against a server that has not deployed #1058 can silently discard the new semantic signal; server deployment verification is therefore a release gate, not a merge claim.
- A legitimate downward cost correction on a day marked incomplete remains deferred until a later submission has complete pricing. This is the deliberate server-side trade-off documented in #1044 and #1058.
- This change preserves tokens present in parsed source data; it does not infer token usage that a source never recorded.
- Owner review has not yet been submitted on the final SHA; no unresolved owner request or failed required context remains.

### DIFF
--- a/crates/tokscale-cli/src/main.rs
+++ b/crates/tokscale-cli/src/main.rs
@@ -4628,6 +4628,10 @@ struct TsDailyTotals {
     tokens: i64,
     cost: f64,
     messages: i32,
+    /// Absent means complete for compatibility with servers and clients that
+    /// predate #1044. Only incomplete days pay a wire-format cost.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    cost_is_complete: Option<bool>,
 }
 
 #[derive(serde::Serialize)]
@@ -4773,6 +4777,10 @@ fn to_ts_token_contribution_data(
                     tokens: d.totals.tokens,
                     cost: d.totals.cost,
                     messages: d.totals.messages,
+                    cost_is_complete: graph
+                        .incomplete_cost_dates
+                        .contains(&d.date)
+                        .then_some(false),
                 },
                 intensity: d.intensity,
                 token_breakdown: TsTokenBreakdown {
@@ -5751,28 +5759,19 @@ fn report_excluded_tokenless_rows(excluded: &[ExcludedTokenlessRow]) {
     println!();
 }
 
-fn report_unpriced_submission_exclusions(
-    excluded: &[tokscale_core::UnpricedSubmissionExclusion],
-    has_remaining_usage: bool,
-) {
+fn report_unpriced_submission_usage(unpriced: &[tokscale_core::UnpricedSubmissionUsage]) {
     use colored::Colorize;
 
-    for row in excluded {
-        let remaining_usage_message = if has_remaining_usage {
-            " Remaining priced usage will be submitted."
-        } else {
-            ""
-        };
+    for row in unpriced {
         println!(
             "{}",
             format!(
-                "  Warning: excluded {} unpriced {}/{} message(s) ({} tokens): {}.{}",
+                "  Warning: submitting {} unpriced {}/{} message(s) ({} tokens) at $0.00: {}. Affected days are marked cost-incomplete so they cannot lower previously recorded spend.",
                 row.message_count,
                 row.provider_id,
                 row.model_id,
                 format_tokens_with_commas(row.total_tokens),
                 row.reason,
-                remaining_usage_message,
             )
             .yellow()
         );
@@ -5783,12 +5782,12 @@ fn report_unpriced_submission_exclusions(
     // custom-pricing docs added after them) all had to read core sources to
     // discover that an exact-match entry in custom-pricing.json — including
     // explicit 0 rates for free models and routing labels — is the supported fix.
-    if !excluded.is_empty() {
+    if !unpriced.is_empty() {
         let pricing_path = crate::paths::get_config_dir().join("custom-pricing.json");
         println!(
             "{}",
             format!(
-                "  Hint: excluded models stay unsubmitted until priced. Add exact-match entries to\n          {}\n          keyed by the model id alone (the `model` half of the `provider/model` above),\n          where an explicit 0 declares a free model or a known routing-label rate, then\n          re-check with `tokscale submit --dry-run` and `tokscale pricing <model-id>`.",
+                "  Hint: unpriced usage is included in token totals with zero cost. Add exact-match entries to\n          {}\n          keyed by the model id alone (the `model` half of the `provider/model` above),\n          where an explicit 0 declares a free model or a known routing-label rate. Re-check\n          with `tokscale submit --dry-run` and `tokscale pricing <model-id>`, then resubmit\n          to replace the temporary cost floor with a complete total.",
                 pricing_path.display(),
             )
             .bright_black()
@@ -5956,10 +5955,7 @@ fn run_submit_command(
     // left out, so a single legacy charge can't block the whole submission.
     let excluded_rows = exclude_tokenless_cost_contributions(&mut graph_result);
     report_excluded_tokenless_rows(&excluded_rows);
-    report_unpriced_submission_exclusions(
-        &graph_result.unpriced_submission_exclusions,
-        graph_result.summary.total_tokens > 0,
-    );
+    report_unpriced_submission_usage(&graph_result.unpriced_submission_usage);
 
     println!("{}", "  Data to submit:".white());
     println!(
@@ -7091,7 +7087,8 @@ mod tests {
             years: calculate_years(&contributions),
             contributions,
             time_metrics: None,
-            unpriced_submission_exclusions: Vec::new(),
+            unpriced_submission_usage: Vec::new(),
+            incomplete_cost_dates: std::collections::BTreeSet::new(),
         }
     }
 
@@ -8565,6 +8562,31 @@ mod tests {
             payload.device.as_ref().unwrap().name.as_deref(),
             Some("Test device")
         );
+    }
+
+    #[test]
+    fn submit_payload_marks_only_incomplete_cost_days() {
+        let mut graph = graph_result_with_contributions(vec![
+            daily_contribution("2026-12-30", 10, 0.0, "opencode", "unknown"),
+            daily_contribution("2026-12-31", 20, 2.50, "codex", "model-b"),
+        ]);
+        graph.incomplete_cost_dates.insert("2026-12-30".to_string());
+
+        let payload = to_ts_token_contribution_data(&graph, None, None);
+        assert_eq!(
+            payload.contributions[0].totals.cost_is_complete,
+            Some(false)
+        );
+        assert_eq!(payload.contributions[1].totals.cost_is_complete, None);
+
+        let json = serde_json::to_value(&payload).unwrap();
+        assert_eq!(
+            json.pointer("/contributions/0/totals/costIsComplete"),
+            Some(&serde_json::Value::Bool(false))
+        );
+        assert!(json
+            .pointer("/contributions/1/totals/costIsComplete")
+            .is_none());
     }
 
     #[test]
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -5022,7 +5022,7 @@ fn test_root_with_group_by() {
 }
 
 #[test]
-fn test_submit_excludes_unpriced_usage_and_keeps_the_rest() {
+fn test_submit_includes_unpriced_usage_at_zero_and_keeps_the_rest() {
     let tmp = create_temp_fixture_dir();
     // Healthy pricing that does not happen to cover the unpriced model below.
     // Without this the fixture holds no pricing at all, which is a different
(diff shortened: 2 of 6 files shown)


junhoyeo:
This went `DIRTY` when the rest of the queue merged (#1232, #1233, #1235, #1236,
#1239, #1240, #1241 all landed within about ten minutes). I test-merged it
locally so you know the scope before you rebase: **one conflict, one hunk, in a
single `use` statement** in the `lib.rs` test module. No logic conflict anywhere.

```
crates/tokscale-core/src/lib.rs:6004
```

`main` now imports `GraphSink`, `MessageSink`, `GRAPH_SINK_BATCH` and
`UnpricedSubmissionExclusion`; this branch removes those and adds
`prepare_submission_pricing` and `UnpricedSubmissionUsage`.

It is textually trivial but I would not resolve it for you, because there is a
real question inside it: `GraphSink` / `MessageSink` / `GRAPH_SINK_BATCH` arrived
in `main` from #1233 (streaming VS Code messages into the submit fold) *after*
you branched. Whether your restructured submission path still wants them is
something only you can answer — taking "ours" or "theirs" on that hunk would
decide it silently.

Worth noting the comment sitting immediately below the conflict:

> Kept as its own statement rather than folded into the list above: that list is
> edited by nearly every PR that touches this file, and sharing it made this
> branch conflict on every single upstream merge.

Someone hit this exact problem before and split one import out to escape it. The
conflict is in the shared list they left alone.

No hurry from my side, and nothing else blocks this — CI was green before the
queue moved underneath it.

PR #1235, opened 2026-08-29
fix(dsh): key summaries by compaction id

### LINKED ISSUE #1187: DSH: an idless row's `seq` identity is unique per file, and `compaction/summary` already carries a globally unique id
#1173 gave idless DSH rows a `seq:` identity so a fork whose header lost
`seedLength` can no longer bill a copied `compaction/summary` twice. The Codex
review objected that `seq` is unique only within a file. I am not writing to
argue that probability. I went to look at what else is in the event, and there
is a per-call uuid sitting in it.

## What is in the event

Every `compaction/summary` carries `data.compactionId`. On the corpus behind
#1152 — 4 sessions, 39 usage-bearing rows — it is present on 3 of 3 summaries,
and it is copied verbatim by a fork exactly the way `message.id` is:
`session-001e8887`'s `compaction/start` at `seq 1007`, `time 1786781204037`,
`compactionId 1ad33c8f-5255-4158-b607-7555f3c26cd0` appears byte-identical in
its child `session-e61d64ec`.

So the property `msg:{id}` relies on — a per-call id the fork copies — is
available for summaries too. It was not used.

## What `seq` contributes

Not much, and this part is measured rather than reasoned. On the same corpus:

| | |
|---|---|
| usage-bearing rows | 39 |
| `assistant/message` reaching the `msg:` branch | 36 |
| `compaction/summary` reaching the `seq:` branch | 3 |
| distinct `seq` values appearing in more than one file | 17 of 22 |

`seq` is dense from zero in every file, so it collides across files as a matter
of course rather than by coincidence. Once it is in the key, what separates two
unrelated summaries is the millisecond timestamp, the routing pair and the five
buckets. The comment in the diff says as much; I am only putting a number on
how little the `seq` component itself carries.

## Two legs, one fixture

`build_fixture.py` derives the expected totals by arithmetic before writing
anything.

**Leg A** — two unrelated sessions, no `parentSession`, no `seedLength`, whose
summaries agree on `seq`, `time`, provider, model and every usage bucket and
differ only in `compactionId`. Two separately billed summarize calls.

**Leg B** — a parent and a fork whose header lost `seedLength`, the child's
prefix repeating the parent's summary verbatim. One billed call. This is the
case the `seq:` fallback exists for and it has to keep collapsing.

Measured, cold cache, isolated `HOME`, `DSH_HOME` pinned at the fixture,
binaries built from source:

| field | `6db1458e` (main) | main + `cmp:` | correct |
|---|---|---|---|
| totalInput | 870 | 1,177 | 1,177 |
| totalOutput | 101 | 132 | 132 |
| totalCacheRead | 8,700 | 11,770 | 11,770 |
| totalCacheWrite | 16 | 23 | 23 |
| totalMessages | 6 | 7 | 7 |

Current main lands on the `seq`-keyed column in every bucket: one of leg A's
two summarize calls is dropped, 3,415 tokens. Leg B collapses correctly on
both, which is the half that says the change does not cost you what #1173
bought.

And it is a no-op where it should be. Same two binaries over the #1152 corpus,
cold cache, isolated `HOME` each:

```
main   input 66,094  output 8,281  cacheRead 249,728  cacheWrite 0  messages 22
+cmp:  input 66,094  output 8,281  cacheRead 249,728  cacheWrite 0  messages 22
```

Bucket for bucket identical, and both land on the post-#1162 figures I reported
on that PR. Every summary there reaches `cmp:` instead of `seq:` and nothing
moves, because no two of them were colliding in the first place.

## The change

One `.or_else` ahead of the existing one. The `seq` fallback stays, so nothing
that depends on it regresses; it stops being the first thing an idless row
reaches.

```rust
.map(|id| format!("msg:{id}"))
.or_else(|| {
    value
        .pointer("/data/compactionId")
        .and_then(Value::as_str)
        .map(str::trim)
        .filter(|id| !id.is_empty())
        .map(|id| format!("cmp:{id}"))
})
.or_else(|| {
    value.get("seq").and_then(Value::as_i64).map(|seq| format!("seq:{seq}"))
})
```

`parser_version` would want 3 → 4 on the same reasoning as #1173, and for the
same reason as #1152 that costs nothing today: the DSH client is still not in
any release. `v4.13.0`'s `-c` lists 46 clients and `dsh` is not among them, so
there is no user-facing history to migrate, only caches written from `main`.

## What I am not claiming

Leg A is **constructed**. Nothing in the corpus shows two unrelated sessions
agreeing on a millisecond, and I have no basis for saying how often that
happens. The argument is not that the collision is likely. It is that a
globally unique id is in the same event at the same cost, so the question can
be removed rather than priced.

Leg B is constructed too, and for a reason worth saying out loud: both forks in
this corpus carry `parentSession` and `seedLength` together, so the
seedLength-less fork does not occur here at all and neither does the double
count #1173 fixes. Four sessions, one machine, two provider routes.

Harness:
[`tokscale-dsh-seq-key-check/`](https://github.com/lizhuojunx86/token-accounting-conformance/tree/main/tokscale-dsh-seq-key-check).
Fixture is committed, so it needs no corpus and no DeepSeek account;
`key_entropy.py` and `headers.py` are the read-only probes behind the two
tables and write nothing.

Happy to run it against a branch, same as on #1162.

IvGolovach:
## Summary

- Key DSH `compaction/summary` usage by its globally stable `data.compactionId` before considering transcript-local sequence numbers.
- Keep the existing `seq` identity as a compatibility fallback for legacy or damaged summaries that do not carry a compaction ID.
- Bump the DSH parser cache identity from 4 to 5 so released seq-keyed rows are reparsed instead of remaining silently undercounted.
- Add parser, cache-migration, and cold/warm CLI regressions that prove distinct summarize calls survive while a copied fork summary still collapses.

Fixes #1187.

## Why

Sequence numbers restart in every DSH transcript. Before this change, an idless compaction summary used `seq` as its primary cross-file identity, so two unrelated summarize calls could collapse when their local sequence, millisecond timestamp, provider, model, and token buckets matched. That loses real usage.

The same event already carries a per-call `data.compactionId`. DSH copies that UUID verbatim when a fork replays the parent summary, which gives the deduplicator exactly the property it needs: unrelated calls remain distinct, while the same billed call copied into another transcript retains one identity. The previous `seq` fallback remains necessary for older summaries without a compaction ID and continues to protect the seedLength-less fork case introduced by #1173.

## Diff scope

- `crates/tokscale-core/src/sessions/dsh.rs`: prefer a trimmed, non-empty `data.compactionId` for compaction summaries after `message.id` and before `seq`; retain the legacy sequence fallback; add direct regressions for distinct summaries, copied summaries, and missing IDs.
- `crates/tokscale-core/src/message_cache.rs`: bump the DSH parser version from 4 to 5 and verify a released v4 seq-keyed shard is rejected, reparsed with `cmp:` identity, and persisted under v5.
- `crates/tokscale-cli/tests/cli_tests.rs`: exercise the complete collision boundary through the CLI on both cold and warm cache passes.
- No database schema, dependency, CLI surface, frontend behavior, workflow, release configuration, or unrelated parser changed.

## Branch integrity

- Base: `main`
- Head SHA: `1c39c7800796bf3b1d11438a9b649bc36521087f`
- Ahead/behind: 0 behind / 1 ahead
- `origin/main` is an ancestor of the branch; all diff proof was computed against the freshly fetched base.

## Commit integrity

- One atomic commit: `1c39c7800796bf3b1d11438a9b649bc36521087f fix(dsh): key summaries by compaction id`
- The final diff contains only the DSH identity change, its source-cache migration, and directly related tests.
- Version: release manifests are unchanged; the repository publish workflow owns synchronized package version bumps.

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: three intended modified files.

## Validation mode and proof

Validation mode: Mode 3 — cross-session usage identity and persisted source-cache interpretation directly affect accounting totals.

- `cargo test -p tokscale-core dsh --lib`: PASS, 35 passed.
- `cargo test -p tokscale-cli --test cli_tests dsh`: PASS, 5 passed after the new fixture.
- `cargo test -p tokscale-core --lib`: PASS on the final repeat, 1941 passed and 2 ignored.
- `cargo test -p tokscale-cli --test cli_tests`: PASS, 174 passed.
- The first full core run had one unrelated Antigravity timestamp test failure; that test passed in isolation, and the complete core suite then passed on repeat. No Antigravity file is changed by this branch.
- Clippy is not reported as passing: the workspace baseline reaches a pre-existing `sessions/pi.rs` `needless_lifetimes` finding outside this diff.

## Cache and migration notes

- This is an application cache migration, not a database migration.
- DSH parser identity changes from v4 to v5 because completed transcripts can retain the same source fingerprint while their deduplication meaning changes.
- Existing v4 DSH cache rows are invalidated and rebuilt once from the original transcript; they are not translated in place.
- The migration is now user-facing because DSH shipped before this fix. The one-time reparse is required to recover summaries that v4 could have collapsed.
- Database downgrade and data repair are not applicable; source transcripts remain authoritative.

## CI context confirmation

- CI workflow files and context names are unchanged.
- Pending — required GitHub CI starts after this PR is created and must pass on the final SHA before merge.
- The PR is ready for review, not merge-ready, until required remote checks and review gates pass.

## Runtime safety

- `message.id` remains the highest-priority identity when present.
- `compactionId` is used only for `compaction/summary` rows and only when it is a non-empty string after trimming.
- Summaries without a usable compaction ID retain the existing `seq` fallback, preserving compatibility with old or damaged transcripts.
- A forked copy with the same compaction UUID still collapses; unrelated summaries with different UUIDs cannot collide merely because their transcript-local coordinates match.
- No new lock, queue, background worker, network access, or unbounded collection is introduced.

## Documentation integrity

Not applicable — no command, configuration, data location, public API, or operational runbook changed.

## Rollback plan

- Rollback: `git revert <post_merge_commit_sha>` after merge.
- Operational caveat: reverting restores seq-first summary identity and causes the active parser cache identity to rebuild again.

## Known residual risks

- Legacy summaries without `compactionId` still depend on the existing composite `seq` fallback and retain its theoretical cross-file collision boundary.
- The first DSH read after upgrading performs a one-time cache reparse for affected shards.
- Required remote CI has not run because the PR has not yet been created.

---
## Summary by cubic
Fixes #1187 so DSH compaction summaries are keyed by their globally stable `data.compactionId` before transcript-local `seq`, which stops unrelated summarize calls from collapsing while copied fork summaries still deduplicate. The DSH parser cache identity bumps 4→5, so released v4 seq-keyed rows are invalidated and reparsed once from source transcripts.

**Notes**
- Summaries without a usable `compactionId` keep the `seq` fallback for legacy or damaged transcripts.
- `message.id` remains the highest-priority identity when present.

<sup>Written for commit 1c39c7800796bf3b1d11438a9b649bc36521087f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -1505,6 +1505,48 @@ fn write_dsh_seeded_pair_without_seed_length(base: &Path) {
     }
 }
 
+/// Two unrelated summaries with identical local coordinates but distinct
+/// compaction UUIDs, plus one parent/child replay sharing a UUID. This is the
+/// complete #1187 collision boundary: retain two calls from the first leg and
+/// collapse the copied call from the second.
+fn write_dsh_compaction_identity_fixture(base: &Path) {
+    let fixtures = [
+        (
+            "unrelated-a",
+            r#"{"type":"session","id":"unrelated-a","createdAt":1,"cwd":"/tmp/dsh-workspace"}"#,
+            r#"{"type":"compaction/summary","seq":4,"time":1786669450002,"data":{"compactionId":"compact-a","message":{"source":{"provider":"p","model":"m"}},"usage":{"inputTokens":10,"outputTokens":20}}}"#,
+        ),
+        (
+            "unrelated-b",
+            r#"{"type":"session","id":"unrelated-b","createdAt":2,"cwd":"/tmp/dsh-workspace"}"#,
+            r#"{"type":"compaction/summary","seq":4,"time":1786669450002,"data":{"compactionId":"compact-b","message":{"source":{"provider":"p","model":"m"}},"usage":{"inputTokens":10,"outputTokens":20}}}"#,
+        ),
+        (
+            "fork-parent",
+            r#"{"type":"session","id":"fork-parent","createdAt":3,"cwd":"/tmp/dsh-workspace"}"#,
+            r#"{"type":"compaction/summary","seq":9,"time":1786669450003,"data":{"compactionId":"compact-copied","message":{"source":{"provider":"p","model":"m"}},"usage":{"inputTokens":30,"outputTokens":40}}}"#,
+        ),
+        (
+            "fork-child",
+            r#"{"type":"session","id":"fork-child","createdAt":4,"cwd":"/tmp/dsh-workspace","parentSession":"fork-parent"}"#,
+            r#"{"type":"compaction/summary","seq":9,"time":1786669450003,"data":{"compactionId":"compact-copied","message":{"source":{"provider":"p","model":"m"}},"usage":{"inputTokens":30,"outputTokens":40}}}"#,
+        ),
+    ];
+
+    for (session_id, header, summary) in fixtures {
+        let dir = dsh_sessions_root(base)
+            .join("-tmp-dsh-workspace")
+            .join(session_id);
+        fs::create_dir_all(&dir).unwrap();
+        let payload = format!("{header}\n{summary}\n");
+        fs::write(
+            dir.join("session.jsonl.zstd"),
+            zstd::encode_all(payload.as_bytes(), 3).unwrap(),
+        )
+        .unwrap();
+    }
+}
+
 fn write_jcode_session(base: &Path) {
     let sessions_dir = base.join(".jcode/sessions");
     fs::create_dir_all(&sessions_dir).unwrap();
@@ -2476,6 +2518,28 @@ fn test_dsh_repeated_message_ids_across_transcripts_count_once_cold_and_warm_cac
         assert_eq!(json["totalMessages"].as_i64(), Some(1), "{pass} cache pass");
         assert_eq!(json["totalInput"].as_i64(), Some(2885), "{pass} cache pass");
         assert_eq!(json["totalOutput"].as_i64(), Some(2), "{pass} cache pass");
+    }
+}
+
+#[test]
+fn test_dsh_compaction_identity_is_global_cold_and_warm_cache() {
+    let tmp = create_empty_fixture_dir();
+    write_dsh_compaction_identity_fixture(tmp.path());
+
+    for pass in ["cold", "warm"] {
+        let output = cmd_with_home(tmp.path())
+            .args(["models", "--json", "--client", "dsh", "--no-spinner"])
+            .output()
+            .unwrap();
+        assert!(
+            output.status.success(),
+            "{pass} cache pass failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+        let json: serde_json::Value = serde_json::from_slice(&output.stdout).unwrap();
+        assert_eq!(json["totalMessages"].as_i64(), Some(3), "{pass} cache pass");
+        assert_eq!(json["totalInput"].as_i64(), Some(50), "{pass} cache pass");
+        assert_eq!(json["totalOutput"].as_i64(), Some(80), "{pass} cache pass");
     }
 }
 
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1254,7 +1254,11 @@ fn parser_version(client: ClientId) -> u32 {
         // request model. Finished DSH transcripts are append-only and keep the
         // same fingerprint, so only this bump replaces cached alias/substitute
         // attribution and its derived pricing.
-        ClientId::Dsh => 4,
+        // v4->v5: compaction summaries now use their globally stable
+        // `data.compactionId` before the per-transcript `seq` fallback. Reparse
+        // released v4 rows so unrelated summaries with otherwise identical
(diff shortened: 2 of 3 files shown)


junhoyeo (approved):
Reviewed via subagent: builds correctly on top of #1221 served-model attribution, no revert. Compaction-id keying occupies the identity slot (distinct from #1221's model_id slot); no summary collision/drop/double-count; parser_version chained v4->v5. No conflict with concurrent opencode PRs (disjoint message_cache.rs locations).

PR #1236, opened 2026-08-29
fix(opencode): namespace idless legacy message keys

### LINKED ISSUE #1198: OpenCode: id-less legacy messages dedupe by filename stem, so same-named files in different directories collide
A legacy OpenCode JSON message with no embedded `id` falls back to its filename stem as its deduplication identity:

```rust
// crates/tokscale-core/src/sessions/opencode.rs:70-75
let dedup_key = msg.id.or_else(|| {
    path.file_stem()
        .and_then(|s| s.to_str())
        .map(|s| s.to_string())
});
```

The scan walks legacy message files recursively, and a single OpenCode-wide `HashSet` suppresses repeated keys across the SQLite source and every JSON path. A filename stem is only unique within its own directory, so two distinct id-less messages that live in different project or session directories and happen to share a basename collide: the first wins and the second is silently dropped, undercounting usage with no warning.

The embedded-id path is unaffected and should stay as it is — matching ids across SQLite and JSON is the intended overlap suppression.

Suggested fix: namespace the no-id fallback by something stable and unique to the file's location (session id plus path, or the path relative to the scan root) rather than the bare stem, and add an end-to-end test with two id-less files sharing a basename in separate directories asserting both survive, alongside one confirming that matching embedded ids still deduplicate across the two sources.

Not a recent regression: `git blame` puts this fallback at ce53bbcd (2026-02-16), so it predates v4.13.0 and has been shipping for months. Filing rather than fixing inline because it wants a real fixture and a deliberate choice of identity, not a quick patch.

IvGolovach:
## Summary

- Give id-less legacy OpenCode JSON messages a canonical path-scoped identity instead of a globally colliding filename stem.
- Keep real embedded message IDs unnamespaced so intended deduplication across SQLite, legacy JSON, forks, and channel databases remains unchanged.
- Tighten the pre-open SQLite migration fast path to require both session ID and message ID before suppressing a legacy JSON file.
- Bump the OpenCode parser cache identity from 2 to 3 so already cached stem-keyed messages are rebuilt with the corrected identity.
- Add parser, cache, full aggregation, local-client, and cold/warm CLI regressions for same-named files in different sessions.

Fixes #1198.

## Why

The legacy OpenCode scanner walks `storage/message/<session>/<file>.json` recursively. When a JSON message had no embedded `id`, the parser used only the filename stem as its deduplication key. That stem is local to a directory, not globally unique, so two real assistant messages in different sessions with the same filename silently collapsed and undercounted usage.

There was a second path with the same assumption: the SQLite-first migration optimization skipped a JSON file before opening it whenever its stem appeared anywhere in the SQLite key set. An unrelated id-less file in another session could therefore be discarded even before receiving a corrected fallback identity.

This change separates the two identity classes deliberately. A real embedded message ID remains global because it identifies the same migrated message across sources. An id-less legacy row uses the canonical physical file location because that is the only stable identity the source provides.

## Diff scope

- `crates/tokscale-core/src/sessions/opencode.rs`: replace the bare-stem fallback with `legacy-json-path:<canonical path>`; preserve embedded IDs; fail open with no dedup key if a path cannot be represented losslessly; add direct same-name regressions.
- `crates/tokscale-core/src/lib.rs`: index SQLite identities by session and message ID for the pre-open migration check; preserve the existing global embedded-ID dedup set; cover both streaming aggregation and `parse_local_clients` on cold and warm caches.
- `crates/tokscale-core/src/message_cache.rs`: bump OpenCode parser identity from 2 to 3 and prove a v2 stem-keyed shard is rejected, reparsed, and persisted with path identity.
- `crates/tokscale-cli/tests/cli_tests.rs`: verify two same-named id-less files in separate sessions both contribute to CLI totals on cold and warm passes.
- No database schema, dependency, CLI surface, frontend behavior, workflow, release configuration, or unrelated client parser changed.

## Branch integrity

- Base: `main`
- Head SHA: `32063a8d77c69d46de791b9eaa5c6c8d0fc0c17a`
- Ahead/behind: 0 behind / 1 ahead
- `origin/main` is an ancestor of the branch; all diff proof was computed against the freshly fetched base.

## Commit integrity

- One atomic commit: `32063a8d77c69d46de791b9eaa5c6c8d0fc0c17a fix(opencode): namespace idless legacy message keys`
- The final diff contains only OpenCode identity handling, its source-cache migration, and directly related tests.
- Version: release manifests are unchanged; the repository publish workflow owns synchronized package version bumps.

## Diff hygiene

- `git diff --name-status origin/main...HEAD`: four intended modified files.

## Validation mode and proof

Validation mode: Mode 3 — deduplication identity and persisted source-cache interpretation directly affect usage-accounting totals.

- `cargo test -p tokscale-core opencode --lib`: PASS, 127 passed and 1 ignored.
- `cargo test -p tokscale-core --lib`: PASS, 1942 passed and 2 ignored.
- `cargo test -p tokscale-cli --test cli_tests opencode_same_named_idless`: PASS, 1 passed.
- `cargo test -p tokscale-cli --test cli_tests`: PASS, 174 passed.
- Clippy is not reported as passing: the workspace baseline reaches a pre-existing `sessions/pi.rs` `needless_lifetimes` finding outside this diff.

## Cache and migration notes

- This is an application source-cache migration, not a database migration.
- OpenCode parser identity changes from v2 to v3 because an unchanged JSON file can retain its fingerprint while its fallback deduplication identity changes.
- Existing v2 OpenCode shards are invalidated and rebuilt once from the original SQLite/JSON sources; they are not translated in place.
- Embedded IDs remain byte-for-byte compatible, while only id-less legacy rows receive the new path namespace.
- Database downgrade and data repair are not applicable; source data remains authoritative.

## CI context confirmation

- CI workflow files and context names are unchanged.
- Pending — required GitHub CI starts after this PR is created and must pass on the final SHA before merge.
- The PR is ready for review, not merge-ready, until required remote checks and review gates pass.

## Runtime safety

- Canonicalization makes two path spellings of the same physical legacy file converge on one identity.
- If canonicalization fails, the original path is used; if that path is not valid UTF-8, the parser leaves the dedup key absent rather than inventing a lossy key that could undercount.
- The SQLite fast path now suppresses a JSON file only when its parent session and filename stem match a known SQLite location.
- Global deduplication by real embedded message ID remains unchanged.
- The session-location index is proportional to the already parsed OpenCode SQLite identities and is released after the OpenCode lane; it adds memory overhead but does not add a queue, background worker, lock, or extra database scan.

## Documentation integrity

Not applicable — no command, configuration, data location, public API, or operational runbook changed.

## Rollback plan

- Rollback: `git revert <post_merge_commit_sha>` after merge.
- Operational caveat: reverting restores globally colliding stem identities and causes the active parser cache identity to rebuild again.

## Known residual risks

- Moving an id-less legacy file changes its fallback identity because the source provides no embedded stable ID; this is preferable to collapsing unrelated messages that share a basename.
- The first OpenCode read after upgrading performs a one-time cache reparse for affected shards.
- The additional session-location index temporarily duplicates SQLite identity strings during the OpenCode lane.
- Required remote CI has not run because the PR has not yet been created.

---
## Summary by cubic
Fixes OpenCode usage undercounting where two id-less legacy JSON messages with the same filename in different sessions collapsed into one message because the filename stem alone was used as the dedup key.

**Bug Fixes**
- Id-less messages now use a canonical path-scoped dedup key (`legacy-json-path:<path>`) instead of the bare filename stem; embedded message IDs stay unnamespaced so cross-source dedup is unchanged.
- The SQLite migration fast path now requires both session ID and filename stem to match a known SQLite row before suppressing a legacy JSON file.
- The OpenCode parser cache version bumps from 2 to 3, invalidating cached stem-keyed entries for a one-time rebuild.
- Adds regressions for same-named id-less files across sessions covering streaming aggregation, `parse_local_clients`, and cold/warm CLI passes.

Fixes #1198.

<sup>Written for commit 32063a8d77c69d46de791b9eaa5c6c8d0fc0c17a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1236?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -1166,6 +1166,21 @@ fn add_reasoning_only_opencode_message(base: &Path) {
     fs::write(session.join("reasoning-msg.json"), message).unwrap();
 }
 
+fn add_same_named_idless_opencode_messages(base: &Path) {
+    let root = base.join(".local/share/opencode/storage/message");
+    for (session_id, input, output) in [("idless-session-a", 13, 2), ("idless-session-b", 17, 3)] {
+        let session = root.join(session_id);
+        fs::create_dir_all(&session).unwrap();
+        fs::write(
+            session.join("same-name.json"),
+            format!(
+                r#"{{"sessionID":"{session_id}","role":"assistant","modelID":"gpt-4o","providerID":"openai","tokens":{{"input":{input},"output":{output},"reasoning":0,"cache":{{"read":0,"write":0}}}},"time":{{"created":1700000000000}}}}"#
+            ),
+        )
+        .unwrap();
+    }
+}
+
 fn write_fireworks_pricing_cache(base: &Path) {
     let now = SystemTime::now()
         .duration_since(UNIX_EPOCH)
@@ -2213,6 +2228,28 @@ fn test_models_with_client_filter_opencode() {
     let entries = json["entries"].as_array().unwrap();
     for entry in entries {
         assert_eq!(entry["client"].as_str().unwrap(), "opencode");
+    }
+}
+
+#[test]
+fn test_opencode_same_named_idless_files_survive_cold_and_warm_cache() {
+    let tmp = create_empty_fixture_dir();
+    add_same_named_idless_opencode_messages(tmp.path());
+
+    for pass in ["cold", "warm"] {
+        let output = cmd_with_home(tmp.path())
+            .args(["models", "--json", "--client", "opencode", "--no-spinner"])
+            .output()
+            .unwrap();
+        assert!(
+            output.status.success(),
+            "{pass} cache pass failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+        let json: serde_json::Value = serde_json::from_slice(&output.stdout).unwrap();
+        assert_eq!(json["totalMessages"].as_i64(), Some(2), "{pass} cache pass");
+        assert_eq!(json["totalInput"].as_i64(), Some(30), "{pass} cache pass");
+        assert_eq!(json["totalOutput"].as_i64(), Some(5), "{pass} cache pass");
     }
 }
 
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -770,14 +770,28 @@ fn parse_all_messages_with_pricing_with_env_strategy(
 /// lanes are scanned and the SQLite lane runs first, so every migrated file is
 /// parsed only for its message to be dropped by the dedup filter that follows.
 ///
-/// The file stem is the message id, which is exactly the dedup key the SQLite
-/// lane inserted, so the duplicate can be recognized from the path alone —
-/// before the file is opened. On a migrated install that skips nearly the whole
-/// directory (#1209); unmigrated files miss the set and are still parsed.
-fn opencode_json_superseded_by_sqlite(path: &Path, sqlite_keys: &HashSet<String>) -> bool {
-    path.file_stem()
-        .and_then(|stem| stem.to_str())
-        .is_some_and(|stem| sqlite_keys.contains(stem))
+/// The file stem is the message id and its parent directory is the session id.
+/// Requiring both preserves the pre-open migration fast path from #1209 while
+/// preventing an id-less file in another session from being suppressed merely
+/// because its local filename matches an unrelated SQLite id (#1198).
+fn opencode_json_superseded_by_sqlite(
+    path: &Path,
+    sqlite_locations: &HashMap<String, HashSet<String>>,
+) -> bool {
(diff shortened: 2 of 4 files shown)


junhoyeo (approved):
Reviewed via subagent: fixes #1198 legacy id-less key collision (now namespaced by canonical path), tightens JSON-vs-SQLite dedup to require both session+message id. Removes a latent undercount, introduces none; Claude path untouched; operates downstream of #1232 with no conflict (verified clean merge against updated main).

Issue #1237, opened 2026-08-29
Pin the Rust toolchain: local clippy is older than CI's, so a green local gate still fails Lint


junhoyeo:
## What

There is no `rust-toolchain.toml`, so a local checkout runs whatever `rustup` last installed while CI uses `dtolnay/rust-toolchain@stable`. The two drift, and the drift is invisible until CI fails.

Concretely, on 2026-08-29:

```
local   clippy 0.1.93 / rustc 1.93.1   (2026-02-11)
CI      stable, resolved to 1.98.0     (2026-08-18)
```

## Why it bites

CI's `Lint` job runs exactly the command a contributor would run locally:

```
cargo clippy --locked --workspace --all-features -- -D warnings
```

Same command, same flags — and it still fails where local passes, because newer clippy ships new lints. On #1232 this was `clippy::question_mark` firing on two `match x { Some(v) => …, None => return None }` arms that 1.93 accepts without comment.

The failure mode is bad in a specific way: the contributor has no signal that their gate is weaker than CI's. It does not look like a version problem, it looks like a flaky or mysterious lint. I ran clippy in both forms (bare and `--all-targets`) all through a review session believing that was sufficient, and it was not.

## Proposal

Add a `rust-toolchain.toml` at the repo root:

```toml
[toolchain]
channel = "1.98.0"
components = ["clippy", "rustfmt"]
```

Then `cargo` picks it up automatically for everyone, and CI's `@stable` should become the same pinned channel so the two cannot diverge again.

Trade-off worth naming: a pin means new stable lints stop arriving on their own, so the pin needs bumping deliberately. That is the point — a bump becomes a reviewable change that fixes its own fallout, instead of arriving unannounced in someone else's PR.

## Alternative considered

Leave it unpinned and tell contributors to `rustup update stable` first. That relies on everyone knowing to, and gives no signal when they have not — which is the situation this issue is about.

PR #1241, opened 2026-08-30
fix(zai): accept CREDIT_LIMIT for session/weekly windows on V3 GLM Coding plans

### LINKED ISSUE #1231: Z.ai: V3 GLM Coding plans return CREDIT_LIMIT, so usage reports no metrics
### What happens

`tokscale usage` lists Z.ai with the correct plan but **no metrics at all**:

```json
{
  "provider": "Z.ai",
  "plan": "Pro",
  "email": null,
  "metrics": []
}
```

The API key is valid and both requests succeed — the quota endpoint returns
HTTP 200 with real usage. Claude, Codex and Grok Build in the same run report
their metrics normally.

### Root cause

Current Z.ai GLM Coding plans (the subscription entry reports `"version": "V3"`)
return quota entries typed **`CREDIT_LIMIT`** rather than `TOKENS_LIMIT`.

`crates/tokscale-cli/src/commands/usage/zai.rs` matches only two type strings:

```rust
match limit.limit_type.as_deref() {
    Some("TOKENS_LIMIT") => { ... }   // zai.rs:115 — session + weekly
    Some("TIME_LIMIT")   => { ... }   // zai.rs:139 — web search
    _ => {}                            // CREDIT_LIMIT entries land here
}
```

So every limit falls through the catch-all, all three `Option<UsageMetric>`
slots stay `None`, and `metrics` is built empty.

The plan still resolves because it falls back to `level` from the quota
response at `zai.rs:97` (`"pro"` → `capitalize` → `"Pro"`), which never needed
the limits.

### Actual payload from `/api/monitor/usage/quota/limit`

```json
// illustrative values; structure verbatim from a live V3 GLM Coding account
{
  "code": 200,
  "msg": "Operation successful",
  "data": {
    "limits": [
      { "type": "CREDIT_LIMIT", "unit": 3, "number": 5, "usage": 2000,  "currentValue": 40, "remaining": 1960, "percentage": 2, "nextResetTime": 1788029492559 },
      { "type": "CREDIT_LIMIT", "unit": 6, "number": 1, "usage": 10000, "currentValue": 40, "remaining": 9960, "percentage": 1, "nextResetTime": 1788616038998 }
    ],
    "level": "pro"
  },
  "success": true
}
```

Note the `(unit, number)` pairs are **unchanged**: `(3, 5)` is the 5-hour session
window and `(6, 1)` the weekly one, exactly what `zai.rs:124` and `zai.rs:130`
already match on. `percentage` is still supplied directly too. So only the type
string differs — the whole window classification and percentage mapping below it
already works.

`TIME_LIMIT` is simply absent for this account, so omitting the Web Search
metric is correct here rather than a second bug.

### Suggested fix

Accept either type for the session/weekly arm, which keeps older plans working:

```rust
Some("TOKENS_LIMIT") | Some("CREDIT_LIMIT") => { ... }
```

The same change was accepted upstream in OpenUsage
([robinebers/openusage#1104](https://github.com/robinebers/openusage/issues/1104),
closed), which likewise recommends treating `CREDIT_LIMIT` exactly like
`TOKENS_LIMIT` and leaving the window classification alone.

### Environment

- tokscale 4.15.0 (`@tokscale/cli-linux-x64-gnu`), run via `npx tokscale@latest`
- Ubuntu, x86_64
- Key supplied via `ZAI_API_KEY`
- Subscription endpoint reports `productName: "GLM Coding Pro"`, `version: "V3"`

I have not opened a PR because I have no Rust toolchain on this machine to
compile or test the change, and did not want to submit unverified Rust to a repo
this active. Happy to follow up with one if useful.

taimurrabuske:
Fixes #1231.

## What

V3 GLM Coding plans (subscription `version: "V3"`) return quota entries typed `CREDIT_LIMIT` rather than `TOKENS_LIMIT`. The `(unit, number)` window pairs and `percentage` fields are unchanged, so every limit fell into the catch-all arm and Z.ai usage output showed the correct plan with empty `metrics`.

This accepts either type in the session/weekly arm:

```rust
Some("TOKENS_LIMIT") | Some("CREDIT_LIMIT") => { ... }
```

with a deterministic preference for `CREDIT_LIMIT` (`target.is_none() || prefer`) so a plan that ever emits both types for one window cannot last-write-wins. `TIME_LIMIT` handling is untouched.

## Verification

Built with stable 1.98.0 (`cargo build`, `cargo clippy`, `cargo fmt` clean) and run against a live V3 account (`level: max`, two `CREDIT_LIMIT` entries, `percentage: 1` for both windows):

**Before** — `tokscale usage --json`:

```json
{
  "provider": "Z.ai",
  "plan": "Max",
  "email": null,
  "metrics": []
}
```

**After:**

```json
{
  "provider": "Z.ai",
  "plan": "Max",
  "email": null,
  "metrics": [
    { "label": "Session", "used_percent": 1.0, "remaining_percent": 99.0, "remaining_label": null, "resets_at": null },
    { "label": "Weekly", "used_percent": 1.0, "remaining_percent": 99.0, "remaining_label": null, "resets_at": null }
  ]
}
```

Percentages match the API's `percentage` fields exactly; older `TOKENS_LIMIT` plans are unaffected since the arm is additive.

---
## Summary by cubic
Fixes usage metrics showing empty for Z.ai V3 GLM Coding plans by accepting `CREDIT_LIMIT` quota entries alongside `TOKENS_LIMIT`. Previously these plans matched no window and output the correct plan with empty `metrics`; now Session and Weekly percentages populate. Prefers `CREDIT_LIMIT` when both types appear for the same window so the last entry cannot silently win. `TIME_LIMIT` handling is unchanged, and existing `TOKENS_LIMIT` plans are unaffected since the arm is additive. Fixes #1231.

<sup>Written for commit db24ef4eaebea0c90305ed3039fcffd369bf1557. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1241?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/usage/zai.rs
+++ b/crates/tokscale-cli/src/commands/usage/zai.rs
@@ -112,28 +112,25 @@ pub fn fetch() -> Result<UsageOutput> {
                 };
 
                 match limit.limit_type.as_deref() {
-                    Some("TOKENS_LIMIT") => {
-                        let metric = UsageMetric {
-                            label: String::new(),
-                            used_percent: pct,
-                            remaining_percent: 100.0 - pct,
-                            remaining_label: None,
-                            resets_at: None,
+                    // V3 GLM Coding plans report the same (unit, number)
+                    // windows as CREDIT_LIMIT instead of TOKENS_LIMIT. Prefer
+                    // CREDIT_LIMIT so a plan that ever emits both for one
+                    // window cannot silently last-write-wins.
+                    Some("TOKENS_LIMIT") | Some("CREDIT_LIMIT") => {
+                        let prefer = limit.limit_type.as_deref() == Some("CREDIT_LIMIT");
+                        let (target, label) = match (limit.unit, limit.number) {
+                            (Some(3), Some(5)) => (&mut session_metric, "Session"),
+                            (Some(6), Some(1)) => (&mut weekly_metric, "Weekly"),
+                            _ => continue,
                         };
-                        match (limit.unit, limit.number) {
-                            (Some(3), Some(5)) => {
-                                session_metric = Some(UsageMetric {
-                                    label: "Session".into(),
-                                    ..metric
-                                });
-                            }
-                            (Some(6), Some(1)) => {
-                                weekly_metric = Some(UsageMetric {
-                                    label: "Weekly".into(),
-                                    ..metric
-                                });
-                            }
-                            _ => {}
+                        if target.is_none() || prefer {
+                            *target = Some(UsageMetric {
+                                label: label.to_string(),
+                                used_percent: pct,
+                                remaining_percent: 100.0 - pct,
+                                remaining_label: None,
+                                resets_at: None,
+                            });
                         }
                     }
                     Some("TIME_LIMIT") => {


junhoyeo:
Thank you for sending this, and for answering the question rather than routing
around it.

I checked the two things I flagged on #1231:

**The both-types case is genuinely order-independent, not just usually right.**
`if target.is_none() || prefer` gives `CREDIT_LIMIT` the window regardless of
which order the array arrives in — `[TOKENS, CREDIT]` overwrites on `prefer`,
`[CREDIT, TOKENS]` refuses on `!is_none() && !prefer`. That is stronger than
what I asked for; I would have accepted "my account only emits one type".

**`_ => continue` is equivalent to the `_ => {}` it replaces.** I checked
rather than assumed, because a `continue` inside a rewritten loop is exactly
where a silent behaviour change hides: the outer `match` is the last statement
in the `for` body, nothing follows it, so falling out of the old empty arm and
skipping to the next iteration reach the same place. It also matches the
`percentage` guard immediately above, which already `continue`s.

Restructuring to `(&mut target, label)` rather than building a template
`UsageMetric` and spreading it is a real simplification — the old shape
constructed a metric with an empty label on every iteration including the ones
that matched no window.

One observation, not a blocker and not a regression: two `CREDIT_LIMIT` entries
for the *same* window would still last-write-wins between themselves, since
`prefer` is true for both. The API returns one per window and the old code had
the same property, so nothing is lost — worth knowing only if Z.ai ever splits a
window across entries.

Your before/after against a live `max` account is what I could not produce; I
have no V3 account to test against, which is why I offered to write this only if
you would rather not. Building on stable 1.98.0 is also the right call — this
repo pins no toolchain and CI resolves to 1.98, which has bitten a PR here this
week (#1237).

CI is green apart from cubic, which reports NEUTRAL without reviewing. Looks
right to me; leaving the merge to @junhoyeo, who is going through the PR queue.

junhoyeo (approved):
Reviewed via subagent: correct and backward-compatible (TOKENS_LIMIT | CREDIT_LIMIT share the same (unit,number) window disambiguation); deletions are a build-once refactor, no case dropped; no mis-attribution risk (percentage sourced only from limit.percentage, prefer-guard makes CREDIT_LIMIT deterministic); display-only vendor-reported data; no revert per blame.

PR #1239, opened 2026-08-29
feat(clients): add Unsloth Studio usage parsing

The-Minion-oOo:
## Summary

- Add Unsloth Studio as a parse-local client across the Rust core, CLI/TUI, and frontend registry, displayed as `Unsloth` in the app.
- Read exact internal-chat usage from completed assistant metadata and authenticated external API usage from the content-free `api_usage_events` table in `$UNSLOTH_STUDIO_HOME/studio.db` (fallback: `~/.unsloth/studio/studio.db`).
- Preserve input, output, cache-read, cache-write, and reasoning token categories; deduplicate stable message and request IDs; and invalidate cached scans when SQLite WAL state changes.
- Avoid selecting prompts, responses, credentials, previews, or other message content. Training metrics are excluded, and local inference is reported at $0.
- Support older database schemas without the API usage table and document the integration in the maintained English, Japanese, Korean, and Simplified Chinese READMEs.

## Testing

- `cargo fmt --all -- --check`
- `cargo clippy --locked -p tokscale-core --all-features --all-targets -- -D warnings`
- `cargo test -p tokscale-core unsloth --lib` (8 passed)
- Core regression suite (1,946 passed, 2 ignored; one Windows symlink-permission test excluded)
- CLI unit suite (1,182 passed, 1 ignored)
- Applicable CLI integration suite (159 passed; six host-sensitive Windows timezone/Hermes cases excluded after reproducing them outside this change)
- `bun run --cwd packages/frontend typecheck`
- Frontend test suite (951 passed, 6 skipped)
- Real-database smoke test: 130 input, 6,545 cache-read, and 2,887 output tokens across 6 messages (9,562 total), exactly matching the source counters

---
## Summary by cubic
Adds Unsloth Studio as a new trackable client, reading exact inference usage from its local SQLite database without touching message content.

**New Features**
- Adds `unsloth` to the CLI client list, TUI hotkey `U`, and frontend source registry.
- Reads internal chat usage from assistant-message metadata and authenticated API usage from the `api_usage_events` table.
- Falls back to `~/.unsloth/studio/studio.db` when `$UNSLOTH_STUDIO_HOME` is unset.
- Deduplicates stable message/request IDs and refreshes cached scans when SQLite WAL state changes.
- Skips training metrics and reports local inference at $0.
- Supports older database schemas without the API usage table and updates all four READMEs.

<sup>Written for commit 5870b0ae7c852a38e69f6b0820a1543862ceb6b4. Summary will update on new commits.</sup>

[...]

PR #1242, opened 2026-08-30
chore: address review follow-ups for zai, unsloth, and pricing errors

junhoyeo:
## Summary

Follow-up nits gathered while reviewing the recently merged #1239 (Unsloth Studio), #1240 (network error surfacing), and #1241 (Z.ai CREDIT_LIMIT). No behavior change — this is test coverage, one behavior-preserving refactor, and documentation.

## Changes

**`zai` — testable window mapping + coverage (nit from #1241)**

The session/weekly window logic added in #1241 lived inline inside `fetch()`, unreachable without live network + env, so it had no unit tests. This extracts a pure `build_metrics(...)` function (a faithful move of the existing loop — `search_reset` is now computed once and threaded in as a parameter) and adds tests for:
- `CREDIT_LIMIT` `(3, 5)` → Session and `(6, 1)` → Weekly
- `CREDIT_LIMIT` preferred over `TOKENS_LIMIT` for the same window, in both orderings
- unrecognized `(unit, number)` codes skipped

It also documents the opaque `(unit, number)` window codes that were previously bare magic numbers.

**`unsloth` — document intentional zeroed buckets (nit from #1239)**

The `api_usage_events` lane passes `0` for cache_read/cache_write/reasoning because that table only records prompt/completion/total. Added a comment noting this is intentional until Unsloth exposes those columns; the chat/contextUsage path already carries them when present.

**`pricing` — document `describe_error` URL handling (nit from #1240)**

`describe_error` was made `pub` in #1240 and prints request URLs verbatim. Added a doc paragraph warning that callers must not pass it errors from clients that embed tokens/credentials in URL query parameters. Current callers use header-based auth with parameter-free URLs, which is safe.

## Dropped nits (verified stale/infeasible)

- **`pi.rs` `needless_lifetimes`** — clippy emits zero warnings after a forced recompile; the one explicit-lifetime fn is correctly not flagged.
- **Autosubmit error-cause test** — not feasible as a pure unit test (inline `reqwest` construction, no injectable seam). The load-bearing property is already tested at `pricing/fetch.rs` and `pricing/litellm.rs`.

## Testing

```
cargo build -p tokscale-cli -p tokscale-core   # clean
cargo test -p tokscale-cli zai                 # 4 new tests pass
cargo test -p tokscale-core unsloth            # 8 pass
cargo test -p tokscale-core describe_error     # existing regression test passes
cargo clippy -p tokscale-cli -p tokscale-core --lib   # no warnings
```

---
## Summary by cubic
Addresses review follow-ups from the merged Z.ai, Unsloth, and pricing error changes without changing behavior.

**Changes**
- Extracts Z.ai's session/weekly window mapping from `fetch()` into `build_metrics` and adds tests for `CREDIT_LIMIT` windows and `CREDIT_LIMIT`-over-`TOKENS_LIMIT` preference.
- Documents Z.ai's opaque `(unit, number)` window codes.
- Notes that `api_usage_events` intentionally passes `0` for cache_read/write and reasoning until Unsloth exposes those columns.
- Documents that `describe_error` prints request URLs verbatim, so callers must not pass errors from clients that embed tokens or credentials in URL query parameters.

<sup>Written for commit 185c4e1d6aa2886f8696ad43cf836f30d1453002. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1242?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/usage/zai.rs
+++ b/crates/tokscale-cli/src/commands/usage/zai.rs
@@ -72,6 +72,64 @@ pub fn has_credentials() -> bool {
         .is_ok()
 }
 
+/// Translate Z.ai's limit windows into the session/weekly/web-search metrics
+/// tokscale surfaces.
+///
+/// Z.ai encodes each window as an opaque `(unit, number)` code pair rather than
+/// a name: `(3, 5)` is the 5-hour rolling session window and `(6, 1)` is the
+/// 1-week window. Unrecognized codes are skipped rather than guessed at.
+fn build_metrics(
+    limits: &[Limit],
+    search_reset: Option<String>,
+    session_metric: &mut Option<UsageMetric>,
+    weekly_metric: &mut Option<UsageMetric>,
+    search_metric: &mut Option<UsageMetric>,
+) {
+    for limit in limits.iter() {
+        // Skip limits with no percentage rather than fabricating
+        // "0% used / 100% left" from a missing field.
+        let pct = match limit.percentage {
+            Some(p) => p.clamp(0.0, 100.0),
+            None => continue,
+        };
+
+        match limit.limit_type.as_deref() {
+            // V3 GLM Coding plans report the same (unit, number)
+            // windows as CREDIT_LIMIT instead of TOKENS_LIMIT. Prefer
+            // CREDIT_LIMIT so a plan that ever emits both for one
+            // window cannot silently last-write-wins.
+            Some("TOKENS_LIMIT") | Some("CREDIT_LIMIT") => {
+                let prefer = limit.limit_type.as_deref() == Some("CREDIT_LIMIT");
+                let (target, label) = match (limit.unit, limit.number) {
+                    (Some(3), Some(5)) => (&mut *session_metric, "Session"),
+                    (Some(6), Some(1)) => (&mut *weekly_metric, "Weekly"),
+                    _ => continue,
+                };
+                if target.is_none() || prefer {
+                    *target = Some(UsageMetric {
+                        label: label.to_string(),
+                        used_percent: pct,
+                        remaining_percent: 100.0 - pct,
+                        remaining_label: None,
+                        resets_at: None,
+                    });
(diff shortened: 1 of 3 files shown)


junhoyeo:
The `describe_error` doc addition is a good catch on my change, and I verified the claim it makes rather than taking it on faith — since making that function `pub` in #1240 is what created the hazard.

**"Current callers use header-based auth with parameter-free URLs" holds.** Five real call sites:

| site | URL |
|---|---|
| `pricing/fetch.rs:35`, `litellm.rs:538`, `openrouter.rs:264` | public pricing endpoints, unauthenticated |
| `main.rs:6048`, `main.rs:6129` | `{api_url}/api/submit`, auth via `Authorization: Bearer` header |

And nothing in the tree puts a secret in a URL: no `format!` interpolates a token/key/secret into an `https://` string, and the only `.query(` in the codebase is a rusqlite statement, not a reqwest builder.

So the comment is accurate today and is the right form of mitigation — the hazard is future callers, not present ones, and a doc constraint is what catches those. Worth keeping it phrased as a requirement on callers rather than a note about the function, which is how it reads now.

The `zai` extraction is the more valuable half. #1241's window logic was correct but only reachable through `fetch()` with live network and env, so I reviewed it by reading and reasoning about ordering rather than by running it. `build_metrics(...)` makes the CREDIT_LIMIT-vs-TOKENS_LIMIT preference testable in both orderings, which is exactly the property I could only argue for before. Documenting `(3, 5)` and `(6, 1)` matters too — those were bare magic numbers that only the issue thread explained.

PR #1246, opened 2026-08-31
fix(commandcode): parse v3 transcripts with authoritative usage and cost

alifakbar111:
Fixes Command Code usage never being recorded on `tokscale submit`: the session parser was written for an assumed flat JSONL schema (`role`/`content`/`sessionId`) that does not match the real Command Code v3 transcript format, so every transcript parsed to zero messages and nothing was ever submitted.

### 1. Parse the real v3 transcript schema
`crates/tokscale-core/src/sessions/commandcode.rs` now parses the actual on-disk format:

- `{"type":"session","version":3,"id":...,...}` header records (session id source)
- `{"type":"message",...,"message":{"role","content"},"usage":{...},"model":"provider/model"}` conversation records
- `{"type":"model_change",...,"model":...}` mid-session model switches
- `*.checkpoints.jsonl` snapshot files are skipped

### 2. Use authoritative per-request usage and cost
v3 transcripts persist the exact `usage` block Command Code ships to its dashboard: `inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheWriteTokens`, `costUsd`. These are now read verbatim and marked provider-reported, so tokscale's estimated pricing never overwrites the real cost.

`inputTokens` is **cache-inclusive** (verified against real transcripts: cache is ≥90% of input on most lines, no line has input < cache), so the cache overlap is subtracted from the input bucket — the same cache-inclusive normalization `cline` and `zcode` apply. This keeps `TokenBreakdown::total()` non-overlapping and consistent with every other client.

### 3. Model resolution
Model id comes from the line's own `model` field, falling back to the most recent `model_change`, then `~/.commandcode/config.json`, then `"unknown"`. Gateway ids (`MiniMaxAI/MiniMax-M3-Free`) are canonicalized (org prefix + `-Free` promo suffix stripped) and the provider hint (`minimax`) is recovered for pricing.

### 4. Legacy transcripts still parse
Flat-schema transcripts without a `usage` block fall back to the existing ~4 chars/token estimation using the configured agent model.

### 5. Robustness (from review)
- Tool results arrive under `role:"user"` but are turn continuations — they no longer start spurious turns
- Explicit `costUsd: 0` (free-promo models) stays provider-reported instead of being repriced at the paid rate
- Token values are clamped to a 1e12 ceiling and cost to $1M so a tampered line cannot poison aggregates; all sums saturate
- Turn-start marker survives dropped zero-total assistant lines
- `model_from_config` anchors on the `projects` directory instead of counting parents, so nested transcripts can't read a wrong `config.json`

## Verification

- 16 parser unit tests (realistic v3 sessions, cache-inclusive input, model_change, zero cost, legacy fallback, adversarial values)
- 1880 `tokscale-core` tests + 172 CLI integration tests pass
- `rustfmt` clean
- Live dry-run against real `~/.commandcode` transcripts: previously `Total tokens: 0`; now reports the full usage with cost matching Command Code's `costUsd` sum

## Notes for maintainers

The Command Code dashboard counts `inputTokens + outputTokens` for its "total tokens" display. tokscale's `Total tokens` sums all buckets, which after the cache-inclusive normalization equals `inputTokens + outputTokens` too, so numbers should now align. If the dashboard also excludes cache reads from input entirely (unverified), that would be a separate counting-policy question.

---
## Summary by cubic
Fixes Command Code usage never being recorded on `tokscale submit`. The parser was written for an assumed flat JSONL schema (`role`/`content`/`sessionId`) that does not match the real Command Code v3 transcript format, so every transcript parsed to zero messages; it now parses the actual v3 schema, submits the authoritative per-request usage and cost, and makes local `tokscale` output match `submit`.

**Bug Fixes**
- Reads provider-reported token counts and `costUsd` verbatim and marks them authoritative, so estimated pricing never overwrites the real cost and reported zero usage is not re-estimated.
- A truncated or corrupt line no longer drops every priced turn above it; the parent-chain walk fails open and counts the whole file.
- `/fork` and `/clone` copies collapse via id+timestamp dedup in both the pricing and local lanes, and legacy dedup keys are scoped by workspace so identically-named sessions in different projects don't collide.
- Abandoned `/rewind` branches are dropped; a billed zero-token line is kept.
- Resolves the model from the message's own field, falling back to the latest `model_change`, then `~/.commandcode/config.json`, then `"unknown"`; org prefixes are stripped (provider hint recovered from the raw id) and `-free` suffixes preserved.
- Usage buckets pass through unchanged so `TokenBreakdown::total()` sums without double-counting; tool results don't start spurious turns, `costUsd: 0` stays provider-reported, and token and cost values are clamped against tampering.
- Keeps the ~4 chars/token estimation fallback for legacy transcripts without a `usage` block.

<sup>Written for commit cc3a5b0acd69d63d7126064b589bb5a6d74c5252. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1246?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo:
Thanks for this — the v3 work is solid, but I have to flag one blocker before it can land, and it is the PR's central premise.

## Blocker: `inputTokens` is cache-**exclusive**, so the subtraction deletes real input

`crates/tokscale-core/src/sessions/commandcode.rs:118-119` subtracts `cacheRead + cacheWrite` from `inputTokens`. Command Code's own cost function (`dist/cli.mjs`, `estimateSessionCostUsd`) charges the **full** `inputTokens` at the input rate and adds the cache buckets on top, which is only correct if they are disjoint.

Your own fixture settles it. For the `deepseek/deepseek-v4-flash` line (input 28534, output 205, cacheRead 7424; catalog rates $0.22 / $0.66 / $0.007 per 1M from the vendor's bundled `models.md`) the transcript records `costUsd: 0.006464748`:

| reading | reconstructed cost | matches transcript |
| --- | --- | --- |
| full `inputTokens` (cache-exclusive) | `0.006464748` | **exact, to 9 s.f.** |
| this PR's subtracted input | `0.004831468` | no |

I recomputed this independently — the un-subtracted arithmetic reproduces the recorded value exactly, so that field is unmistakably a real transcript value and the buckets are disjoint. Shipping the subtraction understates that line by **25.3%**, and by your own description of the data ("cache is >=90% of input on most lines") it would remove most of Command Code's input tokens fleet-wide, leaving tokscale reporting a cost its own buckets cannot reproduce. It also understates locally-priced cost for BYOK models, where `costUsd` is absent and tokscale prices the deflated bucket itself.

Fix is one line — `let input = raw_input;` — but the three cache-inclusive tests, the module doc comment, the PR body and the README all encode the wrong assumption and need updating with it.

## Also worth addressing

**Abandoned branches are billed.** The v3 transcript is a tree, not a log. `/rewind` and `/tree` call `SessionStoreV3.branch(id)`, which moves the leaf pointer backwards, and `persistEntry` is append-only, so orphaned assistant entries keep their `usage` and `costUsd` on disk. The vendor's own `computeSessionStats` iterates `getBranch()` (the parent chain from the last entry) and never counts them; the linear scan at `:193-330` counts everything, and `parentId` is explicitly ignored at `:44`. Because those rows are marked provider-reported, repricing cannot correct it later.

**`/fork` and `/clone` double-count.** `cloneSessionFile` copies the same entries — same per-entry `id`, `usage` and `costUsd` — into a new session file and leaves the original. The dedup key at `:302` is `{session}:{index}`, so the copies get distinct keys, and the Command Code lane in `lib.rs:2228-2244` runs no `should_keep_deduped_message` at all, unlike the Augment/gjc/Copilot lanes beside it. One `/fork` doubles that branch. Careful with the fix though: the vendor's `generateEntryId` is `randomUUID().slice(0,8)` — 32 bits, unique only within one session — so a bare entry id is not a safe global key.

**Provider hint regression** (`:285-286`): the hint is now derived from `resolved_model`, which `canonicalize_model` has already org-stripped, where it previously used the full gateway id. Across the 61-id catalog, `meta/muse-spark-1.1`, `1.2` and `1.2-contributor` drop from `Some("meta")` to `None`, and the blast radius is wider for BYOK `org/model` ids. One word: `provider_hint_for_model(&raw_model)`.

**`-free` stripping** (`:404-413`): `poolside/laguna-s-2.1-free` is a current exact catalog id, so canonicalising it to `laguna-s-2.1` invents a model that does not exist. The id the strip was written for (`MiniMaxAI/MiniMax-M3-Free`) is no longer in the v1.39.3 catalog.

## Verified clean

No reasoning double-count (`reasoningTokens` does not appear anywhere in the vendor bundle, so `reasoning: 0` is right), no double-count or zero-out in the authoritative-cost/repricing interaction, no panics on malformed vendor JSON, and mixed v2/v3, truncated trailing lines and missing headers all degrade sanely.

Ground truth for the vendor claims above is the shipped `command-code@1.39.3` npm bundle (plus 0.22.0 and 0.40.6 for history), not live transcripts — there are none on this machine — so the schema claims are static analysis. The cost arithmetic, though, is checked against your own fixture and is not in doubt.

alifakbar111:
Thanks for the thorough review — you're right on every point, and the cost-arithmetic proof is decisive. I've pushed fixes for all of them:

**Blocker — cache subtraction removed.** `inputTokens` is cache-exclusive; the buckets are disjoint and pass through verbatim. I verified your arithmetic against the transcript (full input 28534 + output 205 + cacheRead 7424 at deepseek-v4-flash rates reproduces `0.006464748` exactly; the subtracted reading `0.004831468` does not). The three cache-inclusive tests, module doc, and comments are updated.

**Abandoned branches are billed.** The parser now reconstructs the active branch (the `parentId` chain from the last entry, cycle-guarded) and drops messages whose entry id is off it — matching the vendor's `getBranch()` accounting. Added `test_abandoned_rewind_branch_entries_are_dropped`.

**`/fork`/`/clone` double-count.** Dedup key is now `{entry id}:{timestamp}:{index}` — the id+timestamp survive a copy, so `should_keep_deduped_message` (now wired into the Command Code lane) collapses the copies across files, while the timestamp guards the 32-bit id space against cross-session collisions.

**Provider hint regression.** `provider_hint_for_model` now runs on the RAW model id before org-stripping, so `meta/muse-spark-1.1` resolves to `meta`.

**`-free` stripping.** Removed entirely; `poolside/laguna-s-2.1-free` stays as-is. `canonicalize_model` now only drops the org prefix.

All 17 parser tests + 1881 core + 172 CLI tests pass, clippy strict and rustfmt clean. The live dry-run now reports 66.2M tokens / $7.40 for Aug 28–31, which matches the in-range transcript sum exactly (previously 0).

junhoyeo:
Pushed four commits on top of `91281572`. Two of them fix defects I introduced in the other two, so the summary below separates what was wrong in your parser from what was wrong in my fixes.

**Fixed in your parser**

1. `to_breakdown` returned `None` when the four buckets summed to zero, so a turn reporting `{"inputTokens":0,"outputTokens":0,...}` fell through to the character-count estimator. On a 4000-character prompt that fabricated ~2000 tokens and stamped them provider-reported via the recorded `costUsd`, which makes the row permanently immune to repricing. The gate is now field *presence*, not the sum: any bucket key present means the numbers are the provider's, zeros included. A genuinely absent usage block still estimates.

2. `active_branch_ids` treated "this entry has no parent" and "this entry's named parent is missing from the file" as the same terminal case, so one unparseable or truncated line made every turn above the hole unreachable and silently dropped. Those are now separate arms. A dangling reference fails the walk and falls back to counting every entry in the file — failing open costs at most a re-counted abandoned `/rewind` branch in one session, failing closed lost every priced turn before the hole. This deliberately diverges from the vendor's `buildSessionPath`, which truncates at the same point; truncation is cosmetic when rendering a conversation and lost money when accounting for one.

3. The local lane double-counted forks. Your `dedup_key` was only honoured by the cached pricing lane in `lib.rs`; `parse_local_clients` mapped straight to `ParsedMessage` with no filter, so `tokscale` double-counted a forked session while `submit` did not — the two disagreeing on the same machine. `test_fork_copies_share_dedup_key` could not catch it because it re-implements the filter inline instead of exercising the caller; the new test drives `parse_local_clients` end to end.

**Fixed in my own commits**

4. My all-zero drop ran *before* the cost was read, so an assistant line with zero buckets and a real `costUsd` was discarded along with its charge. Zero buckets next to non-zero money is a real shape: providers that bill per request report no token counts, and a client that failed to parse a usage block still records what it was charged. The drop now requires both an empty breakdown and no positive cost. A genuine `costUsd: 0` still drops, which is what the guard was for.

   The `Rejected:` trailer I wrote on that commit is what left this in, and it is wrong — it reasons about `to_breakdown`'s old sum==0 gate, where estimation happens, then applies the conclusion to the emit-path drop, where nothing can be estimated because `to_breakdown` has already returned `Some`. Two different gates. That trailer is retracted in the new commit message.

5. The positional dedup fallback could collide across projects. Legacy flat-schema lines carry no entry id, and `resolved_session` falls back to the file stem when a transcript has neither a header nor an embedded `sessionId` — so two projects holding identically-named files produced identical `<stem>:<index>` keys and the deduping caller deleted real messages. The key now carries the workspace. It only ever needs to be unique: a legacy `/fork` lands in a differently-named file, so unlike the id-based key it never collapses copies, which makes qualifying it strictly safe.

Both new tests were confirmed to fail on the parser as it stood one commit earlier and pass after — the billed all-zero line was dropped (0 messages where 1 is correct), and the two legacy sessions produced byte-identical dedup keys.

**On your `91281572`**

Kept as-is and it is orthogonal to item 5: you fixed the timestamp source of the *id-based* key, I fixed the *positional fallback*. `test_fork_with_missing_timestamp_still_dedups` still passes. My branch predated that commit, so I replayed my four commits on top of it rather than pushing over it — nothing of yours was rewritten and this was a fast-forward.

Gates on the pushed tree: `cargo test --workspace --all-features` clean (12 suites), `cargo clippy --locked --workspace --all-features -- -D warnings` clean both with and without `--all-targets`, `cargo fmt --check` clean.

Issue #772, opened 2026-06-24
Thoughts on being able to count tokens used in memory systems like hindsight?


George-RD:

junhoyeo:
Thanks for raising this — good question, and worth being precise about since token accounting is something we try hard to keep correct.

Today tokscale attributes usage through per-client session parsers, not a generic "count everything" mechanism. Each supported client (Claude Code, Codex CLI, Gemini CLI, Cursor, Amp, Cline, Roo Code, etc. — see the table in the README) gets its own parser module under `crates/tokscale-core/src/sessions/` (e.g. `claudecode.rs`, `codex.rs`, `cline.rs`). Each parser reads that tool's local session log files (usually JSONL) and maps them into a shared `UnifiedMessage` record with a `TokenBreakdown { input, output, cache_read, cache_write, reasoning }` (see `crates/tokscale-core/src/lib.rs` and `crates/tokscale-core/src/sessions/mod.rs`). That's the whole pipeline: no log format, no counting.

Hindsight is a memory system (retain/reflect/recall), not a coding-agent CLI, so it doesn't fit the pattern out of the box the way Claude Code or Codex do — it's not obvious yet what a "session log" for it looks like, or whether it even writes one locally vs. just making LLM calls internally that get logged elsewhere (e.g. by whatever client is driving it). Adding a source for it would need two things:

1. A parser module following the existing pattern (there's also a non-JSONL precedent in `docs/providers/sakana.md` if the source is scrape/API-based rather than a local log file).
2. Fixtures — a real (or representative, redacted) sample of what a Hindsight session/log/usage record actually looks like, so the field mapping (tokens in/out/cache/reasoning, model id, session id) can be verified against ground truth rather than guessed.

If you can share a concrete example of what Hindsight emits for token usage (a log line, an API response shape, whatever's closest to source-of-truth), that'd be the fastest path to figuring out if/how this fits. Not committing to a timeline, but happy to look at a real sample.

caioribeiroclw-pixel:
There is a concrete source shape now. Hindsight `v0.8.4` exposes optional per-operation token usage in its generated API models:

```json
{
  "usage": {
    "input_tokens": 123,
    "output_tokens": 45,
    "total_tokens": 168,
    "cached_tokens": 0,
    "thoughts_tokens": 0
  }
}
```

Current boundaries from the code/OpenAPI:

[...]

Issue #1250, opened 2026-09-01
Cursor Auth is blocked by Vercel bot protection


RawToast:
It seems after #968 and #1138 I can no-longer authenticate with Cursor. I always get the 'session expired' response, even after logging out and back in to Cursor via the IDE

<img width="470" height="165" alt="Image" src="https://github.com/user-attachments/assets/885774de-7f4b-4dbb-9f69-a3271b88f8a5" />

I ran Opus on a check of this, to see what was going on. It identified #1138 (using rustls) as the culprit introduced in `v4.14.0`. After rolling back to `v4.13.0` I could login as usual.

## Opus Writeup Below

#968 is not the bug. The regression is #1138 (rustls), shipped in `v4.14.0`.

The token is fine. The desktop `state.vscdb` extraction is fine. The HTTP request is fine. `cursor.com` is behind Vercel bot protection that fingerprints the TLS ClientHello, and the rustls-backed client is blocked before the request reaches the app. tokscale reports that block as "Session token expired or invalid", which is why re-logging in can never help.

This affects `cursor sync` as well as `cursor login`, on every platform — and the manual `WorkosCursorSessionToken` paste fallback fails identically, so the usual first triage step ("paste the cookie from your browser instead") does not work either.

### Evidence

Same machine, same account, same token:

| Client | TLS stack | Result |
| --- | --- | --- |
| `curl` | LibreSSL | 200 |
| `python3` urllib | OpenSSL | 200 |
| tokscale 4.10.0 built locally (`native-tls-vendored` → Security.framework) | Secure Transport | login succeeds |
| tokscale 4.15.0 from npm (`rustls-tls-native-roots`) | rustls | fails |
| tokscale 4.15.0 built locally from `main` | rustls | fails |

Interleaved 3x, deterministic each way. The locally built `main` binary failing the same way rules out anything specific to the published npm artifact.

The decisive test: I ran the shipped 4.15.0 binary through a local MITM proxy (CA injected via `SSL_CERT_FILE`, which `rustls-native-certs` 0.8.4 honors). It sent a perfectly correct request — right cookie, right headers, right host:

```
GET /api/usage-summary HTTP/1.1
accept: */*
accept-language: en-US,en;q=0.9
cookie: WorkosCursorSessionToken=user_…  (485 bytes, correct)
referer: https://www.cursor.com/settings
user-agent: Mozilla/5.0 (Macintosh; …) Chrome/120.0.0.0 Safari/537.36
[...]

PR #1257, opened 2026-09-02
feat(clients): add Hindsight session parsing

George-RD:
Adds `hindsight` as a client, covering the LLM calls a self-hosted [Hindsight](https://github.com/vectorize-io/hindsight) memory service makes on its own behalf. Refs #772, where you asked for a parser following the existing pattern plus real fixtures.

Context for why this is worth a client: on my own instance, Hindsight burned **99.5M tokens in ~29 hours** ($50.76 at published rates) driving retain/consolidation for my coding agents. None of that was attributable to any existing parser, so my leaderboard number was materially incomplete.

## What the source actually is

Hindsight writes no local session logs, so there is nothing to scan directly. It persists one row per LLM call in its own database, exposed at `GET /v1/{tenant}/banks/{bank_id}/llm-requests` (gated behind `HINDSIGHT_API_LLM_TRACE_ENABLED`).

Critically, that table is a **rolling window, not history**: rows I observed in one poll were gone from the API ninety minutes later, and no row older than the current service run survives (`?end_date=<yesterday>` returns `total: 0`).

So this follows the Cursor precedent rather than inventing a new shape:

- `tokscale hindsight sync` mirrors the API into an append-only monthly ledger at `$HINDSIGHT_HOME/usage/<YYYY-MM>.jsonl` (default `~/.hindsight`), idempotent on Hindsight's own request UUID, atomic temp-file + rename per month.
- `sessions/hindsight.rs` parses that ledger.
- Reports and submits warn to run the sync when the ledger is missing or empty, mirroring the `cursor-cache` warning.

One note on the discussion in #772: `RetainResponse.usage` is a weaker source than it looks. On my instance `consolidation` is **92.7%** of all tokens, and it never appears in a synchronous retain response — nor does async retain. The per-request trace endpoint covers all of it.

## Ledger row

```json
{"id":"aa4cc970-...","trace_id":"51b96021-...","provider":"ollama-cloud","model":"deepseek-v4-flash:0731","operation":"retain","scope":"retain_extract_facts","started_at":"2026-09-01T08:16:51.357104+00:00","ended_at":"...","duration_ms":11744,"input_tokens":3667,"output_tokens":4063,"cached_tokens":null,"total_tokens":7730,"bank":"omp"}
```

Mapping: `model`/`provider` verbatim (Hindsight is provider-agnostic — do not assume Ollama), `trace_id` as session id (one operation run, several calls), `bank` as workspace, `operation`/`scope` as session title, `hindsight:<uuid>` as dedup key.

Fixtures are real rows from a live instance, trimmed to two banks and three operations. They contain usage metadata only — see below.

## Two accounting decisions, measured rather than assumed

Both are commented in the parser because a future reader would otherwise "fix" them:

- **`reasoning: 0`.** Hindsight's `TokenUsage` schema documents `output_tokens` as *excluding* reasoning, and that is wrong for this path. A probe returning 4 characters of visible content plus 409 characters of `reasoning` was billed **138 completion tokens** — reasoning is already inside `output_tokens`. Splitting a share out would double-count it. `thoughts_tokens` is never populated.
- **`cache_read` from `cached_tokens`, always null in practice.** Ollama Cloud's OpenAI-compatible response carries only `prompt_tokens`/`completion_tokens`/`total_tokens`, with no `prompt_tokens_details`; sending an identical 1,405-token prompt twice produced byte-identical usage. It stays its own bucket rather than folded into input, so a provider that *does* report cache reads prices correctly.

## Privacy

The API also returns `input`, `output`, and `metadata`, which carry the prompt and the memory content being reasoned over. `RawApiLlmRequest` has **no field for any of them**, so serde drops them before they enter the process — a stronger guarantee than reading and discarding. A test deserialises a payload containing marker strings in all three and asserts the serialised ledger row cannot contain them.

## Checklist (per CONTRIBUTING.md)

- [x] Registry entry (`Hindsight = 52`), `ClientId::COUNT` 52 -> 53
- [x] Parser + module registration
- [x] Scanner discovery and `lib.rs` dispatch (the documented silent-failure step — verified by hand against real data)
- [x] CLI: `ClientFilter` variant and mappings, `CLIENT_UI` entry (hotkey `H`, unused)
- [x] Frontend: `SUPPORTED_CLIENT_TYPES`, `SOURCE_DISPLAY_NAMES`, `SOURCE_LOGOS`, `SOURCE_COLORS` (`#0891B2`, AA on both themes)
- [x] Logo asset `.github/assets/client-hindsight.png` (Hindsight's own logo)
- [x] `README.md` plus `README.ko.md`, `README.ja.md`, `README.zh-cn.md`

## Verification

`cargo fmt --all -- --check`, `cargo clippy --locked --workspace --all-features -- -D warnings`, and `cargo test --workspace --all-features` are clean (3,378 tests). `bun run --cwd packages/frontend typecheck` and `clientRegistry.test.ts` pass.

Against real local data:

```
$ tokscale hindsight sync
  Synced 41 new Hindsight LLM request(s) (1810 total, 1155260 tokens).

$ tokscale --client hindsight --no-spinner
│ Hindsight │ Ollama Cloud │ deepseek-v4-flash:0731 │ 91,600,603 │ 7,921,904 │ 723ms │ $50.76 │
  Total: 1,810 messages, 99,522,507 tokens, $50.76

$ tokscale --client hindsight --group-by workspace,model --no-spinner
│ omp       │ deepseek-v4-flash:0731 │ 568ms │ $46.00 │
│ george    │ deepseek-v4-flash:0731 │  2.5s │  $4.77 │
```

`submit --dry-run` reports the same totals under `Clients: hindsight`. I have not submitted from this build.

Happy to split the sync command out of the parser, rename anything, or drop the session-title mapping if you would rather keep it minimal.

---
## Summary by cubic
Adds Hindsight as a tracked client so the LLM usage from its self-hosted memory service shows up in reports instead of being invisible to every existing parser. Because Hindsight writes no local session logs, `tokscale hindsight sync` mirrors its rolling LLM trace API into an append-only monthly JSONL ledger under `$HINDSIGHT_HOME/usage/`; the new parser reads that ledger and reports warn to run the sync when it's missing.

- Reasoning is reported as 0 because Hindsight's `output_tokens` already includes it; splitting it out would double-count.
- `cache_read` maps from `cached_tokens`, which is null on every observed provider response, but stays its own bucket so providers that do report cache reads price correctly.
- Prompt and memory content never enters the process: the raw API request struct has no fields for `input`, `output`, or `metadata`, so serde drops them before they reach disk.
- The sync client is built through the shared `tokscale_core::http` rustls factory rather than `reqwest`'s default builder, so its TLS backend matches the rest of the codebase.

<sup>Written for commit 2dcd4a51f28c7113aba26e158090253d9e3b44ba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1257?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo:
Reviewed this in full. The integration itself is right, and the parts that usually go wrong in a new-client PR are all done: `define_clients!` entry with `PathRoot::EnvVar`, `ClientId::COUNT` 52 → 53, both parse lanes wired (`parse_all_messages_streaming`'s cached lane and `parse_local_clients`) with the same `hindsight:{id}` dedup key, all four frontend registries (`SUPPORTED_CLIENT_TYPES`, `SOURCE_DISPLAY_NAMES`, `SOURCE_LOGOS`, `SOURCE_COLORS`), a real logo at `.github/assets/client-hindsight.png`, and four READMEs. 13 tests across the CLI, the parser and the scanner.

I rebased onto main and pushed one commit (`73c9b703`) rather than sending you round again.

**The reason it needed a fix:** this branch was cut at `49f80a18`, five commits before #1252 landed. On that base your CI is green and correct. Rebased onto current main, `sync_hindsight` fails two independent guards:

```
error: use of a disallowed method `reqwest::Client::builder`
  --> crates/tokscale-cli/src/hindsight.rs:356:24
  = note: use tokscale_core::http::client_builder() so the TLS backend is explicit (#1250)

test no_client_is_constructed_outside_the_allowlist ... FAILED
  crates/tokscale-cli/src/hindsight.rs:357: reqwest::Client::builder()
```

Not a style rule. Cursor's ClientHello workaround (#1250) pulled reqwest's `native-tls` feature into the graph, `native-tls` implies `default-tls`, and `TlsBackend::default()` resolves to native-tls whenever `default-tls` is on — so an unqualified builder is a native-tls client, not the rustls one the surrounding code assumes. Swapped for `tokscale_core::http::client_builder()`.

I checked both guards actually bite before trusting the fix: reverting the one line reproduces the clippy error and `no_client_is_constructed_outside_the_allowlist` fails naming your file; restoring it gives `8 passed; 0 failed`.

**Second change, doc only.** `RawApiLlmRequest`'s doc comment said content fields are "mapped to `Option<Value>` and discarded". They aren't mapped at all — there are no such fields, and serde drops unknown keys, which is what the note at the end of the field list says and what `test_content_fields_are_never_deserialized_or_persisted` asserts. The `///` line was describing a weaker guarantee than the code gives, so I made it agree with the rest.

**Gates on the rebased branch:** `cargo fmt --check` 0, `cargo clippy --locked --workspace --all-features -- -D warnings` 0, the same with `--all-targets` 0, `cargo test --workspace` 0 — 13 suites `ok`, 0 failures.

One thing I did not touch, for the record: `.max(0)` on the token fields in `sessions/hindsight.rs` looked redundant until I checked the types. `HindsightRecord`'s fields are `Option<i64>`, not `Option<u64>` like the API structs in the CLI, so the clamp is load-bearing against a hand-edited ledger. Leaving it.

Small copy nit, not worth a commit: the setup warning reads "Tokscale does not parse local Hindsight database." — wants "the local Hindsight database". It appears twice.

Merging once CI is green on the rebased head.

PR #1267, opened 2026-09-02
fix(droid): show transcript-dated usage on the web graph

Jackson57279:
## Summary

Droid already scans in the TUI. Two independent bugs stopped that usage from showing correctly on the profile.

1. **CLI too old (already fixed on `main`).** Tokscale 4.13 **drops** unpriced Factory models (`ox-alpha-(openrouter)-0` and other routing labels) instead of submitting them at $0. A 4.15+ CLI keeps those tokens (`feat(submit): retain unpriced tokens`). That is why a full submit from current `main` can populate Droid days even before this PR ships.

2. **This PR (server + profile).** After Droid is in the lifetime high-water, a transcript re-date used to leave tokens on the old day: the per-day merge refuses decreases, so yesterday stayed empty. A **full** Droid snapshot that still covers the credited lifetime now **replaces stored Droid cells** (lifetime total does not inflate). Token-positive **$0** days also count as active on the contribution graph.

## Reviewer map

- `packages/frontend/src/lib/db/parserHighWater.ts` — Droid `layoutDays` when the incoming snapshot covers the credited aggregate
- `packages/frontend/src/app/api/submit/route.ts` — apply those cells as replacements, including days that left the payload
- `packages/frontend/src/lib/publicProfileData.ts` / `utils.ts` / `StatsPanel.tsx` — intensity and activity from tokens, not cost
- Tests: `parserHighWater.test.ts`, `submitAntigravityCliHighWater.test.ts` (Droid layout), `graphClientFilter.test.ts`

Copilot and Antigravity stay increment-only. Partial/`--today` Droid submits still freeze.

## Test plan

- [x] `bun run --cwd packages/frontend test` (955 passing)
- [x] `bun run --cwd packages/frontend typecheck`
- [ ] After merge: full `tokscale submit` with **4.15+** (no `--today`) so production can rewrite stored Droid days to the transcript layout

### DIFF
--- a/packages/frontend/__tests__/api/submitAntigravityCliHighWater.test.ts
+++ b/packages/frontend/__tests__/api/submitAntigravityCliHighWater.test.ts
@@ -1,3 +1,6 @@
+import { readFileSync } from "node:fs";
+import { resolve } from "node:path";
+
 import { beforeAll, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { SUPPORTED_VERSIONED_PARSERS } from "../../src/lib/db/parserHighWater";
@@ -285,6 +288,7 @@ function isDailyBreakdownInsert(text: string): boolean {
 
 function installTx(store: Store) {
   const executedSqlArgs: unknown[] = [];
+  const selectedColumns: Array<Record<string, unknown>> = [];
 
   function applyDailyBreakdownWrite(sqlArg: unknown): void {
     const strings: string[] = [];
@@ -360,9 +364,10 @@ function installTx(store: Store) {
       };
       return builder;
     }),
-    select: vi.fn((columns: Record<string, unknown>) =>
-      makeAwaitableBuilder(selectResult(columns, store)),
-    ),
+    select: vi.fn((columns: Record<string, unknown>) => {
+      selectedColumns.push(columns);
+      return makeAwaitableBuilder(selectResult(columns, store));
+    }),
     insert: vi.fn(() => {
       const builder = {
         values: vi.fn(() => builder),
@@ -392,7 +397,7 @@ function installTx(store: Store) {
     async (callback: (transaction: typeof tx) => Promise<unknown>) =>
       callback(tx),
   );
-  return { executedSqlArgs };
+  return { executedSqlArgs, selectedColumns };
 }
 
 function submissionBody(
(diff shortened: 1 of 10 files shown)


junhoyeo:
Thanks — the core fix is right. The Droid parser already dates usage per transcript turn; the bug was the server treating a re-dated snapshot as a reshuffle and crediting nothing. The replace plan mode fixes that and is idempotent with no double counting, and the five tests that fail against main's source all pass here.

I pushed two commits.

**What I changed.**

1. **False warning.** Emptied days that still had another client's stored cell produced a "Preserved X because it disappeared" warning. It now fires only when the day genuinely disappeared for *that* client.
2. **Code and comments disagreed** about deleted local Droid history. Reconciled in favour of what the code does.
3. **The intensity comment cited `buildContributionLayout`**, which does not exist anywhere in the repo — `grep` returns only the comment itself. The real embed function is `layoutContributions` in `lib/embed/embedShared.ts`. The substantive claim checks out (`getUserEmbedStats` and `renderIsometric3DSvg` both recompute intensity from tokens), so I fixed the reference rather than reverting the behaviour. Flagging it because that comment is the only written record of a deliberate divergence from the Rust TUI — if the cost→tokens switch was meant to be part of this PR, it is worth calling out in the description.
4. **One production bug, which I'd like you to sanity-check.** `route.ts` computed `dateStart`/`dateEnd` as `MIN(CASE WHEN tokens > 0 THEN date END)`, but `submissions.date_start` and `date_end` are `NOT NULL`. That aggregate returns NULL when no row in scope has `tokens > 0`, and the STEP 3e update is unconditional — so it writes NULL into a NOT NULL column and **aborts the entire submit transaction**. It hits any user whose stored history is all legacy tokenless Cursor rows (client `cursor`, model `premium-tool-call`, tokens 0, cost > 0), a shape `isLegacyTokenlessCursorClient` explicitly permits. The submit route tests mock the DB and never enforce NOT NULL, so the green suite did not catch it. Now wrapped in `COALESCE(..., MIN(date))`, matching every sibling column in that same SELECT.

**What I left for you.** Two symptoms of one design question:

**A. A layout rewrite that deletes a stored day lowers the client's lifetime cost.** Post `08-07 100,000 tok / $10`, then `08-08 100,000 tok / $1` — tokens tie, so it takes the replace path, 08-07 is dropped, stored cost falls $10 → $1.

**B. A re-date that shrinks a day's cost while tokens stay flat is accepted.** Post `08-07 100,000 tok / 5 msgs / $10`, then `08-07 1,000/1/$0.01` + `08-08 49,000/2/$0.49` + `08-09 50,000/2/$0.50`. Tokens tie at 100,000 and messages tie at 5, so `snapshotCoversCredited` holds, the plan is `replace`, and cost goes $10 → $1.

Both verified against this branch as it now stands.

I tried gating on the date set — floor the cost when the rewrite drops a stored day — and it was wrong twice over, so I backed it out rather than ship it:

- **The floor survived exactly one submission.** `dropsStoredDays` is computed per-submission, so the next identical snapshot drops nothing, takes the ungated path, and `applyCostCompleteness` overwrites the floor away. A CLI that scans on every run sends that identical snapshot immediately.
- **Gating on dates rather than cost has it backwards.** One 1,000-token crumb left on the source day bypasses the gate entirely — and that is the *common* shape, since per-turn re-dating usually leaves a session's first turns in place (your own `PER_GENERATION_DATING` fixture keeps 40,000 tokens on 08-07). The gate would fire on the rare case and not the frequent one.

If you want a floor here, it probably has to key on the client's total cost falling, and be durable across submissions rather than recomputed per request.

junhoyeo (approved):
Reviewed against the audit findings, re-verified locally (fmt, clippy with and without --all-targets, full workspace tests), and CI is green. Maintainer commits on this branch are noted in the comment above.

PR #1271, opened 2026-09-03
fix(pricing): prefer proven archive tariff over unverified reseller guess

thedavidweng:
Live datasets key some first-party models only under reseller roots (`deepinfra/anthropic/claude-opus-4-8`, `z-ai/glm-5.2` — nested provider segments, not first-party roots), so a first-party hint resolves there as an unverified ModelPart guess. Those guesses won by precedence even though the retirement archive holds (or, in follow-ups, will hold) the publishing endpoint's own exact tariff, and the usage submitted at $0.00 with a cost-incomplete day. Verified against the live datasets: before this change `anthropic/claude-opus-4-8` resolved `model_part`/`LiteLLM`/`covers=false`; after, `provider_scoped`/`Tokscale Archive`/`covers=true`.

Every unverified fallback stage in `lookup_auto` (model-part, alias and prefix matches) now routes through a `prefer_proven_archive` guard: when the upstream winner is not submission-safe and the archive holds a submission-safe row for the hinted endpoint, the proven tariff wins. Safe upstream rows (including a live first-party exact key) still win untouched, and anything the archive cannot prove stays exactly as before — display estimates never change, only their publishability when a proven tariff exists. Normalization provenance (`with_normalization`) is preserved on passthrough. `exact_match_archive` additionally accepts non-Claude ids verbatim (previously only Claude ids canonicalized, everything else returned None), still gated on the hint naming the key's publishing endpoint, in preparation for first-party archive rows from other vendors. Two fallback stages are proven necessary by failing-then-passing tests (the provider-pinned chooser for opus-4-8, the model-part fallback for glm-5.2); the rest is uniform application of the same guard, which is an identity function wherever no proven archive row exists.

## What's Changed
* fix(pricing): prefer proven archive tariff over unverified reseller guess
* fix(pricing): apply proven-archive preference at every unverified fallback

## Verification
* `cargo clippy --locked --workspace --all-features -- -D warnings`
* `cargo test --locked -p tokscale-core --lib` (2045 passed on this branch, including the new `unsafe_reseller_row_yields_to_proven_archive` fixture shaped exactly like the live LiteLLM row, plus all pre-existing archive precedence pins such as `live_upstream_price_wins_over_the_retired_model_archive` and `codex_auto_review_is_never_served_from_the_archive`)
* Live-dataset probe before/after for `anthropic/claude-opus-4-8` and six other first-party pairs (probe script removed after use; the other six stay unpriced here and are covered by follow-up archive-row PRs per vendor, stacked on this one)

**Full Changelog**: https://github.com/junhoyeo/tokscale/compare/v4.15.1...fix/pricing-prefer-proven-archive

### DIFF
--- a/crates/tokscale-core/src/pricing/lookup.rs
+++ b/crates/tokscale-core/src/pricing/lookup.rs
@@ -880,21 +880,7 @@ impl PricingLookup {
             self.exact_match_models_dev_for_provider(model_id, provider_id),
             provider_id,
         ) {
-            // A provider-proven first-party snapshot beats an unverified
-            // guess. Reseller rows (`deepinfra/anthropic/...`) match here by
-            // model part while the archive holds the publishing endpoint's
-            // own exact tariff further down; letting the guess win prices
-            // first-party usage at a third party's rate and reports it as
-            // unpublishable. Safe upstream rows still win untouched.
-            if !result.evidence.is_submission_safe() {
-                let proven = self
-                    .exact_match_archive(model_id, provider_id)
-                    .filter(|archived| archived.evidence.is_submission_safe());
-                if let Some(archived) = proven {
-                    return Some(archived);
-                }
-            }
-            return Some(result);
+            return Some(self.prefer_proven_archive(result, model_id, provider_id));
         }
 
         if let Some(result) = exact_litellm {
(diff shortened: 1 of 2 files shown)


thedavidweng:
Consolidated into #1274, which combines the archive precedence engine and the Zhipu, Xiaomi, and Tencent first-party tariffs into a single PR.

PR #1273, opened 2026-09-03
fix(pricing): archive Xiaomi MiMo-V2.5 first-party tariff

thedavidweng:
No live dataset carries a first-party Xiaomi row — only the `openrouter/xiaomi/*` marketplace key — so a `xiaomi` hint for `mimo-v2.5` resolved as an unverified guess and submitted at $0.00. Part of the per-vendor archive-row series: depends on #1271 (proven-archive precedence) and #1272 (Zhipu rows, the parent commit here); merge in order and this shrinks to the single Xiaomi commit.

Adds `xiaomi/mimo-v2.5` to the archive at $0.14/$0.28 per 1M with $0.0028 cached input, verified 2026-09-03 against https://mimo.mi.com/docs/en-US/pricing (flat to 1M context since the 2026-05-27 permanent cut). Xiaomi publishes no cache-write tariff, so the row carries none and cache-write-bearing usage stays unpriced rather than billed at an invented rate (pinned by test).

## What's Changed
* fix(pricing): archive Xiaomi MiMo-V2.5 first-party tariff

## Verification
* `cargo fmt --all -- --check`
* `cargo clippy --locked -p tokscale-core --all-features -- -D warnings`
* `cargo test --locked -p tokscale-core --lib` (2048 passed, including new `xiaomi_mimo_rate_resolves_first_party_over_marketplace_row` with an `openrouter/xiaomi/` shadow fixture plus the unpublished-bucket pin)

---
## Summary by cubic
Archives first-party tariffs for Xiaomi MiMo-V2.5 (and Zhipu GLM-5.2/5.3) and makes proven archive rows win over unverified reseller guesses, so a `xiaomi` or `zhipu` hint no longer resolves to a $0.00 guess.

**Behavior change**
- Archive tariffs now take precedence over unverified upstream matches at every fallback stage.
- Models without a published cache-write tariff leave that bucket unpriced instead of inventing a rate.

**New archives**
- Adds `xiaomi/mimo-v2.5` at $0.14/$0.28 per 1M with $0.0028 cached input.
- Adds `zhipu/glm-5.2` and `zhipu/glm-5.3` at $1.40/$4.40 per 1M with $0.26 cached input.

<sup>Written for commit 444a3cd74d999378c309430a1dfff70cfbc762fc. Summary will update on new commits.</sup>

[...]

PR #1272, opened 2026-09-03
fix(pricing): archive Zhipu GLM-5.2/5.3 first-party tariffs

thedavidweng:
No live dataset carries a first-party Zhipu row — only the `z-ai/*` reseller keys — so a `zhipu` hint for `glm-5.2`/`glm-5.3` resolved as an unverified guess and submitted at $0.00. Stacked on #1271 (which lets a proven archive tariff win over such guesses); without it these rows would stay shadowed.

Adds `zhipu/glm-5.2` and `zhipu/glm-5.3` to the archive at $1.40/$4.40 per 1M with $0.26 cached input, verified 2026-09-03 against https://docs.z.ai/guides/overview/pricing and https://bigmodel.cn/pricing (flat 1M context, no tiers). Zhipu publishes no cache-write tariff, so the rows carry none: cache-write-bearing usage stays unpriced rather than billed at an invented rate (pinned by test). Same-provider proxy spellings (`babeltown/`, `owo_ollama/`, `custom/` prefixes) intentionally still resolve as estimates — only the publishing endpoint's own hint is submission-safe.

## What's Changed
* fix(pricing): archive Zhipu GLM-5.2/5.3 first-party tariffs

## Verification
* `cargo fmt --all -- --check`
* `cargo clippy --locked -p tokscale-core --all-features -- -D warnings`
* `cargo test --locked -p tokscale-core --lib` (2047 passed, including new `zhipu_glm_rates_resolve_first_party_over_reseller_row` with a `z-ai/` reseller shadow fixture and `zhipu_cache_write_usage_stays_unpriced` for the unpublished bucket)

Depends on #1271 — the diff currently includes its two commits; merge that first and this shrinks to the single Zhipu commit.

---
## Summary by cubic
Fixes `zhipu` hints for `glm-5.2`/`glm-5.3` resolving as unverified guesses billed at $0.00 by archiving first-party tariffs ($1.40/$4.40 per 1M, $0.26 cached input), and makes proven archive tariffs win over unverified reseller matches at every fallback stage.

**Bug Fixes**
- Zhipu publishes no cache-write tariff, so cache-write-bearing usage stays unpriced rather than billed at an invented rate.
- Reseller rows (`z-ai/*`, `deepinfra/*`) that match only by model part no longer shadow the publishing endpoint's archived tariff; same-provider proxy spellings still resolve as estimates.

Depends on #1271 — merge that first; this diff includes those changes.

[...]

PR #1274, opened 2026-09-03
fix(pricing): archive first-party tariffs for Zhipu, Xiaomi and Tencent models

thedavidweng:
Live datasets key many first-party models only under reseller roots (`deepinfra/anthropic/claude-opus-4-8`, `deepinfra/tencent/Hy3`, `openrouter/xiaomi/mimo-v2.5`, `z-ai/glm-5.2`), so a first-party hint resolves as an unverified ModelPart guess. Those guesses won by precedence even though the retirement archive holds the publishing endpoint's exact tariff, and the usage submitted at $0.00 with a cost-incomplete day.

This consolidates first-party tariff archives for Chinese LLM providers (Zhipu GLM-5.2/5.3, Xiaomi MiMo-V2.5, Tencent Hunyuan HY3/HY4-Preview) and makes proven archive rows win over unverified reseller guesses at every fallback stage:

- **Provenance precedence**: Every unverified fallback stage in `lookup_auto` routes through `prefer_proven_archive`. When an upstream winner is not submission-safe and the archive holds a submission-safe row for the hinted endpoint, the proven tariff wins.
- **Provider-qualified model IDs**: Strip leading provider prefixes before archive matching non-Claude models so qualified IDs like `tencent/hy3`, `zhipu/glm-5.2`, and `xiaomi/mimo-v2.5` match directly as submission-safe.
- **Zhipu GLM-5.2/5.3**: Archive at $1.40/$4.40 per 1M with $0.26 cached input, verified 2026-09-03 against Zhipu and BigModel pricing.
- **Xiaomi MiMo-V2.5**: Archive at $0.14/$0.28 per 1M with $0.0028 cached input, verified 2026-09-03 against Xiaomi MiMo docs. Supports both `xiaomi` and `mimo` endpoint roots.
- **Tencent Hunyuan HY3/HY4-Preview**: Archive at $0.132/$0.528 (HY3) and $0.834/$2.501 (HY4-Preview) per 1M with verified cache-hit rates from Tencent Cloud international USD sheet. Supports both `tencent` and `hunyuan` endpoint roots.
- **Unpublished cache-write handling**: Models without published cache-write tariffs leave that bucket unpriced rather than billing an invented rate.

Supersedes and consolidates #1271, #1272, and #1273 into a single cohesive PR.

## What's Changed
* fix(pricing): prefer proven archive tariff over unverified reseller guess
* fix(pricing): apply proven-archive preference at every unverified fallback
* fix(pricing): archive Zhipu GLM-5.2/5.3 first-party tariffs
* fix(pricing): archive Xiaomi MiMo-V2.5 first-party tariff
* fix(pricing): archive Tencent Hunyuan HY3/HY4-Preview first-party tariffs
* fix(pricing): preserve archive resolution for provider-qualified model ids

## Verification
* `cargo clippy --locked -p tokscale-core --all-features -- -D warnings`
* `cargo test --locked -p tokscale-core --lib` (2049 passed, covering all provider hints, qualified IDs, shadow reseller fixtures, and unpublished-bucket behavior)

### DIFF
--- a/crates/tokscale-core/src/pricing/lookup.rs
+++ b/crates/tokscale-core/src/pricing/lookup.rs
@@ -880,21 +880,7 @@ impl PricingLookup {
             self.exact_match_models_dev_for_provider(model_id, provider_id),
             provider_id,
         ) {
-            // A provider-proven first-party snapshot beats an unverified
-            // guess. Reseller rows (`deepinfra/anthropic/...`) match here by
-            // model part while the archive holds the publishing endpoint's
-            // own exact tariff further down; letting the guess win prices
-            // first-party usage at a third party's rate and reports it as
-            // unpublishable. Safe upstream rows still win untouched.
-            if !result.evidence.is_submission_safe() {
-                let proven = self
-                    .exact_match_archive(model_id, provider_id)
-                    .filter(|archived| archived.evidence.is_submission_safe());
-                if let Some(archived) = proven {
-                    return Some(archived);
-                }
-            }
-            return Some(result);
+            return Some(self.prefer_proven_archive(result, model_id, provider_id));
         }
 
         if let Some(result) = exact_litellm {
@@ -918,11 +904,11 @@ impl PricingLookup {
         // matching key falls through to the canonical resolution below.
         if provider_id.is_some() {
             if let Some(result) = self.exact_match_models_dev_for_provider(model_id, provider_id) {
-                return Some(result);
+                return Some(self.prefer_proven_archive(result, model_id, provider_id));
             }
         }
         if let Some(result) = self.exact_match_openrouter_model_part(model_id) {
-            return Some(result);
+            return Some(self.prefer_proven_archive(result, model_id, provider_id));
         }
 
         // Separator-normalized exact passes against the canonical sources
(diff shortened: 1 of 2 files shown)


junhoyeo:
Thanks — I pushed one commit to this branch (`2f0a6849`) and left the rest for you.

**What I changed.** Archived tariffs were reachable through the generic suffix-stripping fallback, so `mimo-v2.5-pro`, `glm-5.3-flash`, `glm-5.2-air` and `hy3-instruct` resolved to the *base* model's tariff and were marked submission-safe, where on main they are unpriced. The MiMo Code parser's only fixture model is the `-pro` SKU, so this fires on real data.

The fix holds the archive out of the lookup for eroded candidates (`allow_archive`) rather than filtering it out of the result. Filtering the result discards the live row too, because `prefer_proven_archive` *replaces* the live row rather than sitting beside it — I tried that first and every eroded SKU resolved to $0.00.

Two things I deliberately did **not** touch:

**1. Hint spelling decides whether the archive can override a live first-party row.** The PR treats `z-ai/` as a reseller root, but z.ai is Zhipu's own platform and is already listed among the original provider prefixes. `provider_identity.rs` has no `zhipu` entry, so a `zhipu` hint never canonicalizes to `zai`, and depending on which spelling the client reports, a live first-party row can be silently overridden by the archive.

I left this to you because the fix is a design call: folding the spellings in `provider_identity.rs` also changes how suffix-eroded ids get classified. When I tried it, `annotate_direct` started promoting eroded lookups under a `zhipu` hint from `ModelPart` to `ProviderScoped`, which republishes unknown GLM SKUs (`glm-5.2-air`, `-flash`, `-x`, `-thinking`, `glm-5.3-air`) at the base rate and marks them submission-safe — the same outcome as the bug above, arriving by another route. Worth knowing before you pick an approach.

**2. `lookup.rs:1552` overreaches.** The doc ends: *"so display estimates never change, only their publishability when a proven tariff exists."* Where the archive does prove the endpoint, the displayed estimate does change — `mimo-v2.5` with hint `xiaomi` shows $2.4800 on main and $0.4228 on this branch (1M input / 1M output / 1M cache-read). Worth rewording to say estimates change exactly where a proven tariff replaces an unverified one.

The Zhipu and Tencent rates check out against the vendor pages. I could not verify the Xiaomi MiMo rate — that pricing page is JS-rendered — so I left the number alone.

junhoyeo (approved):
Reviewed against the audit findings, re-verified locally (fmt, clippy with and without --all-targets, full workspace tests), and CI is green. Maintainer commits on this branch are noted in the comment above.

PR #1276, opened 2026-09-03
feat(pricing): price dead free-preview models at published zero

thedavidweng:
`ox-alpha` and `x-preview-f-free` are the August 2026 stealth preview of Z.AI's GLM-5.3-Flash, served free for a week and then retired — high-attention usage that submitted at $0.00 with a cost-incomplete day and a warning per provider/model pair. Both ids carry a published $0 (OpenRouter: "This model is free to use", Z.AI: "tested GLM-5.3-Flash anonymously as ox-alpha on OpenCode and OpenRouter", Stealth EULA: access "free of charge", Zen catalog: `Ox Alpha Free (Unlimited)` at input = 0, output = 0, cache_read = 0, now deprecated), so they submit at $0 with no warning. Depends on #1271 (proven-rate precedence + verbatim archive fallback); merge that first and this shrinks to the free-preview commits. The diff currently also renames #1271's `prefer_proven_archive` helper to `prefer_proven_rate` as it gains the second table — flagging since it touches unmerged code.

Why a compiled-in table when upstream models.dev tracks these (it does — live api.json carries `opencode/x-preview-f-free` and `opencode-go/ox-alpha-free` at $0, both deprecated; tokscale fetches upstream `https://models.dev/api.json` directly, no mirror needed)? Two gaps verified live: sessions spell the gateway `opencode-zen`/`opencode_zen` against the `opencode` root, so the row only ever matches as an unverified cross-provider guess — and proving that identity by folding would widen blast radius to every zen-hinted paid resolution, which needs endpoint-billing equivalence this PR does not claim. And both upstream rows omit the cache-write bucket while the opencode parser always populates it, so write-bearing usage would stay uncovered. No `ox-alpha` row exists under any provider at all (only the `-free`-suffixed sibling), so for that id compiled-in data or `custom-pricing.json` are the only paths. The $0 here is final (deprecated ids), not a guess — contrast `openai/codex-auto-review`, which stays unpriced.

Adds a `free_preview` built-in table (`ox-alpha`, `x-preview-f-free`, every bucket an explicit zero, no tiers) matching bare with `BuiltIn` evidence regardless of gateway, at the Cursor/Sakana precedence slot plus consultation from the proven-rate guard so the live models.dev `opencode/x-preview-f-free` model-part guess cannot shadow it. All-zero rows cover every usage shape via the existing `quotes_zero_for_every_published_rate` rule (same treatment as the documented-free `opencode/nemotron-3-ultra-free` row). A live first-party exact row still beats the $0 snapshot, and `has_upstream_dataset` keeps excluding compiled-in tables. Deliberate non-goals: no generic `:free`-suffix rule (convention is weaker evidence than a per-id published $0; those ids stay on `custom-pricing.json`) and no `ox-alpha-free` row (seen in the wild, no citable tariff).

## What's Changed
* feat(pricing): price dead free-preview models at published zero

## Verification
* `cargo clippy --locked --workspace --all-features -- -D warnings`
* `cargo test --locked -p tokscale-core --lib` (2048 passed, including `free_preview_ids_resolve_at_zero_from_any_gateway` across all gateways in the motivating report plus router-qualified `stealth/ox-alpha`, `free_preview_zero_beats_unverified_reseller_guess` against a paid reseller fixture, and `live_upstream_exact_wins_over_free_preview_zero` pinning live-beats-snapshot)
* `cargo test -p tokscale-cli --test cli_tests test_submit_free_preview_models_upload_without_warnings` (opencode fixtures for both ids submit in dry-run with zero `unpriced` output and counted tokens)
* Live-dataset probe before the change (`ox-alpha` unresolvable, `x-preview-f-free` unverified model-part guess; probe script removed after use)

### DIFF
--- a/crates/tokscale-cli/tests/cli_tests.rs
+++ b/crates/tokscale-cli/tests/cli_tests.rs
@@ -5096,6 +5096,82 @@ fn test_submit_includes_unpriced_usage_at_zero_and_keeps_the_rest() {
     );
 }
 
+/// Dead free-preview incarnations (`ox-alpha`, `x-preview-f-free`) carry a
+/// published $0, so they submit with the priced usage: no per-row warning,
+/// no aggregate line, no fix hint.
+#[test]
+fn test_submit_free_preview_models_upload_without_warnings() {
+    let tmp = create_temp_fixture_dir();
+    prime_pricing_cache_with_a_priced_model(tmp.path());
+    write_fake_credentials(tmp.path());
+    let preview_dir = tmp
+        .path()
+        .join(".local/share/opencode/storage/message/free-preview");
+    fs::create_dir_all(&preview_dir).unwrap();
+    for (file, id, session, provider, input, output) in [
+        ("ox.json", "ox", "free-preview-ox", "stealth", 1000, 500),
+        (
+            "xpreview.json",
+            "xpreview",
+            "free-preview-x",
+            "opencode-zen",
+            2000,
+            100,
+        ),
+    ] {
+        let model_id = if id == "ox" {
+            "stealth/ox-alpha".to_string()
+        } else {
+            "x-preview-f-free".to_string()
+        };
+        fs::write(
+preview_dir.join(file),
+format!(
+r#"{{
+"id": "{id}",
+"sessionID": "{session}",
+"role": "assistant",
+"modelID": "{model_id}",
+"providerID": "{provider}",
+"cost": 0,
+"tokens": {{ "input": {input}, "output": {output}, "reasoning": 0, "cache": {{ "read": 0, "write": 0 }} }},
+"time": {{ "created": 1736510400000.0 }}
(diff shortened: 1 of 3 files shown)


thedavidweng:
Withdrawn in favor of #1277.

Review feedback showed the compiled-in table duplicates live upstream data: models.dev already tracks `opencode/x-preview-f-free` and `opencode-go/ox-alpha-free` at $0.00 (both deprecated), and I verified live that the canonical pairs resolve exact/Models.dev/safe/covers with cache-write-bearing usage. #1277 instead pins the shorthand spellings (`ox-alpha`, `stealth/ox-alpha`, `x-preview-f-free`) to those canonical keys via the existing alias mechanism — no hardcoded prices, no new lookup tables, no constructor churn — following the `minimax-m3` precedent already in `aliases.rs`.

PR #1278, opened 2026-09-04
feat(openclaw): read the per-agent SQLite transcript store and own Codex-harness rollouts

PJunhyuk:
Adds support for the SQLite transcript store current OpenClaw (2026.x) writes, and fixes how usage OpenClaw runs through the Codex app-server harness is attributed. No new client id; this is additional storage for the existing `openclaw` client, so nothing changes on the frontend or in the submit protocol.

## Problem

Since OpenClaw moved live transcripts into `~/.openclaw/agents/<agentId>/agent/openclaw-agent.sqlite` (table `transcript_events`, one JSON event per row), the `sessions/` directory only receives `.jsonl.deleted.<ts>` / `.jsonl.reset.<ts>` archives published from that database. Tokscale scanned `*.jsonl*` only, so on a machine running OpenClaw 2026.8.2 the 4.15.0 binary reported **zero** OpenClaw usage for the current month while the agent stores and their Codex rollouts held about two thousand token-bearing assistant events for that month. Verified against the [upstream schema](https://github.com/openclaw/openclaw/blob/main/src/state/openclaw-agent-schema.sql) and the [session storage docs](https://github.com/openclaw/openclaw/blob/main/docs/concepts/session.md).

## What changes

**SQLite store.** The scanner discovers `<agents root>/<agentId>/agent/openclaw-agent.sqlite` under every OpenClaw root it already walks (default, legacy rebrand paths, `extraScanPaths.openclaw`, `TOKSCALE_EXTRA_DIRS`) into `ScanResult::openclaw_dbs`, skipping `-wal`/`-shm` sidecars and the `incognito-` sentinel. `parse_openclaw_sqlite` opens the store read-only with a short busy timeout, probes `sqlite_master` for `transcript_events`, and streams a single query that prefilters in SQLite (`instr(event_json, '"usage"')` and the two model-bookkeeping event types) so user prompts and tool results never leave the database or get JSON-decoded. It joins `session_windows` (falling back to the older `sessions` table, then to no join) for a provider/model fallback and uses the row's `created_at` when a message has no timestamp. Missing tables, malformed rows, and a locked store degrade to warnings, never a failed scan; a read that stops partway (a locked or damaged store) is reported for that scan but not cached, so it cannot become a persistent undercount.

**One interpretation of the event format.** The JSONL parser and the SQLite parser now share `ingest_openclaw_entry`, so `model_change` / `model-snapshot` state, assistant-only filtering, `cost.total`, and the new `reasoningTokens` split (a subset of `output`, moved into the reasoning bucket the same way the Codex parser treats `reasoning_output_tokens`) behave identically in both stores. Both lanes run SQLite first, then JSONL, and dedup on a key built from the event itself (`openclaw:<event id>:<timestamp>:<input>:<output>`), so a transcript `openclaw doctor --fix` imported into SQLite while leaving the JSONL original in place counts once, and so do the copies `/fork` writes under a new session id. Events without their own timestamp are keyed by id and usage alone (`openclaw:<event id>:<input>:<output>`), so their `/fork` copies collapse too. The message cache uses the existing WAL-aware SQLite fingerprint, so rows committed only to the WAL invalidate a cached entry.

**Codex-harness turns.** OpenClaw mirrors each turn it runs through Codex app-server as one assistant row carrying only the *last* model response's usage (`extensions/codex/src/app-server/event-projector-result.ts`), stamped with `idempotencyKey: codex-app-server:<threadId>:<turnId>:assistant`. The complete per-response record is Codex's own rollout, which by default lives in the agent's `agent/codex-home/sessions/` (`appServer.homeScope: "agent"`) where no Codex scanner looks. The OpenClaw lane now reads those rollouts (they come along with the agents-root walk and are classified by path), attributes every response in them to `openclaw` under the OpenClaw session that mirrored the thread, and drops the mirror row of each turn the rollout holds. The match is per turn: the mirror's `idempotencyKey` names the `turnId`, the rollout's `task_started` / `turn_context` carry the `turn_id`, and the parser records which turns a rollout emitted usage under (`CodexTurnCoverage`, kept in the cached parse state). A mirror row for a turn the rollout lacks, because the rollout was cut short or the turn ran after it was read, stays; a rollout written before Codex stamped turn ids can only be matched thread-wide and is. When the Codex roots overlap an agent's codex-home (`CODEX_HOME` pointed at it), the scanner lists the rollout for openclaw only, so ownership is decided once per file. Rollouts OpenClaw creates in a shared user Codex home (`homeScope: "user"`, or a supervision branch) carry `session_meta.originator: "openclaw"`, which Codex app-server stamps from the `initialize` handshake's `clientInfo.name`; the Codex parser tags those `openclaw` and the codex lane hands them over instead of counting them. A thread the codex lane counted under `codex` (a user-created thread OpenClaw resumed through supervision) suppresses the mirror rows of the turns it holds so nothing is counted twice; a thread counts under codex only when its messages pass the client filter, so `--client openclaw` alone keeps the mirror rows and `--client synthetic` still keeps Codex usage a synthetic gateway served. A mirror row whose rollout cannot be found anywhere is kept as is.

**Smaller fixes found on real data.** OpenClaw's own bookkeeping assistant rows (`api: "openclaw-transcript"`, model `delivery-mirror` / `gateway-injected`, zero usage) are skipped; they were previously counted as zero-token messages. The `*.jsonl*` matcher now accepts every `<id>.jsonl.<suffix>` copy OpenClaw makes (doctor backups `*.jsonl.pre-doctor-*.bak`, quarantined `*.jsonl.broken-*`) except the Codex plugin's `*.jsonl.codex-app-server.json*` binding sidecars, which are JSON metadata. On the test machine about 200 assistant rows existed only in such backups.

**Cache.** OpenClaw `parser_version` 1 -> 2 (messages gained dedup keys and the reasoning split) and Codex `parser_version` 7 -> 8 (rollouts OpenClaw originated are now tagged `openclaw`). Both mean one cold reparse of those lanes for existing users; no `CACHE_FORMAT_VERSION` change, so no other client's cache is touched.

## Reading order

1. `crates/tokscale-core/src/sessions/openclaw.rs`: shared ingest, `parse_openclaw_sqlite`, dedup keys, path classification.
2. `crates/tokscale-core/src/scanner.rs`: `discover_openclaw_agent_dbs`, `openclaw_dbs`, the `*.jsonl*` matcher.
3. `crates/tokscale-core/src/sessions/codex.rs`: `originator` handling and `thread_id_from_rollout_path`.
4. `crates/tokscale-core/src/lib.rs`: the OpenClaw lane in `parse_all_messages_streaming` and the mirror of it in `parse_local_clients`.
5. `crates/tokscale-core/src/sessions/utils.rs` (`CamelUsage::to_breakdown_with_reasoning`), `message_cache.rs` (versions), `crates/tokscale-cli/src/main.rs` (`tokscale clients` lists the stores), READMEs.

## Validation

- `cargo fmt --all -- --check`, `cargo clippy --locked --workspace --all-features -- -D warnings`, `cargo test --workspace --all-features` all pass; 46 new tests cover SQLite-only usage, multi-agent discovery, usage field preservation (reasoning, cache), user/tool-result exclusion, malformed JSON and missing/old schemas, legacy JSONL regression, SQLite+JSONL and fork dedup, WAL-only commits invalidating the cache, an exclusively locked store and a store cut off mid-table (reported, not cached), provider/model fallback order, transcript-bookkeeping rows, and every Codex-harness attribution rule (codex-home rollouts, `originator: openclaw` in the user home, codex-counted threads, mirror rows without a rollout, per-turn replacement including a rollout cut short and an incremental resume, rollouts without turn ids, overlapping Codex roots, codex-only and synthetic-only filters).
- Synthetic 1.2 GB store (3k sessions, 120k assistant events among 480k rows): totals match an independent `sqlite3` JSON aggregation exactly; cold scan 9 s, warm 0.25 s.
- Real OpenClaw 2026.8.2 install (4 agents, schema v19, 3,447 Codex rollouts all carrying `turn_id`, WAL active, gateway running), read-only, cross-compiled binary vs. released 4.15.0 and vs. an independent read-only aggregation written from the file formats (SQLite rows, every `*.jsonl*` under the agents root including the unreferenced legacy transcripts `openclaw doctor` parks in `session-sqlite-import-archive/`, and the rollouts with their own `token_count` snapshot rule):

| | 4.15.0 | this PR | independent read-only aggregation |
|---|---|---|---|
| OpenClaw messages, all time | 5,908 | 31,953 | 31,953 |
| OpenClaw reasoning tokens | 0 | 2,087,774 | 2,087,774 |
| OpenClaw input / output / cache-read tokens | — | 171,146,343 / 4,778,382 / 1,560,077,244 | same |

  The two agree exactly, per session (0 of 4,243 sessions differ). For the current month 4.15.0 reports 0 messages and this PR 3,298. Cold scan on that 2-vCPU host is 33 s (4.15.0: 24 s; it now reads the rollouts plus the archives it used to skip); warm scans are 8.5 s.

## Notes for review

- Session ids for rollout-derived messages are the OpenClaw session that mirrored the thread when one exists, so per-session views keep grouping by OpenClaw sessions.
- The one case this cannot resolve is a rollout that lives on another machine (OpenClaw paired node); the mirror row is kept there as a lower bound.
- Cost for Codex-harness turns is estimated from pricing like every other `openclaw` row, since OpenClaw records `cost.total: 0` for them.

---
## Summary by cubic
Fixes OpenClaw usage reporting for OpenClaw 2026.x, which moved live transcripts into per-agent SQLite databases that the JSONL-only scanner never read, causing recent OpenClaw usage to report as zero. The scan now reads the SQLite stores and the Codex-harness rollouts, attributing that usage to the existing `openclaw` client with no changes to the frontend or submit protocol.

**Bug Fixes**
- Discovers `openclaw-agent.sqlite` stores under every OpenClaw agents root and reads them read-only alongside the legacy JSONL files.
- Attributes turns OpenClaw runs through the Codex app-server harness to `openclaw` by reading the full Codex rollouts and dropping the per-turn mirror rows that carried only the last response's usage; mirror rows stay when the rollout is truncated, appended after being read, or lacks a clear turn id (an announced turn id stops applying once that turn emits usage, so a later id-less `turn_context` starts a new turn).
- Claims each rollout once even when a Codex root overlaps the agent's `codex-home`, and decides a forked child's ownership from its own `session_meta` metadata.
- Walks Codex roots as lookup inputs for openclaw-only requests, and only records a thread as Codex-counted when Codex was actually requested, fixing a `--client synthetic` regression.
- Dedupes on event content so SQLite rows, retained JSONL originals, their compressed `.jsonl.zst` archives, doctor backups, and `/fork` copies count once.
- Excludes only `.json` / `.json.migrated` suffixes from the JSONL matcher, never caches incomplete SQLite scans, and decodes compressed `.jsonl.zst` archives folded in from the compressed-archive fix that landed first.
- Skips OpenClaw's own bookkeeping assistant rows that previously counted as zero-token messages.

**Migration**
- OpenClaw parser version bumps 1→3 and Codex 7→8, so existing caches get one cold reparse for those clients; no other client's cache changes.

<sup>Written for commit 12d583426867b1fd06c8dd029faf520b87e3b099. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


PJunhyuk:
Pushed a follow-up commit addressing CI and the review comments:

- **Windows test failure**: `test_scan_all_clients_discovers_openclaw_agent_sqlite_databases` compared a `/`-joined substring; it now checks path components.
- **openclaw without codex** (`lib.rs`): the scanner now walks the Codex roots as lookup inputs for an OpenClaw-only request (same shape as the Devin Desktop → Devin CLI lookup), so `originator: "openclaw"` rollouts in a shared user Codex home still reach the openclaw lane. Codex usage is counted only when `codex` was requested, so a thread is never marked as counted under codex while its rows are filtered out. Covered by `test_scan_all_clients_openclaw_request_scans_codex_roots_as_lookup` and the openclaw-only assertions added to `test_openclaw_originated_rollouts_in_the_user_codex_home_are_openclaw_usage`.
- **Forked child ownership** (`codex.rs`): the originator is read from the file's own `session_meta` inside the active-metadata block, after the fork replay gate, so a replayed parent `session_meta` cannot claim the child. Covered by `test_replayed_parent_originator_does_not_claim_a_forked_child` in both directions.
- **`*.jsonl*` matcher** (`scanner.rs`): only suffixes ending in `.json` or `.json.migrated` are excluded; a backup whose repair name mentions json is kept. Test updated.
- **`tokscale clients`** (`main.rs`): the SQLite stores are listed from every OpenClaw root the scanner ingests (default, legacy rebrands, `extraScanPaths.openclaw`, `TOKSCALE_EXTRA_DIRS`), deduplicated by canonical path.

junhoyeo (changes requested):
This is careful work — the problem statement is backed by real measurements, the 35 new tests are real (I ran the full suite: **3546 passed, 0 failed**, with `cargo fmt`, `cargo clippy --locked --workspace --all-features` and the same with `--all-targets` all clean), and several things I went looking for were already handled well:

- `parser_version` Codex 7 → 8 does not collide; main is still on 7.
- The ownership logic is duplicated in `parse_all_messages_streaming` and `parse_local_clients`, and I diffed the two regions — they are genuinely equivalent, not drifted.
- Mirror rows are dropped with `continue` *before* entering the dedup set, so a rollout message can never be falsely deduped against a mirror.
- `originator` handling is thorough: case-insensitive, other originators stay `codex`, and `test_replayed_parent_originator_does_not_claim_a_forked_child` covers the fork-replay case.

I then ran an adversarial Codex review over the branch diff. It returned **"Do not ship: rollout ownership can drop or double-count usage, and interrupted SQLite scans can become persistent undercounts."** I verified all four findings against the code myself before posting; each one checks out.

### 1. Thread-level `recorded_threads` drops uncovered turns — `lib.rs:2470-2474`

`recorded_threads` is built as `rollout_messages.iter().map(|m| m.session_id).collect()`. That is thread granularity with no turn-coverage check, and the comment above it ("Threads with a complete record somewhere") states the property rather than establishing it. If a rollout is truncated or copied and holds only T1 while SQLite has mirrors for T1 and T2, the whole thread is marked recorded and **both** mirrors are dropped — T2's usage is lost. A concurrent append between the rollout read and the SQLite read produces the same gap. Both lanes share the rule.

Suggested: track coverage by (thread, turn id) and suppress only mirrors whose turn is actually represented.

### 2. Overlapping Codex roots double-count — `lib.rs:2451-2459`

```rust
if message.client != sessions::codex::OPENCLAW_CLIENT_ID {
    message.client = sessions::codex::OPENCLAW_CLIENT_ID.to_string();
```

This retags by *location*, unconditionally. Point `CODEX_HOME` at an agent's `agent/codex-home` with a rollout whose originator is missing or non-OpenClaw: the codex lane emits it as `codex`, this lane reparses and emits it as `openclaw`, and the two lanes keep separate dedup sets (`codex_seen` vs `openclaw_seen`), so the same usage survives under both clients. I had started down this path myself and could not confirm it; Codex closed it with the concrete recipe.

Suggested: decide ownership once per canonical rollout path, before either lane emits.

### 3. Interrupted SQLite scans cached as complete — `openclaw.rs:585-590`

```rust
SqliteScan::Ran | SqliteScan::Incomplete | SqliteScan::NotExecuted => break,
```

`SqliteScan::Incomplete` exists precisely to say "the rows handed to the sink are a prefix of the result, not the result" — the enum's own doc says so — and this call site collapses that distinction into the success arm. The prefix is returned as an ordinary message vector, the cache adapter stores any nonempty result under the DB/WAL fingerprint, and warm scans then replay the incomplete usage until the source file changes. A step error partway through becomes a durable undercount. The first two queries also pass no diagnostic label, so the underlying error is not surfaced.

### 4. Synthetic-only requests lose gateway usage — `lib.rs:2097 / 2114`

`include_codex = include_all || clients.iter().any(|c| c == "codex")`. For `clients = ["synthetic"]` that is `false`, so `else if include_codex && …` drops Codex messages at 2114 — before `retain_for_requested_clients` (187-201) can apply the synthetic rule that would have kept qualifying gateway traffic. The uncached lane repeats it at 5171/5179.

---

None of this touches the core contribution, which is clearly right: OpenClaw 2026.x moved transcripts into SQLite and we were reporting zero. It is the ownership and completeness edges around it. Happy to take another pass whenever you have pushed.

Two smaller things, not blocking:

- An event with no timestamp of its own falls back to `openclaw:<session id>:<event id>`. Since `/fork` copies land under a *new* session id, those events look distinct across the fork and can be counted twice — the inverse of the case the content-keyed scheme was designed to collapse.
- The PR reports 30,677 messages against an independent aggregation of 30,642 (and reasoning tokens 2,018,637 vs 2,018,480). The description attributes it to collapsing repeated `token_count` snapshots, which is plausible, but it is an over-count of about 0.1% and over-counting is the worse direction for a leaderboard. Worth pinning down before this lands.

PJunhyuk:
Pushed 4fc31d5 addressing the four findings and the first of the smaller points; the second smaller point is settled by measurement below.

### 1. Thread-level `recorded_threads` → per-turn coverage
The codex parser now records which turns a rollout emitted usage under (`CodexTurnCoverage` on `CodexParseState`, from the `turn_id` on `task_started` / `turn_context`; it lives in the cached parse state, so a warm hit and an incremental resume carry it). The mirror row's dedup key carries the turn: `openclaw:codex-mirror:<thread>:<turn>:…`, from `idempotencyKey: codex-app-server:<thread>:<turn>:assistant`. Both lanes replace a mirror row only when a rollout read in that scan holds that turn (`RecordedCodexTurns::covers`); a truncated rollout, or a turn appended after the read, leaves its mirror row in place. A rollout written before Codex stamped turn ids has no per-turn signal and still stands in for its whole thread — that is all it allows, and it is documented and tested as the fallback. Tests: `test_openclaw_rollout_stands_in_only_for_the_turns_it_holds` (cold, warm, incremental resume after the missing turn is appended, and the uncached lane), `test_openclaw_rollout_without_turn_ids_stands_in_for_its_whole_thread`, `test_turn_coverage_records_the_turns_usage_was_emitted_under`.

### 2. Overlapping Codex roots
Ownership is decided once, in the scanner: a file the OpenClaw scan classifies as a codex-home rollout is removed from the Codex list by canonical path, so with `CODEX_HOME` pointed at an agent's codex-home the codex lane never sees it and only the openclaw lane emits it. Without openclaw in the request the agents tree is not walked and the file stays Codex's by its own metadata. Tests: `test_scan_all_clients_openclaw_codex_home_rollout_is_listed_under_openclaw_only`, `test_openclaw_codex_home_rollout_counts_once_when_codex_home_points_at_it` (your recipe: originator `codex-tui`, both clients requested; cached cold and warm, and uncached).

### 3. Interrupted SQLite scans
`parse_openclaw_sqlite` is now a wrapper over `scan_openclaw_sqlite`, which returns `{ messages, complete }`; `complete` is false for `Incomplete`, `NotExecuted`, `NotPrepared`, an open failure, and a probe that could not run (a locked store). The cached lane goes through `load_or_parse_source_with_fingerprint_and_policy` with that flag as `cacheable`, so an incomplete read is reported for the scan but never stored under the fingerprint, and a stale entry is invalidated. The schema fallback is a silent prepare-only probe (`session_windows` → `sessions` → bare) followed by one labeled query, so a step error is logged with its cause. Tests: `test_scan_openclaw_sqlite_reports_a_store_it_could_not_read_to_the_end`, the locked-store test now asserts `!complete`, and `test_openclaw_store_read_partway_is_reported_but_not_cached` (no cache entry after a cut-off store, one after an intact store).

### 4. Synthetic-only requests
The `include_codex` gate is gone. Codex messages always reach the lane buffer and the flush filter decides, as on main; in `parse_local_clients` the same filter is applied at the push so the per-client count agrees with what is returned. The only difference from main is when a thread is marked as counted under codex: when at least one of its messages passes `retain_for_requested_clients`. Test: `test_synthetic_request_keeps_codex_usage_a_synthetic_gateway_served` (both lanes; a `model_provider: synthetic` rollout is kept under `--client synthetic`, an `openai` one is not).

### Smaller points
- Timestamp-less events are keyed `openclaw:<event id>:<input>:<output>` now — no store-filled timestamp and no session id, so a `/fork` copy collapses onto its original (`test_fork_copies_share_a_dedup_key_across_sessions` updated; the store's own timestamp still reaches the message).
- The 30,677 vs 30,642 gap. I rebuilt the independent aggregation from the file formats alone (SQLite rows, every `*.jsonl*` under the agents root, rollouts with their own snapshot rule for `token_count`) and ran it against the same host as this branch (OpenClaw 2026.8.2, 4 agents, 3,447 rollouts, every one carrying `turn_id`). With this commit the two agree exactly, per session: 31,953 messages, input 171,146,343, output 4,778,382, cache read 1,560,077,244, reasoning 2,087,774, zero sessions differing out of 4,243. The earlier gap was on the independent side, not in the parser: it had not read `session-sqlite-import-archive/`, where `openclaw doctor` parks unreferenced legacy transcripts as `archive-tier.<id>.jsonl.imported-<ts>` without importing them into SQLite (1,220 files on this host), so it missed part of what tokscale counts. The total is higher than in the description because the host has two more days of usage and because per-turn coverage keeps the mirror rows whose turns no rollout holds (95 here). Cold scan 33 s, warm 8.5 s on the 2-vCPU host.

Cache: no `CACHE_FORMAT_VERSION` change. The Codex parse-state layout and the OpenClaw mirror key changed within this PR under the versions it already bumps (Codex 8, OpenClaw 2), so a user coming from main takes the same single reparse; only a cache written by an earlier commit of this branch is rebuilt.

Two things I noticed on the real data and left alone as out of scope: tokscale reports the archive-tier transcripts under the session id `archive-tier.<id>` (the archived filename) rather than `<id>`, and `~/.openclaw/agents/main/agent/cli-auth/codex/<profile>/sessions/` holds a handful of `codex_exec` rollouts (April, about 50 responses) that no lane attributes, since only `agent/codex-home` is recognized as a Codex home. Happy to follow up on either.

Full suite: 3,513 passed, 0 failed; `cargo fmt --all -- --check` and `cargo clippy --locked --workspace --all-features` with and without `--all-targets` clean.

PJunhyuk:
Pushed 8a0eb26 for the cubic round on 4fc31d5. Three of the five are taken; two are kept as they are, with the reasoning.

- **codex.rs:454 (turn without an id after identified turns)** — taken. `task_started` and `turn_context` now each replace the current turn id, with `None` when they carry none, so a turn resumed under an older Codex is recorded as `without_turn_id` rather than under the previous turn's id. The one exception is a `turn_context` without an id right after its own `task_started`, which keeps the announced id (`turn_id_announced_by_task_started` on the parse state). Both cases are added to `test_turn_coverage_records_the_turns_usage_was_emitted_under`.
- **openclaw.rs:674 (probe passes, query fails)** — taken, as suggested: the two join candidates are prepared in full and the first that prepares runs, so a partial schema falls through to the next candidate instead of failing the read.
- **scanner.rs:5949 (`CODEX_HOME` restore on panic)** — taken; the test uses `EnvGuard`.
- **lib.rs:852 (mixed coverage marks the whole thread)** — kept, deliberately; the comment on `RecordedCodexTurns::record` now says why. A rollout that mixes id-less and identified turns holds its id-less turns, so their mirror rows should yield, and only a thread-wide match can make them yield. Matching per turn would keep every id-less turn's mirror row beside the rollout's own record of it, a double count that is permanent. What thread-wide gives up is the guard for a turn the rollout lacks; a turn that ran after the read is picked up by the next scan, so that gap closes on its own. On the host I measured, every one of the 3,447 rollouts carries `turn_id`, so neither branch is exercised there.
- **openclaw.rs:100 (timestamp-less key without session scope)** — kept. The key without the session is what lets a `/fork` copy of a timestamp-less event collapse onto its original, which the earlier review asked for; a session-scoped key counts every such copy again. The collision the comment describes needs two unrelated events with the same event id *and* the same input and output counts. On the real store the ids are 36-character UUIDs for 19,821 of 22,797 assistant rows and 8-character hex for the older 2,976, and the number of assistant rows without a timestamp is zero, so this branch of the key is not reached on current data at all.

Core library tests: 2,090 passed; `cargo fmt` and `cargo clippy --locked --workspace --all-features --all-targets` clean.

PJunhyuk:
Pushed 49a922b1 for the cubic finding on 8a0eb26 (codex.rs:467): the announcement flag is now cleared when the announced turn emits usage, so a later `turn_context` without an id starts a turn of its own, recorded as `without_turn_id`, instead of reusing the announced id. Case added to `test_turn_coverage_records_the_turns_usage_was_emitted_under` (task_started with usage, no turn_context, then an id-less turn_context with usage → coverage {turn-a, turn-d} plus `without_turn_id`). Core library tests 2,090 passed; fmt and clippy `--all-targets` clean.

junhoyeo (approved):
Re-reviewed at `49a922b1`. All four blocking findings are closed, and I verified each against the code rather than taking the commit messages for it.

**1. Turn-level coverage** — `RecordedCodexTurns` (lib.rs:838) replaces the thread-level `HashSet`, with `turns: HashMap<thread, HashSet<turn_id>>` and `covers(thread, turn)`. A rollout now stands in only for the turns it actually holds.

**2. Ownership decided once** — `test_openclaw_codex_home_rollout_counts_once_when_codex_home_points_at_it` (lib.rs:11953) builds exactly the overlap case: `CODEX_HOME` aimed at the agent's `codex-home`, rollout originator not naming OpenClaw. It asserts `Codex = 0`, `OpenClaw = 2`, runs the parse twice for cold and warm cache, and covers the codex-only request separately. This was the vector I could not close on my own pass.

**3. Scan completeness** — `SqliteScan::Incomplete` now maps to `complete: false` (openclaw.rs:655-662) and rides through `load_or_parse_source_with_fingerprint_and_policy` as `(scan.messages, scan.complete)`, so a prefix no longer gets stored under the DB/WAL fingerprint. The schema probe change (prepare each join candidate in full, fall through to the next) is a good call on its own.

**4. Synthetic filtering** — `include_codex` is gone; the lane calls `retain_for_requested_clients` directly, which is the central synthetic-aware predicate. Both lanes match.

Both non-blocking items from my last pass are also handled. The timestamp-less key dropping the session scope closes the fork double-count I raised, and the validation table now agrees exactly rather than by 0.1%: **31,953 = 31,953** messages, **2,087,774 = 2,087,774** reasoning tokens, 0 of 4,243 sessions differing.

Gates on my machine at this head: `cargo fmt`, `cargo clippy --locked --workspace --all-features`, the same with `--all-targets`, and `cargo test --workspace --all-features --no-fail-fast` — **3,555 passed, 0 failed**.

Approving. Two notes, neither blocking:

**A third option for the mixed-coverage thread (lib.rs:861).** You and cubic are each right about a different risk: per-turn matching on a mixed rollout leaves the id-less turns' mirrors beside the rollout's own record (permanent double count), while thread-wide drops mirrors for turns the rollout lacks. Both are avoidable, because `CodexTurnCoverage` already distinguishes them and mirror keys carry their turn (`codex_mirror_turn_from_dedup_key`). Splitting `whole_threads` in two would do it:

```rust
fn record(&mut self, thread: &str, coverage: &CodexTurnCoverage) {
    if coverage.turn_ids.is_empty() {
        self.whole_threads.insert(thread.to_string());   // no ids at all
    } else {
        if coverage.without_turn_id {
            self.idless_threads.insert(thread.to_string());  // mixed
        }
        self.turns.entry(thread.to_string()).or_default()
            .extend(coverage.turn_ids.iter().cloned());
    }
}
```

with `covers(thread, None)` consulting `idless_threads | whole_threads` and `covers(thread, Some(t))` consulting `whole_threads | turns[thread]`. Id-less mirrors still yield, identified turns the rollout lacks keep their mirror. Your "the next scan picks it up" argument holds for a concurrent append but not for a permanently truncated or copied rollout, which is the case I originally raised — though with all 3,447 rollouts on your host carrying `turn_id`, none of this is reachable on current data, so it is your call whether it is worth the third set.

**Commit subjects.** The last three are Korean; the last 300 on `main` are all English. Squash-merge takes the PR title so this mostly does not reach `main`'s history, just flagging the convention.

junhoyeo:
Cross-referencing a conflict with #1285 (compressed OpenClaw archives), which landed as a draft after my approval here.

On `main`, OpenClaw has no explicit arm in `parser_version` and falls through to `_ => 1`. Both this PR and #1285 add `ClientId::OpenClaw => 2`, for unrelated reasons — here it is the dedup keys and the reasoning split, there it is discarding archive entries that were cached empty.

Whichever lands first takes users from 1 to 2 and gets its reparse. When the second lands, the number is already 2, so nothing is invalidated and the second PR's fix never reaches anyone with a warm cache. Each PR's own v1→v2 test passes on its own branch, so CI cannot see it. The repo has hit this before with Copilot at 8.

If this merges first — which it should, it is approved and much further along — then #1285 needs `=> 3`. I have left the same note there. The two branches also conflict today, three hunks across `scanner.rs` and `openclaw.rs`, so #1285 will need a rebase regardless.

Worth knowing in the other direction too: #1285 has a double-count where a plain `S.jsonl` and its `S.jsonl.zst` copy are both parsed, which exists precisely because OpenClaw messages carry no dedup key on `main`. The content-keyed dedup this PR introduces collapses that pair on its own, so merging this first makes that finding largely disappear rather than needing its own fix.

No action needed here — this PR stands as approved.

PJunhyuk:
Thanks for the re-review and the approval.

**Mixed coverage.** I looked at the third option and want to be precise about why I am leaving `whole_threads` as it is rather than adding `idless_threads`. The turn id on a mirror row does not come from the rollout file: OpenClaw's Codex extension stamps `idempotencyKey: codex-app-server:<thread>:<turn>:assistant` from the app-server event stream (`turn/started`), so a mirror row carries its turn id whether or not the rollout's `turn_context` was written with one. That means the `covers(thread, None)` path the split would add is essentially never reached; the mirror of an id-less rollout turn arrives as `Some(t)` with a `t` that is not in `turns[thread]`, which the split would keep and count beside the rollout's own record. So the choice stays the same two-way trade-off — an id-less turn the rollout does hold versus a turn a truncated copy lacks — and either rule gets one of them wrong. I kept the one whose failure is transient for a live rollout, and the reasoning is on `record` for whoever reaches this case with real data. Agreed that with every rollout on the host carrying `turn_id` neither branch is reachable today, so I would rather not add a third set to an approved diff for it.

**Commit subjects.** Noted; those three subjects came from my own convention and I will keep future subjects in English here. Since the squash takes the PR title, I have left the branch as is rather than force-pushing over an approved head.

**#1285.** Understood on the `parser_version` collision; nothing to do on this side, and I will keep an eye on the rebase there if it helps.

junhoyeo:
#1285 landed first, so I rebased this branch onto it — merge commit `6fe35347`, pushed here. Three things worth knowing about what that merge contained.

**`parser_version` is now 3, not 2.** Both branches had added `ClientId::OpenClaw => 2`. Git auto-merged `message_cache.rs` by keeping *both* arms, which clippy rejected as unreachable — the only thing that stopped it landing silently. The real issue is that the numbers collided: #1285 took v2 for the archive decode, so a warm v2 cache written by it has neither the dedup keys nor the reasoning split this branch adds. Reusing 2 would have left every reader who already scanned under #1285 unmigrated. The arm comment now records both steps, and `test_openclaw_parser_version_invalidates_keyless_v2_entries` asserts 3.

**Two genuine conflicts, resolved as the union.** `scanner.rs`: both sides rewrote the `*.jsonl*` matcher; yours is the general one and already admits `.jsonl.zst`, so it stands. `openclaw.rs`: #1285 routes lines through `for_each_transcript_line` so archives get decoded, and this branch parses each line through the shared `parse_json_line` helper — resolved as the archive-aware reader feeding the shared parser.

**This branch closes #1285's double-count.** On `main` right now a plain `S.jsonl` and its `S.jsonl.zst` copy are both scanned with no dedup key, so the same events count twice. I checked the rebased branch directly: a plain transcript and its zstd copy with a top-level `id` produce the identical key (`openclaw:msg1:1788566869012:100:50`) and the lane keeps one. That is a property of your dedup design, not anything I added — noting it because #1291 was opened this morning to fix the same thing independently and would collide with this branch on all three files and on `=> 3`.

Gates on the merge: fmt, clippy with and without `--all-targets`, full workspace suite — 3,582 passed, 0 failed. Approval stands.

junhoyeo:
One more maintainer commit, `12d58342`: the aggregate-path regression from #1291, carried over as agreed when that PR was closed as superseded by this one. Your dedup tests covered SQLite-vs-JSONL and fork copies but none named a `.zst` file beside its plain original; this one does — five messages and 254 input tokens through cold, warm and direct, shared event keyed once, id-less events kept apart. With the id-to-key derivation disabled it reports six, so it fails for the right reason. No production code changed.

PR #1280, opened 2026-09-05
fix(antigravity): stop dropping a language server slower than 400ms

junhoyeo:
Discovery probed each candidate port with a full quota request bounded at 400ms, and that bound was never only a probe bound: `has_credentials()` reports Antigravity as present exactly when discovery succeeds, and `usage/mod.rs` drops a provider from the report before any fetch runs when it does not. A language server that answers a loopback round trip in more than 400ms — which a loaded machine does routinely — was missing from `tokscale usage` altogether, against a server that was answering correctly.

### What this replaces

The first version of this branch gave `fetch_all` its own 10s budget and left discovery at 400ms. That was a no-op — `fetch_all` reaches `call_rpc` only through `discover_port`, so the provider never got that far — and the test it shipped called `call_rpc` directly and could not observe the gate. Codex's adversarial review caught it; this is the rewrite.

### Design

- **Discovery keeps what it fetched.** The probe *is* the quota request, so the summary that proves a port good is the summary `fetch_all` needs. It used to discard it and ask the same port the same question again. That removes the second request, and with it the question of what to bound it with.
- **Candidates are asked together under one deadline per source**, not one after another under a budget each. A port with nothing behind it refuses outright, so a machine without Antigravity never approaches the deadline; a port that accepts and then goes quiet is waited on once per round rather than once per candidate. Each source's deadline is computed when its own round starts — a single shared deadline let a stalled logged port consume all of it before process detection was ever asked.
- **A round returns the best-ranked answer, not the first to arrive.** `ports_from_cli_log` orders candidates newest first because the last port the log names is the current one. Resolving by completion order threw that away: a stale `agy` left over from a restart could win on scheduling noise, and a signed-out one (an empty group list parses fine) turned into an intermittent "running but not signed in" for a signed-in user. There is no grace period shorter than the round — a round waits for a better-ranked candidate exactly as long as it waits for any candidate, and returns the moment nothing ranked above its answer can still arrive.
- **No trust store, no eager client.** A round with no candidates returns before building anything, and the client a round does build asks for no trust store (it only posts to `http://127.0.0.1`, refuses redirects, has nothing to verify). `has_credentials()` on a machine without Antigravity: 50–54ms on main, 187–313ms on the previous revision, 52–56ms now.

The runtime isolation from #1269 (`off_caller_runtime`) is kept.

### Tests

Every round-level proof runs under a budget the test owns (30s stall guard), never against the production 5s as a ceiling. `a_better_ranked_candidate_is_waited_for_the_whole_round` delays the newer server 7s — past the production round by 2s, pinned by a compile-time assert so it cannot be lowered back under 5s and silently stop proving anything — and fails with `Stale Server` against both a 4s-grace mutant and a dead-`round`-parameter mutant. Client builds are counted through an injected builder into a `Cell` on the test's own stack, so no `#[cfg(test)]` hook ships in production code and the test needs no `#[serial]`.

An intermittent failure was reported against `a_round_with_no_candidates_builds_no_client` at an earlier revision. Every archived failure came from two short windows in which every run failed and none outside them; the identical panic is produced by an eager-build mutation that was being run in that worktree at the time, and the committed code produces neither signature. Under load: 40/40, 45/45, 51/51 across three revisions at loadavg 100–260.

### Provenance

Found while reviewing #1269. Independent of #1290, which edits `https_rpc_request` in the same file (comment-only overlap).

### DIFF
--- a/crates/tokscale-cli/src/antigravity.rs
+++ b/crates/tokscale-cli/src/antigravity.rs
@@ -2208,15 +2208,15 @@ fn https_rpc_request(
         // context entered (#1264): the TUI usage refresh and `tokscale
         // usage` reach this function while `has_credentials`'s own
         // current-thread runtime is mid-`block_on`, because
-        // `discover_port` probes the IDE's language server
-        // synchronously. A dedicated OS thread never carries a runtime
-        // context, so the nested-runtime panic is structurally
-        // impossible for every caller; the scope (not
-        // `std::thread::spawn`) is what lets the request keep borrowing
-        // `connection`/`method`/`body`. The ~microsecond spawn cost is
-        // irrelevant next to a loopback RPC with a 10s timeout, and a
-        // worker panic now degrades to `Err` instead of unwinding into
-        // whatever thread called us.
+        // `discover_quota` reaches the IDE's language server through
+        // `detect_antigravity_connections`, which is synchronous. A
+        // dedicated OS thread never carries a runtime context, so the
+        // nested-runtime panic is structurally impossible for every
+        // caller; the scope (not `std::thread::spawn`) is what lets the
+        // request keep borrowing `connection`/`method`/`body`. The
+        // ~microsecond spawn cost is irrelevant next to a loopback RPC
+        // with a 10s timeout, and a worker panic now degrades to `Err`
+        // instead of unwinding into whatever thread called us.
         let joined: std::result::Result<Result<Value>, Box<dyn std::any::Any + Send>> =
             std::thread::scope(|scope| {
                 scope
--- a/crates/tokscale-cli/src/commands/usage/antigravity.rs
+++ b/crates/tokscale-cli/src/commands/usage/antigravity.rs
@@ -73,25 +73,24 @@ const RPC_PATH: &str = "/exa.language_server_pb.LanguageServerService/RetrieveUs
 /// rounds sits well inside the 12s-30s the cloud-backed providers here allow
 /// themselves, and leaves a healthy loopback answer -- milliseconds when idle,
 /// low seconds on a loaded machine -- several times the room it needs.
+///
+/// This is also how long a round waits for a *better-ranked* candidate once it
+/// holds an answer from a worse one. [`ports_from_cli_log`] yields candidates
+/// newest first, because the last port the log names is the current server,
+/// and an older entry can still be answered by a stale `agy` left over from a
+/// restart or an account switch, whose quota is another account's or nobody's
+/// (signed out, which parses fine and reports as "not signed in"). So a round
+/// keeps the best-ranked answer it has seen and returns the moment nothing
+/// better can still arrive. What it is waiting on until then is the current
+/// server -- the server this budget was sized for -- so there is no shorter
+/// grace for that wait. An earlier revision cut it at 2s from the first answer,
+/// which contradicted the premise above: a current server answering in 2-5s on
+/// a loaded machine lost to a stale sibling that answered at once. The cost of
+/// a single bound is confined to a newest logged port that accepts and then
+/// goes quiet while an older one still answers: that round runs to its
+/// deadline before believing the older answer, which is the worst case one
+/// stalled candidate already costs.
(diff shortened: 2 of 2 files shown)


junhoyeo:
Converting to draft — @chatgpt-codex-connector's P1 is correct and it invalidates this PR's headline claim.

`discover_port()` still probes every candidate at `PROBE_TIMEOUT`, and both `has_credentials()` and `fetch_all()` go through it. So in exactly the case this PR targets — a loaded machine where the quota RPC takes longer than 400ms — Antigravity is filtered out as "not running" before `call_rpc(..., FETCH_TIMEOUT)` is ever reached. The split is a no-op for the scenario it was written for.

My test bypassed `discover_port()` entirely, which is the gap the commit's own `Not-tested:` trailer already admitted. It pinned that `call_rpc` honours the budget it is handed; it could not see that nothing hands it the bigger one in a real run.

Rebased onto main (resolving the conflict with #1269, which wrapped `fetch_all` in `off_caller_runtime`) so this is ready to pick up, but it needs a design decision rather than a patch:

A wrong port normally **refuses** the connection, which fails instantly. `PROBE_TIMEOUT` exists only for the rarer accept-then-go-quiet case. So the options are roughly:

1. **Two-pass discovery** — probe all candidates at `PROBE_TIMEOUT`, and if every one fails, retry them once at `FETCH_TIMEOUT`. Cheap when nothing is listening (stale ports refuse instantly), but worst case is N x 10s against stale ports owned by other processes that accept and stall — and usage providers share a fan-out, so that stalls the whole `tokscale usage` report.
2. **Escalate only on connect** — treat "TCP connect succeeded but the body was slow" as worth a longer wait, and keep refusing fast. More precise, more code.
3. **Leave discovery alone** and accept that a machine loaded enough to blow 400ms just does not report Antigravity usage that run.

Holding until that is settled rather than merging a confident commit message over a no-op.

junhoyeo:
Ran `/codex:adversarial-review` against this branch. It confirms the hold and points at a better repair than the three options I listed above.

> **Verdict: needs-attention** — Do not ship as a loaded-machine fix: the 400 ms discovery gate still prevents slow servers from reaching the new budget.
>
> **[high]** The longer fetch timeout sits behind the same failing quota request (`crates/tokscale-cli/src/commands/usage/antigravity.rs:289-296`). Both discovery loops require the complete quota RPC to succeed within 400 ms. A healthy server consistently responding in 1.5 seconds therefore fails `has_credentials()`, and `usage/mod.rs` filters Antigravity out before fetching. Calling `fetch_all()` directly repeats discovery and reports "not running." The split only helps when earlier probes succeed quickly and the final request slows down. This gate predates the branch, but leaves its stated failure unresolved. The new test bypasses discovery by handing the rejected port directly to `call_rpc`, so it cannot verify the intended recovery.
>
> **Recommendation:** Remove the mandatory 400 ms quota-success gate from availability and discovery. Use bounded concurrent candidate requests under an overall deadline, retain the first validated summary, and avoid fetching it again. Add a regression through provider selection and fetch with responses delayed beyond 400 ms.

The part I had missed: **discovery already obtains a complete, valid quota summary and then discards it.** `discover_port` calls `call_rpc` purely as a validity probe, and `fetch_all` then issues the identical request a second time. So the right change is not to split the timeout — it is to keep the summary discovery already has and never refetch, which removes the second request and the second budget along with it.

Also worth recording: `has_credentials()` runs the same gated discovery and `usage/mod.rs` drops Antigravity from the provider list *before* any fetch, so a consistently slow server never reaches `fetch_all` at all. Any regression test has to run through provider selection rather than calling `call_rpc` directly — which is precisely why the test on this branch proves nothing about the scenario.

Leaving this as draft. When it is picked up it should probably be rewritten around "reuse the discovery response" rather than rebased.

junhoyeo:
Follow-up: #1301 corrects the comment on the `JoinError` arm of `race_for_quota_with`, which claimed a panicked probe makes the round run to its deadline. It does not — `join_next` yields `None` once the other requests finish and the loop ends there. Comment-only; found by the adversarial verifier re-running on the merged tree.

PR #1285, opened 2026-09-06
fix(openclaw): read compressed transcript archives

### LINKED ISSUE #1284: OpenClaw usage missing after v2: zstd transcript archives parsed as plain JSONL
This reporting gap was observed after upgrading OpenClaw: recent OpenClaw usage stopped appearing in Tokscale even though the archived transcripts still contained nonzero token usage. The confirmed incompatibility is Tokscale reading OpenClaw’s zstd-compressed transcript archives as plain JSONL, silently dropping those records.

OpenClaw archives transcripts as `<session>.jsonl.deleted.<timestamp>.<nonce>.zst` (and reset variants). Tokscale 4.15.1 discovers these files through `*.jsonl*`, but `parse_openclaw_transcript` passes the compressed bytes to the plain JSONL reader. Every record is rejected, producing an empty result that is then cached. A plain `<session>.jsonl.zst` is not discovered at all.

This makes recent OpenClaw usage disappear after archival even though the archived assistant messages still contain their original model, provider, timestamp, and token usage. On an affected host, recent compressed archives contained nonzero `message.usage` while a recent OpenClaw report returned no entries.

Minimal reproduction:

1. Create `~/.openclaw/agents/main/sessions/example.jsonl` containing `{"type":"message","message":{"role":"assistant","provider":"anthropic","model":"claude-sonnet-4-6","usage":{"input":100,"output":50},"timestamp":1788566869012}}`.
2. Confirm `tokscale --no-spinner --client openclaw --json` reports 150 tokens.
3. Compress it with zstd to `example.jsonl.deleted.timestamp.nonce.zst` and remove only the synthetic uncompressed fixture.
4. Rescan: the current parser returns zero messages instead of the same 150 tokens.

The fix needs archive decoding, discovery of compressed transcript names, and a client-specific cache parser-version bump so unchanged archives previously cached as empty are reparsed. Invalid or truncated archives should not contribute fabricated partial usage, and decompression must be bounded.

### OpenClaw upgrade context

The missing usage was reported following an OpenClaw upgrade. Related current release: [OpenClaw v2026.9.2](https://github.com/openclaw/openclaw/releases/tag/v2026.9.2), published September 5, 2026. The upgrade is the reported trigger; compressed archive parsing is the reproduced failure mechanism. We have not yet established the first OpenClaw version that introduced this archive format, so this should not be read as proof that v2026.9.2 introduced compression.

shunkakinoki:
This reporting gap was observed after upgrading OpenClaw: recent OpenClaw usage stopped appearing in Tokscale even though the archived transcripts still contained nonzero token usage. The confirmed incompatibility is Tokscale reading OpenClaw’s zstd-compressed transcript archives as plain JSONL, silently dropping those records.

OpenClaw's archived `.jsonl.deleted.*.zst` and `.jsonl.reset.*.zst` transcripts were discovered but parsed as plain JSONL, silently dropping their token usage. Decode those archives with bounded zstd expansion, discover `.jsonl.zst` files, and bump only OpenClaw's parser cache version so previously empty entries rebuild.

Regression coverage compares plain and compressed usage, exercises discovery through parsing, rejects corrupt/truncated archives and excessive expansion, and proves an old empty cache entry is replaced and stays correct on a warm scan.

Validation: the new parity regression failed before the fix; all 20 OpenClaw-focused tests pass; the full core suite passes (2,058 passed, 2 ignored); `cargo clippy -p tokscale-core --lib -- -D warnings` passes.

Closes #1284.

### OpenClaw upgrade context

The missing usage was reported following an OpenClaw upgrade. Related current release: [OpenClaw v2026.9.2](https://github.com/openclaw/openclaw/releases/tag/v2026.9.2), published September 5, 2026. The upgrade is the reported trigger; compressed archive parsing is the reproduced failure mechanism. We have not yet established the first OpenClaw version that introduced this archive format, so this should not be read as proof that v2026.9.2 introduced compression.

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1066,6 +1066,9 @@ pub fn parser_generation() -> u64 {
 
 fn parser_version(client: ClientId) -> u32 {
     match client {
+        // v1->v2: compressed OpenClaw archives were scanned as plain JSONL and
+        // cached as empty. Their bytes do not change when decoding is fixed.
+        ClientId::OpenClaw => 2,
         // These clients accumulated parser-only invalidations under the old
         // global schema. Their independent counters start from those histories
         // so future changes have an obvious local version to increment.
@@ -3544,6 +3547,49 @@ mod tests {
         // its fingerprint keeps matching and only the version bump discards the
         // v1 lock-timestamp anchor.
         assert_eq!(parser_version(ClientId::Droid), 7);
+    }
+
+    #[test]
+    #[serial_test::serial]
+    fn openclaw_compressed_archives_discard_cached_empty_v1_results() {
+        let temp_home = TempDir::new().unwrap();
+        let _cache_env = sandbox_cache_env(temp_home.path());
+        let source = temp_home.path().join("session.jsonl.deleted.timestamp.zst");
+        let content = br#"{"type":"message","message":{"role":"assistant","model":"example","usage":{"input":100},"timestamp":1700000000000}}"#;
+        fs::write(&source, zstd::encode_all(&content[..], 0).unwrap()).unwrap();
+        let identity = CacheIdentity::for_client(ClientId::OpenClaw);
+        let old_identity = CacheIdentity {
+            parser_version: 1,
+            ..identity
+        };
+        let fingerprint = SourceFingerprint::from_path(&source).unwrap();
+        let entry = CachedSourceEntry::new(
+            old_identity,
+            &source,
+            fingerprint.clone(),
+            Vec::new(),
+            Vec::new(),
+            None,
+        );
+        let shard = cache_shard_path(identity, &source);
+        ensure_cache_dir(shard.parent().unwrap()).unwrap();
+        write_shard_with_limit(&shard, old_identity, &[entry], MAX_CACHE_SHARD_BYTES).unwrap();
+
+        let mut cache = SourceMessageCache::load();
+        assert!(cache.get(identity, &source).is_none());
+        let parsed = crate::sessions::openclaw::parse_openclaw_transcript(&source);
+        assert_eq!(parsed.len(), 1);
+        cache.insert(CachedSourceEntry::new(
+            identity,
+            &source,
+            fingerprint,
+            parsed.clone(),
+            Vec::new(),
+            None,
+        ));
+        cache.save_if_dirty();
+        let warm = SourceMessageCache::load();
+        assert_eq!(warm.get(identity, &source).unwrap().messages, parsed);
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


junhoyeo:
Early feedback while this is still a draft — no rush, and I have not marked it blocking.

Gates are clean at `c8758096`: `cargo fmt`, `cargo clippy --locked --workspace --all-features`, the same with `--all-targets`, and `cargo test --workspace --all-features --no-fail-fast` (**3,536 passed, 0 failed**).

The important thing is not in the diff itself — it is the interaction with #1278, which touches the same three files and is approved and close to landing.

### These two PRs silently disable each other's cache invalidation

On `main`, OpenClaw has no explicit arm in `parser_version` and falls through to `_ => 1`. Both this PR and #1278 add `ClientId::OpenClaw => 2`, for unrelated reasons:

- #1278 — messages gained dedup keys and a reasoning split
- this PR — compressed archives were scanned as plain JSONL and cached empty

Whoever lands first takes users from 1 to 2 and gets its reparse. When the second lands, the number is *already* 2, so no cache is invalidated and the second PR's fix never reaches anyone with a warm cache. If this one lands second, the empty archive entries it exists to discard stay exactly where they are.

The second to land needs `=> 3` with its own reason recorded. This has bitten the repo before (Copilot, at 8). Each PR's own v1→v2 test passes on its own branch, so nothing catches it.

Merging also conflicts today — three hunks across `scanner.rs` and `openclaw.rs`.

### Rebasing onto #1278 would fix one of the two findings for free

An adversarial Codex pass returned **needs-attention** with two high findings:

> **[high] Compressed copies double-count their uncompressed originals** (`scanner.rs:453-457`) — when `S.jsonl` and `S.jsonl.zst` coexist, both are scanned and parsed. Filename normalization gives them the same session ID, but OpenClaw messages have no dedup key, and both the cached and local pipelines append every message without deduplication, doubling tokens and costs on cold and warm scans. The added tests parse copies individually and never check their combined totals.

I checked the premise: `main`'s `openclaw.rs` contains zero `dedup_key` references, and #1278's contains 25. #1278 introduces exactly the content-keyed dedup (`openclaw:<event id>:<timestamp>:<input>:<output>`) that would collapse a plain/archive pair, since a compressed copy holds the same events. So this finding is real against `main` and mostly evaporates on top of #1278 — which is a good argument for basing this on #1278 rather than racing it.

> **[high] The archive limit does not bound JSON parsing memory** (`openclaw.rs:37-42`) — a fixture with 31,457,280 zeros in an ignored JSON array compresses to 5,806 bytes and expands to 62,914,591 bytes, passing the 64 MiB cap. Each decoded row is copied into line and parser buffers, then simd-json allocates structural indexes, tape and stack before Serde ignores the field, implying GiB-scale requests from one tiny archive; parallel archive parsing compounds it.

Codex was explicit that peak memory here was **inferred from allocation code, not measured**, so treat the magnitude as unverified. The shape is worth taking seriously regardless: `MAX_ARCHIVE_BYTES` bounds the decompressed bytes, not what the parser then allocates from them, and this repo has hit OOM on scan memory before (#1209). A record-size or structural-token bound before handing rows to simd-json would close it.

Happy to re-review once you have decided the ordering with #1278.

PR #1287, opened 2026-09-06
fix(antigravity-cli): reject ambiguous competing endianness readings in gen9 payload

### LINKED ISSUE #1256: fix(antigravity-cli): little-endian nanos reading can outrank big-endian millis and misdate a turn
## Summary

`inferred_epoch_ms` tries the little-endian reading of a raw 8-byte `#9.#10` payload before the big-endian one, and `epoch_scalar_to_ms` can promote that little-endian value into a plausible date through its nanoseconds branch. When it does, a genuine big-endian-millisecond timestamp is silently misdated by days.

This also makes `agy_1_1_18_gen9_field_10_dates_the_turn` an intermittent CI failure, because the test derives its fixture from `chrono::Utc::now()` and whether the little-endian reading lands inside the session window depends on today's byte pattern.

## Mechanism

`crates/tokscale-core/src/sessions/antigravity_cli.rs`:

```rust
let raw: [u8; 8] = payload.try_into().ok()?;
[u64::from_le_bytes(raw), u64::from_be_bytes(raw)]
    .into_iter()
    .filter_map(epoch_scalar_to_ms)
    .find(|&ms| accepted(ms))
```

`epoch_scalar_to_ms` documents that "over the plausible window the four unit ranges are disjoint ... so at most one unit can produce an in-window result". That holds for **one** integer. It does not hold across two different integers, and the loop feeds it two: the little-endian and big-endian readings of the same bytes. For a current-era millisecond timestamp the byte-reversed value is ~1.78e18, which the nanoseconds branch divides by 1e6 straight back into a plausible current-era millisecond value.

Both are then in-window, and little-endian wins on ordering.

## Reproduction

Two observed failures, exactly explained by this:

| intended (BE millis) | little-endian value | `/1e6` as nanos | parser returned |
| --- | --- | --- | --- |
| 1788132511000 | 1787084994892595200 | 1787084994892 | 1787084994892 |
| 1788274719000 | 1781589670136709120 | 1781589670136 | 1781589670136 |

The first is off by 12.1 days, the second by 77.4 days. Both were produced by the existing test on unmodified `antigravity_cli.rs` — the first in CI on 2026-08-31, the second locally on 2026-09-01.

## Impact

In production the window is `[session_created - SESSION_START_TOLERANCE_MS, now + FUTURE_TOLERANCE_MS]`, which is wide, so a real agy 1.1.18 row whose `#9.#10` carries big-endian milliseconds can be dated to the wrong day whenever the reversed reading also falls inside its session's lifetime. It is not detectable after the fact: the wrong value is a plausible date.

Separately, every PR's `Tests (ubuntu-latest)` job can fail on this for reasons unrelated to the PR.

## Suggested direction

Rejecting the ambiguity looks more consistent with the module's stated philosophy than picking a winner — `epoch_scalar_to_ms` already returns `None` so that "an unrelated 8-byte field (an id, a hash) falls through to the session-created stamp instead of becoming a wrong date". If both endianness readings are accepted by the window, neither is trustworthy and the row should fall back to the session stamp.

Whatever the decision, the test needs to stop deriving its fixture from `Utc::now()` for this case, or pin the anchor relative to the generated value, so the outcome is not a function of the calendar.

azestkingscrown:
Resolves #1256.

### Summary
In Antigravity CLI v1.1.18 session parsing, an 8-byte `#9.#10` payload evaluated `u64::from_le_bytes` before `u64::from_be_bytes`. For a big-endian millisecond timestamp (~1.78e12), the byte-reversed little-endian integer (~1.78e18) decodes through `epoch_scalar_to_ms`'s nanoseconds branch back into a plausible millisecond date (~12 to 77 days off). When both endianness readings fell inside an extended session window, the little-endian nanoseconds reading was picked first, silently misdating turns and causing intermittent CI test failures in `agy_1_1_18_gen9_field_10_dates_the_turn`.

### Changes
- Updated `inferred_epoch_ms` in `crates/tokscale-core/src/sessions/antigravity_cli.rs` to evaluate both little-endian and big-endian candidates. If both readings clear the session window but disagree on the date, the ambiguity is rejected (`None`), safely falling back to the session anchor rather than choosing a wrong date.
- Anchored `session_fallback` in `agy_1_1_18_gen9_field_10_dates_the_turn` shortly before `expected_ms` (matching real session lifetimes) to prevent the test window from unnaturally widening across months.
- Added regression tests in `ambiguous_competing_endianness_payload_falls_back_to_session_timestamp` reproducing the exact two observed timestamp collisions from #1256 and asserting they fall back to the session stamp.

---
## Summary by cubic
Fixes #1256 by rejecting 8-byte `#9.#10` payloads whose little-endian and big-endian readings both decode as plausible times but disagree, falling back to the session anchor instead of silently misdating turns. The ambiguity check no longer runs inside the session window (whose upper bound is the scan clock, so verdicts used to flip as dates passed); the two byte orders are settled on the bytes alone via a fixed 2020–2100 plausibility bracket and a finer-unit tie-break, and the session window only gates the survivor. Genuine little-endian millisecond readings still date the turn when the big-endian counterpart is only a finer-unit artifact.

**Bug Fixes**
- Anchors `agy_1_1_18_gen9_field_10_dates_the_turn` near the expected timestamp so the window doesn't widen to invite ambiguity.
- Adds regression tests for the two #1256 collisions, genuine LE millis precedence, and that identical bytes and anchor yield one verdict across scan times up to 25 years out.

<sup>Written for commit bd6bd38f98bc810f6008458768cd2001567e8133. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1287?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo (approved):
Approving with two maintainer commits on top of yours (untouched).

**What the Codex adversarial review found.** Both candidate readings were filtered against the session window *before* the code decided whether they compete, and that window's upper bound is the scan clock. So the verdict for unchanged bytes depended on the day you scanned: fixture `000001a05d7ab918` with anchor `1781502653000` read 2026-06-16 on a June scan and fell back to the 2026-06-15 session stamp on a September scan.

**`86d9536f`.** Settle the two byte orders against each other on unit ranges alone, then apply the window to the survivor. The scan clock is injected (`inferred_epoch_ms_at`, `session_window_ms(created, now_ms)`) so a test can hold the bytes and the anchor fixed and vary only the moment of the scan. A two-clock regression test fails on the PR's arms and passes with the reorder, and the `agy_1_1_18` fixture carries a mirror-is-not-a-time self-check so the pure-bytes contest cannot make it flake.

**`97e4c11a` — the same defect, one level down.** Adversarial review caught that the contest still read a clock: the plausibility test it cuts candidates from had an upper bound of `now + 5 years`. That is the same drift deferred, not removed. A genuine big-endian millisecond stamp whose little-endian mirror lands just past the horizon is dated today and withdrawn to the session stamp days later, while `main` dates the same bytes stably. The bound is now a fixed 2020–2100 bracket that never reads the clock, so the contest is a pure function of the bytes and `now_ms` enters exactly once, closing the session window around the survivor. The test scans one payload at clocks twenty-five years apart.

**Deliberate behavior change.** Two plausible readings that disagree are now rejected as ambiguous even when only one falls inside the session window, so slightly more rows fall back to the session stamp than under the PR alone. Letting a data-derived lower bound disqualify a mirror before the contest would recover some of them; that was considered and left out to keep the contest a pure function of the bytes. Dating a row by whichever reading happened to survive a moving window is the worse failure.

Gates observed locally: `cargo fmt --check`, `clippy -D warnings` with and without `--all-targets`, full workspace tests.

PR #1288, opened 2026-09-06
refactor(core): derive pi-format parser versions from shared base

### LINKED ISSUE #1195: Clients sharing the pi-format parser can silently desync parser_version
Four clients parse the same on-disk format through `sessions/pi.rs`, but each carries its own `parser_version()` counter, and one carries none at all. A parsing change inside `pi.rs` therefore only invalidates caches for whichever client ids someone remembers to bump.

Who shares the parser:

- `crates/tokscale-core/src/sessions/pi.rs` — the implementation
- `crates/tokscale-core/src/sessions/kimchi.rs:12` — `parse_pi_format_file_with_dedup(path, "kimchi", "kimchi")`
- `crates/tokscale-core/src/sessions/omp.rs:28` — `parse_pi_format_file(path, "omp", "omp")`
- `crates/tokscale-core/src/sessions/senpi.rs:21` — `parse_pi_format_file(path, "senpi", "senpi")`

What `parser_version()` in `crates/tokscale-core/src/message_cache.rs` currently returns:

- `:1047` — `ClientId::Pi => 2`
- `:1120` — `ClientId::Kimchi => 2`
- `:1211` — `ClientId::Omp => 1`
- Senpi has no arm and falls through to `_ => 1` at `:1212`

So a `pi.rs` fix that bumps `ClientId::Pi` from 2 to 3 leaves Kimchi at 2, Omp at 1, and Senpi on the wildcard. Cached messages for those three replay the pre-fix parse until something unrelated invalidates them. Senpi is the easiest to miss, because there is no line to edit — the wildcard silently absorbs it.

The comment at `message_cache.rs:1208` warns a human to remember this, which is the right instinct but not enforceable.

Options, roughly in order of how much they actually prevent:

Give the shared format one version constant that every delegating client's arm derives from, so a single edit moves all four together. Clients keep their own arm for changes genuinely specific to them (scan roots, dedup namespaces), added on top of the shared base.

Failing that, at minimum give Senpi an explicit arm so all four are visible in one place, and add a test that asserts every client whose parser delegates to `pi.rs` shares the same base version — a test failure is a much better reminder than a comment.

Noted while adding Oh My Pi in #1161; not a live bug today, since `pi.rs` has not changed since these clients diverged.

azestkingscrown:
## Problem

Four clients (`Pi`, `Kimchi`, `Omp`, and `Senpi`) parse the same transcript format through `sessions/pi.rs`, but previously maintained separate `parser_version()` branches in `message_cache.rs`, while `Senpi` was missing an explicit arm entirely and fell through to the wildcard `_ => 1`. A format change inside `pi.rs` therefore only invalidated caches for client IDs someone remembered to bump manually, risking silent cache desynchronization (#1195).

## Solution

1. Define `PI_FORMAT_PARSER_BASE_VERSION = 1` in `crates/tokscale-core/src/sessions/pi.rs` as the authoritative format version for the shared Pi layout.
2. In `crates/tokscale-core/src/message_cache.rs`, derive each delegating client's parser version from `PI_FORMAT_PARSER_BASE_VERSION` while preserving existing effective versions and client-specific historical offsets:
   - `ClientId::Pi`: `PI_FORMAT_PARSER_BASE_VERSION + 1` (= 2, preserving agent attribution offset)
   - `ClientId::Kimchi`: `PI_FORMAT_PARSER_BASE_VERSION + 1` (= 2, preserving namespaced dedup key offset)
   - `ClientId::Omp`: `PI_FORMAT_PARSER_BASE_VERSION` (= 1)
   - `ClientId::Senpi`: `PI_FORMAT_PARSER_BASE_VERSION` (= 1, now explicitly matched instead of falling through to wildcard)
3. Since effective parser versions for all clients remain identical, `parser_generation()` is unchanged and no existing valid caches are prematurely evicted.
4. Add unit test `test_pi_format_shared_parser_version_sync` in `message_cache.rs` verifying that all delegating clients derive from `PI_FORMAT_PARSER_BASE_VERSION`.

Resolves #1195

---
## Summary by cubic
Introduces a shared parser version base for the five clients that parse the Pi transcript format, so a format change in `sessions/pi.rs` invalidates caches for all of them at once. Previously each client had its own `parser_version()` arm (and `Senpi` fell through to the wildcard); effective versions are unchanged, so no caches are evicted.

- Derives Pi, Kimchi, Omp, Senpi, and Prime Agent from `PI_FORMAT_PARSER_BASE_VERSION`, preserving each client's effective version (Pi 2, Kimchi 2, Omp 1, Senpi 1, Prime Agent 4).
- Adds a regression test covering all five delegating clients and resolves #1195.

<sup>Written for commit a1e0b00fc551ca9a40985f1053776a0c9793f34b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/message_cache.rs
+++ b/crates/tokscale-core/src/message_cache.rs
@@ -1096,9 +1096,10 @@ fn parser_version(client: ClientId) -> u32 {
         // when reconstructing VS Code Copilot Chat requests; v8 aggregates
         // carry those sessions as zero-token.
         ClientId::Copilot => 9,
-        // Pi subagent sessions now derive agent attribution from session_info
-        // names; version-1 caches carry those messages without agent metadata.
-        ClientId::Pi => 2,
+        // Pi delegates to the shared pi-format parser. Pi subagent sessions
+        // now derive agent attribution from session_info names; version-1
+        // caches carry those messages without agent metadata.
+        ClientId::Pi => crate::sessions::pi::PI_FORMAT_PARSER_BASE_VERSION + 1,
         // Devin CLI v1 could stop at a malformed chat_message. v2->v3:
         // message timestamp is now back-calculated to the turn start
         // (created_at - total_time_ms) instead of the recorded (end-anchored)
@@ -1174,9 +1175,10 @@ fn parser_version(client: ClientId) -> u32 {
         // of landing unpriced, and uses a UTC-stable synthetic id for the id-less
         // fallback.
         ClientId::Cursor => 3,
-        // v1->v2: Kimchi's Pi-compatible messages now carry stable namespaced
-        // deduplication keys.
-        ClientId::Kimchi => 2,
+        // Kimchi delegates to the shared pi-format parser. v1->v2: Kimchi's
+        // Pi-compatible messages now carry stable namespaced deduplication
+        // keys.
+        ClientId::Kimchi => crate::sessions::pi::PI_FORMAT_PARSER_BASE_VERSION + 1,
         // v1->v2: Prime Agent now strips a leading BOM and recovers records
         // containing undecodable bytes; its accounting scan also continues past
         // those records instead of truncating and misaligning message indices.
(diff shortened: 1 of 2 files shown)


junhoyeo (approved):
The refactor is right: four clients sharing `sessions/pi.rs` now derive their parser versions from one base, so a shared-parser fix cannot silently leave one of them on a stale cache. Codex's adversarial pass found the one gap — Prime Agent also reaches `parse_pi_format_file_inner` but kept a literal `4`, so a base bump would have advanced the four listed clients while Prime kept serving stale messages and `PrimeFileAccounting`, and the new sync test passed because it omitted Prime.

Maintainer commit `a1e0b00f` on top of yours derives `PrimeAgent` as `PI_FORMAT_PARSER_BASE_VERSION + 3`, preserving 4 today, and adds it to the synchronization and stale-cache-rejection tests. Every effective version is numerically unchanged, so this PR triggers no reparse for anyone. Verified by probe: with the base set to 2, a literal Prime arm fails the sync test at exactly the Prime assertion (4 ≠ 5), and the derived arm passes it.

Gates on my machine: fmt, clippy with and without `--all-targets`, full workspace suite — 3,532 pass, 0 fail.

PR #1289, opened 2026-09-06
fix(moderation): scale slopModelName weight by matching models token share

### LINKED ISSUE #1265: moderation: scale slopModelName weight by the matching models' token share
## Problem

`slopModelName` in `packages/frontend/src/lib/moderation/heuristics.ts` is a fixed weight of 35, awarded whenever any model name matches `SLOP_MODEL_REGEX`. It never looks at how many tokens that model actually carries.

Across the whole site, five model names match the pattern:

| model | tokens |
|---|---|
| `slopai/slopllm-5m` | 4.504e15 |
| `slopllm-5m` | 3.860e15 |
| `slopai/slopllm:5m` | 6.434e14 |
| `fake-test-model` | 2 |
| `fake-api` | 0 |

The bottom two are somebody's local test provider — `fake-api` appears in exactly one opencode day row with zero tokens and zero cost. They get the same +35 as an account that pushed 9.0e15 tokens through a model called `slopllm`.

That is 2 of the 9 accounts currently in the review queue, both there solely on this signal, and neither has done anything wrong. Putting a person in front of a moderator because they named a mock provider `fake-api` is the wrong default.

## Fix

Scale the weight by the share of the account's tokens that the matching models carry:

```
weight = 35 * (slop model tokens / account total tokens)
```

Real separation in the current data:

| account | slop tokens | account tokens | share | weight |
|---|---|---|---|---|
| A | 9.007e15 | 9.026e15 | 99.8% | 34.9 |
| B | 2 | 2.622e10 | 7.6e-9% | 0.0 |
| C | 0 | 9.137e8 | 0% | 0.0 |

Ten orders of magnitude between the real case and the false positives, so no threshold constant is needed — the scaling alone does it. Drop the signal when the weight rounds to zero so the account leaves the queue rather than sitting in it at score 0.

Queue effect: 9 candidates -> 7. The top-scoring account goes 99.7 -> 99.6 and stays rank 1; only the two false positives leave.

## Why the original comment still holds

The existing comment argues the name speaks for itself, and it does — but only when the name was used to book usage. A model name with no tokens behind it is a config artifact, not a claim about the leaderboard. The signal keeps its meaning and gains a precondition.

`candidates.ts` already computes `slop_models` per user; it needs to also sum the tokens attributed to those model keys out of `daily_breakdown.source_breakdown`, alongside the existing per-user aggregates.

azestkingscrown:
## Summary
Fixes #1265

This PR scales the `slopModelName` heuristic weight in the moderation review queue by the matching models' share of the account's total tokens (`weight = 35 * (slopTokens / totalTokens)`). When the computed weight rounds to zero (e.g., negligible or zero tokens from client configuration artifacts), the signal is dropped so false-positive accounts do not pollute the review queue.

Key changes:
- `packages/frontend/src/lib/moderation/heuristics.ts`: Added `slopTokens` to `CandidateRow`, scaled `slopModelName` weight proportionally to `slopShare`, and dropped the signal when `Math.round(weight) <= 0`.
- `packages/frontend/src/lib/moderation/candidates.ts`: Added a `slop_usage` CTE to sum tokens booked under models matching `SLOP_MODEL_REGEX` from `daily_breakdown.source_breakdown` for eligible candidates with slop models, safely guarding JSONB operations.
- `packages/frontend/__tests__/lib/moderationHeuristics.test.ts`: Added unit tests covering proportional weight scaling, dropping negligible-share signals, and omitting clean accounts with trivial slop models from `rankCandidates`.

---
## Summary by cubic
Scales the `slopModelName` moderation weight by the matching models' share of the account's total tokens, so accounts flagged only for config artifacts drop out of the review queue. When token attribution can't be verified as complete, `slopTokens` comes back NULL and the full fixed weight of 35 is kept instead of scaling.

- Adds `slopTokens` to `CandidateRow`, summing per-model usage from `source_breakdown` including the legacy client-level `modelId` remainder.
- Treats `unknown`/debris model-map keys and entries whose map outruns their own scalar as unattributed, and requires every daily row to carry a breakdown before scaling applies.
- Drops the `slopModelName` signal when the scaled weight rounds to zero.
- Adds unit tests plus a PostgreSQL integration fixture (17 attribution personas) run in the frontend migration CI job.

<sup>Written for commit fb0c3171d1923f02c952502e69a06cb5e3dad3ee. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1289?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo (approved):
Approving with four maintainer commits on top of your two (untouched).

**What the Codex adversarial review found, and what my own review added.** The weight scaling read nested `models[id].tokens` as if it were the whole story. A legacy row carries the remainder in the scalar `tokens`, so an entry with `tokens: 1,200,000` and one nested `{fake-api: 2}` counted 2 attributed tokens: the JSON's presence marked attribution "available", the weight rounded to 0, and the account dropped out of the moderation queue. Fixed by failing closed — incomplete attribution returns null and the candidate keeps full weight.

Reproduced at two levels before anything was changed. At the write path, feeding the real `mergeClientBreakdownsWithRegressionGuard` the arguments the submit route passes returned one entry with `tokens` 3,100,000 while its model map summed to 6,200,000, because `applyCostCompleteness` unions the two maps but returns the incoming scalar. Then in SQL against a migrated `postgres:16`: attributed 3,100,002 against a total of 3,100,000, weight rounding to 0, signal gone.

**The four commits.**
- Unnamed model cells (`unknown` and friends) count as unattributed rather than as named attribution.
- Unnamed cells are charged against the entry's own scalar: `GREATEST(LEAST(nested_named_tokens, scalar - nested_unnamed_tokens), 0)`. On any entry whose map fits inside its scalar this is algebraically inert, which is why all nine existing personas are unchanged.
- A model map that outruns its own scalar attributes nothing instead of overcounting.
- An entry that contradicts itself and carries no tokens fails closed.

**Two things I checked rather than assumed.** The 17 Postgres tests that carry this proof are skipped without a database, so I confirmed they actually run in CI: the PR wires `test:moderation:db` into the migration workflow's `postgres:16` job beside the existing reconciliation suite, and extends that workflow's path filter to cover `src/lib/moderation/**` and the new test file, so the job triggers on exactly these changes. I also confirmed the mutation coverage claim: deleting either of the two regex filters used to leave the suite green, and both now have a persona that fails.

**Known scope.** The gate is broad on purpose. Any account holding a single `unknown` token in its daily rows gets full weight, so the share scaling from #1265 is inactive for it. That is the fail-closed direction the finding asks for, but how many real accounts sit in that state is unmeasured; no production query was run.

Frontend gates locally: install, typecheck, lint, and the suite at 985 passed with the 23 database-backed tests skipped for want of a local Postgres.

junhoyeo:
Correction to my approval above, so the record matches the merged code.

I described the fix as the clamp `GREATEST(LEAST(nested_named_tokens, scalar - nested_unnamed_tokens), 0)`. That was the intermediate shape, and the later commits in this stack deliberately replaced it. The comment now in `candidates.ts` explains why, and it is worth restating: every clamp of the named cells has a ceiling of the entry's **own** scalar, while the gate's threshold is `total_tokens`, the sum of those same scalars. So a clamped over-nested entry does not land below the gate, it lands exactly on it, which is the passing side. The clamp only bit when an unnamed cell happened to subtract from the ceiling; with the excess under a real model name, or in a map with no unnamed cell at all, attribution still came back equal to the total.

What actually ships is a guard sitting **outside** the attributed sum: `over_nested` is true when an entry's nested map outruns its own scalar, which zeroes that entry's attributed tokens, and the flag is aggregated with `BOOL_OR` across the submission, so one self-contradictory entry makes the whole submission's attribution unknowable and the candidate keeps full weight. That is a stronger fail-closed than the clamp, not a variant of it.

Everything else in the approval stands: the reproduction, the CI wiring for the Postgres suite, the mutation coverage, and the noted breadth of the gate.

PR #1291, opened 2026-09-06
fix(openclaw): deduplicate events across transcript archives

junhoyeo:
## Summary

- Follow up on #1285: count an OpenClaw assistant event once when its plain transcript and compressed archive coexist, while preserving archive-only history and distinct events.
- Apply stable event-ID-based deduplication to cached cold/warm scans and direct aggregation. Events without usable IDs remain separate rather than being collapsed by matching usage alone.
- Advance the OpenClaw parser version from 2 to 3 so previously cached messages without dedup keys are reparsed.

## Verification

- Regression failed before the fix: 6 messages instead of 5.
- OpenClaw tests: 22 passed.
- Independent cache tests: 73 passed.
- Changed-file diagnostics, formatting, strict clippy for core/CLI including all targets, and workspace build passed.
- Independent real CLI fixture: both cold and warm scans report 3 messages, 240 input tokens, and 24 output tokens, preserving the archive-only event.
- Independent requirements, code, security, context, and hands-on QA reviews passed.

## Integration

This is separate from the checkpoint-artifact exclusion follow-up. Merge this parser-version-3 change before that follow-up's version-4 change. Pending #1278 must preserve this decoder/dedup behavior and advance the parser version again if its rebase changes cached message interpretation.

---
## Summary by cubic
Deduplicates OpenClaw assistant events across plain transcripts and compressed archives, so the same event is counted once instead of once per copy. Uses stable event IDs when available; events without IDs remain separate, and archive-only history is preserved. Bumps the OpenClaw parser version to 3 to invalidate cached messages that lack dedup keys.

<sup>Written for commit 3be6ed48abd2a22dc6e9bc5db2fb64441e9df8e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1291?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-core/src/lib.rs
+++ b/crates/tokscale-core/src/lib.rs
@@ -2375,7 +2375,7 @@ fn parse_all_messages_streaming<S: MessageSink>(
         }
     }
 
-    parse_cached_lane(
+    parse_cached_lane_deduped(
         &scan_result,
         &mut source_cache,
         pricing,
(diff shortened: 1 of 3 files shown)


junhoyeo:
Superseded by #1278.

This was a standalone follow-up to #1285, but #1278 has since been rebased onto #1285 and carries the same fix as part of its dedup design: event-level keys from the top-level `id`, applied in both aggregation paths, with `OpenClaw` parser version 3. Verified on `6fe35347` that a plain transcript and its `.zst` copy produce the identical key and the lane keeps one. Keeping both would add three-file conflicts without adding behavior.

The focused cold/warm/direct regression from this branch is being checked against #1278's coverage and carried over if #1278 lacks an equivalent.

PR #1294, opened 2026-09-06
chore(ci): pin Rust toolchain to 1.98.0

### LINKED ISSUE #1237: Pin the Rust toolchain: local clippy is older than CI's, so a green local gate still fails Lint

azestkingscrown:
## Summary
Pins the Rust toolchain version to `1.98.0` in a new `rust-toolchain.toml` at the repository root and updates all CI workflows (`build-native.yml`, `publish-cli.yml`, `test_coverage.yml`) from `dtolnay/rust-toolchain@stable` to `@1.98.0`.

Resolves #1237

---
## Summary by cubic
Pins the Rust toolchain to 1.98.0 via a new `rust-toolchain.toml` at the repo root and switches all CI workflows from `dtolnay/rust-toolchain@stable` to `@1.98.0`. This resolves #1237 and makes local and CI builds use the same compiler.

**Workflow path filters**
- Rust workflows now list `rust-toolchain.toml` in their `paths:` filters so a toolchain-only bump re-runs builds, lint, and tests.
- `workflow_lint.yml` runs a new checker that fails if a Rust workflow's `paths:` omits the pin or `paths-ignore:` lists it, backed by a test suite.

<sup>Written for commit 2f71c05ecc8d36304aaf164ae8c4ace2f5a740c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -125,7 +125,7 @@ jobs:
           path: .
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@1.98.0
         with:
           targets: ${{ matrix.settings.target }}
 
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -17,6 +17,7 @@ on:
       - "packages/cli-*/package.json"
       - "Cargo.toml"
       - "Cargo.lock"
+      - "rust-toolchain.toml"
 
 permissions:
   contents: read
--- a/.github/workflows/launcher_validation.yml
+++ b/.github/workflows/launcher_validation.yml
@@ -20,6 +20,7 @@ on:
       - "bun.lock"
       - "Cargo.toml"
       - "Cargo.lock"
+      - "rust-toolchain.toml"
       - "crates/**"
   pull_request:
     branches: [main]
--- a/.github/workflows/publish-cli.yml
+++ b/.github/workflows/publish-cli.yml
@@ -62,7 +62,7 @@ jobs:
         run: python3 scripts/check-release-workflow-safety.py
 
       - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@1.98.0
 
       - name: Calculate version
         id: bump
--- a/.github/workflows/test_coverage.yml
+++ b/.github/workflows/test_coverage.yml
@@ -8,6 +8,7 @@ on:
       - 'crates/**'
       - 'Cargo.toml'
       - 'Cargo.lock'
+      - 'rust-toolchain.toml'
       - '.github/workflows/test_coverage.yml'
   pull_request:
     branches: [main]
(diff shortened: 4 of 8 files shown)


junhoyeo (approved):
The pin is right, and Codex's adversarial pass found the one gap: `rust-toolchain.toml` now decides the compiler, but `build-native.yml`, `test_coverage.yml` and `launcher_validation.yml` did not list it in their path filters — verified at your head, where a checker run reports five omissions. This PR only triggered CI because it also edits workflows; a later toolchain-only bump would have skipped native builds, lint and tests while changing the compiler, and the first failure would have surfaced in manual publishing.

Two maintainer commits on top of yours: `8b176ac0` adds `rust-toolchain.toml` to all three filters (via the shared anchors, so each workflow's reuse is covered), plus `scripts/check-workflow-toolchain-paths.py` that scans every workflow using cargo or a Rust setup action and fails if its `paths:` filter omits the toolchain file, with a fixture suite in `scripts/test-workflow-toolchain-paths.sh`. `2f71c05e` widens the checker after review: `.yaml` as well as `.yml`, an inline `on: {…}` mapping, and a `cargo`-only marker case, and is otherwise a message-only amend.

Gates on my machine under the pinned 1.98.0: actionlint clean, fmt, clippy with and without `--all-targets`, full workspace suite.

PR #1296, opened 2026-09-06
perf(copilot): read Copilot CLI exports in two passes instead of holding every record

### LINKED ISSUE #1209: Excessive memory usage

azestkingscrown:
## Summary

Fixes #1209.

The Copilot CLI OTEL parser (`~/.copilot/otel/*.jsonl`) read every line's full JSON DOM into one `Vec<Value>` and kept it alive for both the trace-map walk and the usage-emission walk. On the reporter's 9.7 GB export (397,071 records at ~26 KB each, ~2.7x serde_json expansion) that is roughly 26 GB of allocation for a trace map worth tens of MB — before anything is counted.

This changes the parser to read the file twice: pass 1 copies only the small owned fields the trace map needs (model, session id, span hierarchy, agent ids) out of each line and drops the line's JSON value at once; pass 2 re-reads and emits usage against the finished map. Peak memory then scales with trace count, not file size. The double read is I/O rather than memory, a far better trade on files this size.

Output is unchanged: same `parse_copilot_file` signature, same candidate selection, dedup, and merge logic, all existing tests pass unmodified, plus one new regression test proving a usage record exported *before* the invoke_agent span that attributes it still resolves its model, session, and agent (the forward reference the two-pass ordering exists for).

## Verification

- `cargo fmt --all -- --check` passes
- `cargo clippy --locked --workspace --all-features -- -D warnings` passes
- `cargo test --workspace --all-features --no-fail-fast` passes (all suites green, 0 failures)
- `cargo test -p tokscale-core --lib copilot`: 103 passed, 0 failed (102 existing + 1 new `two_pass_forward_reference_resolves_trace_context`)
- Measured on a 13 MB synthetic export (40,000 records, 20,000 traces): identical output (20,000 messages, same token totals), peak RSS 110 MB before → 35 MB after

## What this does not do

- No parser version bump: output is byte-identical in intent, so existing caches stay valid.
- No cap on parallel file parsing (the second half of the maintainer's note in #1209): worth doing, but a separate change with its own review surface.
- No per-line field projection during pass 1: transient-only savings once the two-pass change lands, also separable.

---
## Summary by cubic
Cuts peak memory for parsing Copilot CLI exports (`~/.copilot/otel/*.jsonl`) by holding only the small trace-map fields and one compact candidate per usage record instead of every line's full JSON DOM. Also fixes the initial two-pass approach, which reopened the file and could attribute usage from bytes that landed after the first read hit EOF.

- Reads the file once, so the trace map and usage always come from the same bytes even when Copilot CLI appends to the live export mid-scan.
- Keeps the same `parse_copilot_file` signature and byte-identical output, so existing caches stay valid.
- Adds regression tests covering forward references, malformed lines not shifting dedup keys, and the same-read guarantee.
- On a 9.7 GB export this removes roughly 26 GB of temporary allocation; a 40,000-record synthetic export dropped peak RSS from ~111 MB to ~46 MB.

<sup>Written for commit ce7336ed615647e6afe8c5d5c6e9985fa9230369. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1296?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


junhoyeo (approved):
Approving with three maintainer commits on top of yours (untouched). The two-pass shape is the right memory fix and the numbers below keep it; the first commit closes a window the Codex adversarial review found in it, and the other two are about the test that proves it.

**What the review found.** Pass 2 reopened the export, so a usage/context pair appended between the passes was emitted with model `unknown`, session = trace id, and no agent. A test grafted onto the PR head fails with exactly that (`left: "unknown"`, session `trace-2`). Same shape, second consequence: `for_each_json_line` treats an open failure as EOF, so a failed pass-1 open plus a successful pass 2 produced a non-empty, cacheable, wholly unattributed result.

**`8e78ab94` — the production fix.** One `for_each_json_line` call accumulates the trace map and keeps one compact `PendingUsageCandidate` per usage record (a handful of owned fields, never the JSON value); the candidates resolve against the finished map after the read, so map and usage come from the same open. A failed open now yields an empty, uncacheable result.

**The perf win is preserved, measured (peak RSS):**

| fixture | main | this PR | with fix |
|---|---|---|---|
| 40k records / 20k traces | 111–113 MB | 39–40 MB | 46 MB |
| 8k records × 12 KB | 127–128 MB | 13–14 MB | 13–14 MB |

Identical outputs, and the second file read is gone.

**`a4e65685` and `84edf672` — the regression test, and an honest retraction.** The test proves single-read parsing through a FIFO, and adversarial review found it could hang rather than fail if the parser never opened the export. The first remedy gave the writer a non-blocking open, which was worse: a non-blocking write-only FIFO open returns `ENXIO` unless a reader is already blocked in `open`, and this writer usually reaches its open first, so it died and the parser then waited forever. Measured on that commit: 7 hangs in 40 runs. `84edf672` restores the blocking open, which *is* the rendezvous and cannot lose a race, and bounds the other side instead by running the parse on its own thread with a 30 s deadline. Measured after: 80 runs, 80 passes, no hangs. Both failure modes were reproduced deliberately before and after, and a `Directive` trailer forbids re-adding the non-blocking open.

**Two things to know.** The PR title says "two passes"; after `8e78ab94` the parser makes one read, so I will squash under a title that says so. The test is Unix-only, so the Linux and macOS jobs run it and the Windows job skips it; I verified on macOS.

Gates observed locally: `cargo fmt --check`, `clippy -D warnings` with and without `--all-targets`, full workspace tests.

Issue #986, opened 2026-07-30
opencodereview declares submit_default: true but the submit path never parses it


junhoyeo:
Found while re-studying #960. Verified against `origin/main`.

## The gap

`crates/tokscale-core/src/clients.rs` registers OpenCodeReview with `submit_default: true`, so it is included by default when a user runs `tokscale submit`.

But `parse_opencodereview_file` has exactly **one** call site in the whole crate:

```
$ grep -rn "parse_opencodereview_file" crates/tokscale-core/src/ | grep -v sessions/opencodereview.rs
crates/tokscale-core/src/lib.rs:3205:  sessions::opencodereview::parse_opencodereview_file(path)
```

and it sits inside `parse_local_clients` (`lib.rs:2855`), whose callers are `report.rs`, `wrapped.rs` and `main.rs` — the local reporting surfaces. The scan that feeds the graph and the submit payload, `parse_all_messages_with_pricing_with_env_strategy`, does not reference the client at all.

So OpenCodeReview usage shows up in `tokscale report` and never reaches the leaderboard, while the registry advertises that it will.

## Corroborating signal

`parser_version` for this client is now **3** — bumped twice, most recently by #978 for a dedup correctness fix. Every one of those bumps is inert for submission, because the cache they invalidate is never consulted on a path that runs for this client. #978's fix is real and correct; it simply cannot reach anyone's submitted totals.

That is a good marker for how long this has been true without anyone noticing: the client has been maintained as though it submits.

## Which way to fix it is a judgement call

**Route it through the submit scan.** OpenCodeReview usage starts counting on the leaderboard. That is presumably the intent of `submit_default: true`, but it is a retroactive increase for every existing user of that client — their totals jump on the next submit, with no way to distinguish "we fixed a gap" from "this account inflated". Worth deciding deliberately rather than as a side effect.

**Or set `submit_default: false`.** Honest about current behaviour and a one-line change, but it makes the omission permanent for a client that looks fully supported everywhere else.

I have not picked one. The first is probably right, but it changes leaderboard numbers and that deserves an explicit call.

## Also worth a look while in here

`parse_local_clients` is the only consumer, so if any other registered client is in the same position the same silent gap applies to it. I checked only OpenCodeReview.

Yuxin-Qiao:
Reproduced and audited. The gap described here is already fixed on `main` by #990 (`b9fd593a`, merged 2026-07-30); this issue just wasn't linked/closed.

## Reproduction

[...]

PR #1300, opened 2026-09-06
fix(core): parse MiMo Code and fx on the local report path

MaxFreedomPollard:
## Summary

MiMo Code and fx usage never shows up in `tokscale report`, and `tokscale clients` prints a zero message count for both, even though both clients scan and submit fine.

## Cause

`MiMoCode` and `Fx` both declare `parse_local: true` in `define_clients!` (`crates/tokscale-core/src/clients.rs`), and the submit path parses both: MiMo Code from `scan_result.micode_dbs` at `crates/tokscale-core/src/lib.rs:2047`, fx from `scan_result.get(ClientId::Fx)` at `crates/tokscale-core/src/lib.rs:3133`. `parse_local_clients` (same file, from line 5134) has a block for every other locally parsed client and none for these two, so it returns no rows and sets no count for either.

Everything built on `parse_local_clients` therefore sees nothing. `tokscale report` builds its whole table from the wiki DB, and `populate_wiki_from_sessions` (`crates/tokscale-cli/src/commands/report.rs:114`) fills that DB from `parse_local_clients` alone, so a MiMo Code or fx session is never recorded. `run_clients_command` (`crates/tokscale-cli/src/main.rs:4266`) asks for every `parse_local()` client and then prints `parsed.counts.get(client)`, so both rows read 0 no matter how much local data exists.

This is the mirror of #986, where opencodereview was parsed locally but missing from the submit path.

## Fix

MiMo Code goes in after the OpenCode block and reads `scan_result.micode_dbs`, deduplicating on the payload message id through the existing `should_keep_deduped_message` helper. That part matters: MiMo writes channel-suffixed databases, so one session can appear in both `mimocode.db` and `mimocode-<channel>.db`, and the submit lane already collapses them the same way.

fx goes in after the Mux block and reads `scan_result.get(ClientId::Fx)`. Session titles are deliberately not applied here, because `ParsedMessage` has no title field for the shared `sessions/index.json` lookup to write to.

Both count through `summed_parsed_message_count` rather than the row count, so an fx row contributes its `request_count`.

## Verification

The new test writes a `~/.local/share/mimocode/mimocode.db` with one assistant row and a `~/.fx/sessions/sess-1/usage-v2.json` snapshot, then asserts both reach `parse_local_clients` with their token buckets and client counts.

With the test applied to an otherwise unmodified tree:

```
$ cargo test --locked -p tokscale-core --all-features --lib -- test_parse_local_clients_includes_micode_and_fx
assertion `left == right` failed: MiMo Code assistant row must reach the local parse, got []
  left: 0
 right: 1
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 2110 filtered out
```

With the fix:

```
$ cargo fmt --all -- --check
(clean)

$ cargo clippy --locked -p tokscale-core --all-features --all-targets -- -D warnings
(no warnings)

$ cargo test --locked -p tokscale-core --all-features --no-fail-fast
test result: ok. 2109 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
(plus the eight integration binaries, all ok)
```

The same suite on `main` before the change was 2108 passed, 0 failed, so the only difference is the new test. Ran on macOS 15 with rustc 1.95.0. The test uses only tempdirs and fixed literals, with no clock, locale or timezone dependence, so it behaves the same on the Windows leg.

---
## Summary by cubic
Fixes the local report path so MiMo Code and fx usage now appears in `tokscale report` and `tokscale clients` instead of always showing zero, and preserves provider-reported dates and costs so the figures are correct.

- `parse_local_clients` now parses MiMo Code from `scan_result.micode_dbs`, deduplicating by payload message id since one session can appear in both `mimocode.db` and `mimocode-<channel>.db`, and fx from per-session usage snapshots.
- fx rows now derive their date before conversion so `year`/`since`/`until` filters work on the local lane.
- `ParsedMessage` now carries `cost` and `cost_source`; `compute_msg_cost` uses a provider-reported cost as-is, so a token-free fx session records its snapshot cost instead of $0.00.
- Both clients count through summed message counts, so an fx row contributes its `request_count`.

<sup>Written for commit 8891eb3cbd8ded415cbe693d53342e25846d4c55. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1300?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/tokscale-cli/src/commands/report.rs
+++ b/crates/tokscale-cli/src/commands/report.rs
@@ -13,7 +13,9 @@ use tokscale_core::content_extractor::SessionContent;
 use tokscale_core::content_extractor::{extract_session_content, metadata_only_content};
 use tokscale_core::pricing::PricingService;
 use tokscale_core::wiki::{WikiDb, WikiEntry};
-use tokscale_core::{parse_local_clients, LocalParseOptions, ParsedMessage, TokenBreakdown};
+use tokscale_core::{
+    parse_local_clients, CostSource, LocalParseOptions, ParsedMessage, TokenBreakdown,
+};
 
 pub struct ReportOptions {
     pub json: bool,
@@ -107,10 +109,6 @@ pub fn run_report(opts: ReportOptions) -> Result<()> {
 }
 
 fn populate_wiki_from_sessions(db: &WikiDb, opts: &ReportOptions) -> Result<()> {
-    let existing = db
-        .get_existing_session_ids()
-        .map_err(|e| anyhow::anyhow!("{}", e))?;
-
     let parsed = parse_local_clients(LocalParseOptions {
         home_dir: opts.home_dir.clone(),
         use_env_roots: opts.home_dir.is_none(),
@@ -124,9 +122,24 @@ fn populate_wiki_from_sessions(db: &WikiDb, opts: &ReportOptions) -> Result<()> 
 
     let pricing = load_pricing_service();
 
+    record_new_sessions(db, &parsed.messages, pricing.as_deref())
+}
+
+/// Aggregate `messages` per session and record every session the wiki has not
+/// seen yet. Recorded sessions are never rewritten, so the figures computed
+/// here are the ones the wiki keeps.
+fn record_new_sessions(
+    db: &WikiDb,
+    messages: &[ParsedMessage],
+    pricing: Option<&PricingService>,
+) -> Result<()> {
+    let existing = db
+        .get_existing_session_ids()
+        .map_err(|e| anyhow::anyhow!("{}", e))?;
+
     let mut session_map: HashMap<String, SessionAgg> = HashMap::new();
 
-    for msg in &parsed.messages {
+    for msg in messages {
         let agg = session_map
             .entry(msg.session_id.clone())
             .or_insert_with(|| SessionAgg {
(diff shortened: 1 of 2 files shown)


junhoyeo (approved):
Your diagnosis is right: `MiMoCode` and `Fx` both declare `parse_local: true`, the submit path parses both, and `parse_local_clients` had a block for every other locally parsed client and none for these two — so `tokscale report` and `tokscale clients` showed nothing for either. The two new blocks are the correct fix.

An adversarial Codex pass found two integration defects in the new fx block, both reproduced and fixed in maintainer commit `8891eb3c` on top of yours:

- **Reported cost was dropped.** The fx conversion discarded `cost` and `cost_source`, so an fx snapshot with a positive `total_cost`, empty models and zero tokens — a shape fx explicitly supports — reached the report as a provider-priced `fx-unknown` row at $0, and the wiki then kept that figure on later scans. The conversion now carries the `ProviderReported` provenance through, matching the streaming path; `wiki_keeps_the_reported_cost_of_a_token_free_fx_session` and `test_parse_local_clients_keeps_provider_reported_cost` pin it.
- **fx rows carried an empty date.** `parse_fx_file` returns `date` as an empty string (the only parser under `sessions/` that does), and the direct local path skipped the derived-field refresh the streaming loader performs, so `year` and `since` filters would drop every fx row. The block now refreshes the derived date before conversion; `test_parse_local_clients_fx_rows_carry_a_derived_date` pins it. To be precise about impact: no shipped command exercises a date filter against this block today (`report.rs` and `main.rs` pass `since`/`until`/`year` as `None`, and `wrapped` builds its graph from the streaming lane), so this hardens the public `parse_local_clients` API for future filtered callers rather than fixing a visible symptom.

The MiMoCode block was checked for both and has neither: `micode.rs` derives its own date, and its embedded per-message cost reaches the lane with `ProviderReported` provenance.

One nit, not blocking: `parsed_to_unified`, the public reverse of `unified_to_parsed`, still stamps `CostSource::Unknown`, so a `ParsedMessage` carrying a provider-reported cost loses its provenance on the round trip. Nothing shipped calls it beyond the round-trip test.

Gates on my machine: fmt, clippy with and without `--all-targets`, full workspace suite.

PR #1316, opened 2026-09-08
fix(submit): report the parser high-water deficit instead of adding nothing silently

### LINKED ISSUE #1315: antigravity-cli silently contributes nothing since 2026-08-26: lifetime high-water vs a pruning store
## Summary

`antigravity-cli` stopped contributing on **2026-08-26** and has contributed nothing since, across two devices, while every submit reports success. The cause is structural: the parser high-water bounds a submission by the device/client **lifetime** aggregate, but `antigravity-cli` reads a store that **prunes old conversations**. Once the rolling scan window falls below the credited lifetime total, `tokenBudget = positive(incoming - credited)` is 0 forever and the client is silently frozen.

`~388M tokens / ~6,891 messages` from two devices have been dropped so far.

## Reproduction

CLI 4.15.1, client `antigravity-cli`, model `google/gemini-3.8-flash`.

1. `tokscale submit` (plain, so `scanScope.fullHistory === true`, `parserVersions: {"antigravity-cli": 1}`).
2. Output lists `antigravity-cli` under "Data to submit -> Clients:", API returns 200, "Successfully submitted!", profile `updatedAt` advances.
3. `/u/shunkakinoki` has **zero** `antigravity-cli` rows for every date from 2026-08-27 onward.

Stored `antigravity-cli` on the profile today:

| | |
|---|---|
| days present | 27, `2026-07-30` .. `2026-08-26` |
| lifetime tokens | 610,766,803 |
| lifetime messages | 15,786 |
| last accepted models | `gemini-3.6-flash`, `gemini-3.7-flash`, `gemini-3.7-flash-safety-le` |
| occurrences of `gemini-3.8-flash` | **0** |

Locally, on the same devices, right now:

| device | scan | tokens | messages | dates |
|---|---|---|---|---|
| A (Linux) | all-time `--client antigravity-cli` | 379,533,239 | 6,735 | `2026-09-01` .. `2026-09-08` only |
| B (macOS) | all-time `--client antigravity-cli` | 44,792,750 | 1,060 | spans the pruned window |

Device A's **all-time** scan equals its **last-7-days** scan. The store keeps roughly a week.

## Root cause

`a8329aac` (2026-08-26) added `"antigravity-cli": 1` to `SUPPORTED_VERSIONED_PARSERS`. The date the client disappears from the profile is the date of that commit.

In `packages/frontend/src/lib/db/parserHighWater.ts`:

```ts
let tokenBudget = positive(incomingAggregate.tokens - previousAggregate.tokens);
let messageBudget = positive(incomingAggregate.messages - previousAggregate.messages);
```

`previousAggregate` is the credited **lifetime** ledger (`state.aggregate`, or `aggregateSnapshot(existingLegacyDays)` at the baseline transition). It only grows. `incomingAggregate` is the lifetime total **as the parser can still see it**, which for a pruning store is a rolling window.

So on both devices:

* first post-rollout submit: `hasLegacy === true` -> `baseline-legacy`, `incoming (window) < legacy (months of stored rows)` -> `increments = {}`, `nextState` credits the legacy days
* every submit after: `incremental`, `tokenBudget = positive(window - lifetime) = 0`
* per-cell `tokenCapacity` never gets a chance to matter; nothing is ever written again

`previousObservedAggregate` is not the problem -- `observedDays` is replaced by each snapshot, so it already tracks the window. Only the credited aggregate is unbounded.

There is no warning on this path. `mode` is `incremental`, and only `freeze`, `baseline-legacy` and `replace` push one, so the submit response is byte-identical to a successful one.

## Why the obvious fixes do not work

I tried both and both break existing tests, correctly:

1. **Scope the credited aggregate to the incoming snapshot's date span** (forgive credited days older than `min(incomingDays)`). This breaks `preserves legacy rows and records the first v2 full snapshot as a no-add baseline` and 12 others, and it breaks `does not raise the stored total when a session is re-dated per generation` in `submitAntigravityCliHighWater.test.ts` -- because `SESSION_START_DATING` -> `PER_GENERATION_DATING` moves `2026-08-07` forward past the new floor. A day vanishing from below the floor is exactly the shape of the re-attribution the registry entry exists to stop.
2. **Budget against `observedDays` instead of the credited aggregate.** Breaks `never credits more than lifetime high-water growth across moves, deletions, and replays`, which asserts `credited <= lifetimePeak - 100`. A random walk of positive deltas obviously exceeds the peak.

The two situations are genuinely indistinguishable server-side. "The oldest credited day vanished and the lifetime total dropped" describes both a pruned store and a re-dating, and the design deliberately resolves that ambiguity against crediting.

## Possible fixes

1. **Declare retention in `scanScope`.** The CLI knows which stores prune; the server does not. A per-client retention floor (or a `completeHistory: false` bit distinct from the existing `fullHistory`) lets `planParserHighWaterSubmission` subtract credited days below the floor without guessing. Costs a wire-format field and a CLI change.
2. **Bump the parser generation when dating semantics change.** The comment in `submitAntigravityCliHighWater.test.ts` explains why generation stayed at 1 -- bumping would freeze every shipped CLI. But leaving it at 1 is what forces the lifetime bound to hold at *every* step rather than only across the transition, which is what makes it permanent here.
3. **Unregister `antigravity-cli`.** Restores the data at the cost of reintroducing the re-attribution inflation. Probably not what you want.

## PR

I opened a PR for the part that is unambiguously wrong regardless of which fix you pick: the silence. It carries the deficit on the plan and warns with it, so the submit response says the device is contributing zero for a client instead of looking successful. It does not change what gets credited, and all 1008 frontend tests pass.

Happy to implement (1) if you'd like it -- I'd want your call on the wire-format shape first.

shunkakinoki:
Closes #1315 partially -- this is the diagnosability half, not the fix for what is dropped.

## Problem

A device whose versioned-parser scan reports fewer lifetime tokens than it has already been credited can never allocate growth. `allocateIncrements` budgets `positive(incoming - credited)`, and the credited ledger only grows, so once the incoming aggregate falls below it every later submission adds exactly zero.

Nothing says so. `mode` is `incremental`; only `freeze`, `baseline-legacy` and `replace` push a warning. The submit returns 200 with the client listed under "Data to submit", the CLI prints "Successfully submitted!", and the profile silently stops advancing for that client.

That is how `antigravity-cli` contributed nothing from two of my devices for 13 days -- ~388M tokens, ~6,891 messages -- with no signal anywhere. #1315 has the full analysis of *why* the aggregate falls below the ledger for that client (its store prunes old conversations, so a full scan is a rolling window).

## Change

- `ParserHighWaterPlan` carries `highWaterDeficit`: how far the credited lifetime exceeds what this snapshot reports.
- The submit route warns with it, naming the shortfall and why nothing was added.

No change to what is credited. The bound is deliberate -- a snapshot below the credited lifetime is indistinguishable from a re-attribution, which is the whole point of the registry entry -- so this only makes the state visible.

## Tests

New route regression in `submitAntigravityCliHighWater.test.ts`: baseline a device at 240,000 tokens, then submit a scan whose window has aged past it and reports 90,000 genuinely new tokens. Asserts the bound still holds (stored total stays 240,000, no new day) **and** that the response names the 150,000 shortfall.

`bun run test` 1008 passed / 27 skipped, `bun run typecheck` clean, `bun run lint` no new warnings.

---
## Summary by cubic
Warns in the submit response when a device's parser scan reports fewer lifetime tokens than already credited, so a client silently contributing zero now sees why instead of a successful-looking 200. Partially closes #1315 with the diagnosability half.

**Bug Fixes**
- Carries the deficit on the parser high-water plan and names the shortfall in the submit response.
- Does not change crediting behavior; the bound still holds since a snapshot below the credited lifetime is indistinguishable from a re-attribution.

<sup>Written for commit ceae2711aa529ba836856a64113e491563803ec5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1316?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/frontend/__tests__/api/submitAntigravityCliHighWater.test.ts
+++ b/packages/frontend/__tests__/api/submitAntigravityCliHighWater.test.ts
@@ -606,6 +606,41 @@ describe("POST /api/submit antigravity-cli re-attribution high-water", () => {
     ).toBe(260_000);
   });
 
+  it("reports the deficit when the store ages history out from under the high-water", async () => {
+    const store = newStore();
+    installTx(store);
+    const first = submissionBody("antigravity-cli", SESSION_START_DATING);
+    mockSubmit(first);
+    expect((await post(first)).status).toBe(200);
+
+    // The CLI keeps a rolling conversation store, so weeks later a full scan
+    // no longer reaches the credited session at all: it reports 90,000
+    // genuinely new tokens against a 240,000 credited lifetime.
+    installTx(store);
+    const aged = submissionBody("antigravity-cli", [
+      { date: "2026-09-01", tokens: 90_000, messages: 5 },
+    ]);
+    mockSubmit(aged);
+    const response = await post(aged);
+    expect(response.status).toBe(200);
+    const json = await response.json();
+
+    // The bound holds, which is the point of the registry entry: a snapshot
+    // below the credited lifetime is indistinguishable from a re-attribution.
+    expect(json.metrics.totalTokens).toBe(240_000);
+    expect(store.days.map((day) => day.date)).toEqual(["2026-08-07"]);
+
+    // What must not happen silently. Without the warning this response is
+    // identical to one that stored everything.
+    expect(
+      json.warnings.some(
+        (warning: string) =>
+          warning.includes("Added no Antigravity CLI usage") &&
+          warning.includes("150,000"),
+      ),
+    ).toBe(true);
+  });
+
   it("still inflates for a client that is legitimately not registered", async () => {
     // Claude's parser does not re-attribute submitted history, so it is not in
     // SUPPORTED_VERSIONED_PARSERS and takes the plain day-by-day merge path.
--- a/packages/frontend/src/app/api/submit/route.ts
+++ b/packages/frontend/src/app/api/submit/route.ts
@@ -847,6 +847,17 @@ export async function POST(request: Request) {
             `Ignored ${label} changes because this parser generation or partial snapshot cannot safely advance the device high-water.`
           );
         }
+        if (plan.highWaterDeficit) {
+          // Without this the submit is indistinguishable from a successful
+          // one: the client is listed as scanned, the request returns 200, and
+          // nothing says the device has been contributing zero for this client
+          // since its scan fell below the credited high-water.
+          warnings.push(
+            `Added no ${label} usage because this scan reports ${plan.highWaterDeficit.toLocaleString(
+              "en-US"
+            )} fewer lifetime tokens than this device has already been credited. Local history that is deleted or aged out of the client's own store stays credited, so nothing is added until the scan exceeds the stored high-water again.`
+          );
+        }
       }
       const plannedIncrementClients = [...parserPlans].filter(
         ([, plan]) =>
--- a/packages/frontend/src/lib/db/parserHighWater.ts
+++ b/packages/frontend/src/lib/db/parserHighWater.ts
@@ -127,6 +127,13 @@ export interface ParserHighWaterPlan {
   /** Absolute cells when `mode` is `replace`; omitted otherwise. */
   layoutDays?: Record<string, ClientBreakdownData>;
   nextState?: ParserClientHighWaterState;
+  /**
+   * Lifetime tokens the credited ledger holds beyond what this snapshot
+   * reports. Positive means no growth was allocatable and none can be until
+   * the parser reports at least this much more, which is indistinguishable
+   * from a re-attribution and so is never credited.
+   */
+  highWaterDeficit?: number;
 }
 
 function copyModels(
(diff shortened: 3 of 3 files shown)


shunkakinoki:
Follow-up with the actual fix: #1317 (branched off this one). This PR makes the loss visible; #1317 stops it by narrowing the lifetime baseline to the credited days the client's store can still reach.

junhoyeo (approved):
Thanks!

PR #1317, opened 2026-09-08
fix(submit): bound the parser high-water by the store's retention floor

shunkakinoki:
Full scans of a rolling conversation store can stay below the device's credited lifetime after old sessions are pruned, leaving no token budget for new usage. This accepts and calendar-validates optional `scanScope.retentionFloors[client]` metadata for a future retention-aware protocol.

The current server deliberately keeps the full lifetime high-water even when a client reports a floor. A client-controlled floor cannot prove that old history was pruned rather than re-dated into the incoming snapshot, so using it to reduce the baseline could duplicate previously credited usage. Calendar-invalid floor values are rejected, and token-deficit warnings distinguish tokens from independently credited messages.

Updated against current main, including the merged deficit warning from #1316, with both merge conflicts resolved. Tests cover coordinated floor advances with re-dated snapshots, legacy baselines, partial scans, calendar validation, and positive message increments during a token deficit.

Related: #1315. Retention recovery is intentionally not enabled: a future implementation must persist and validate an authenticated, monotonic store-boundary transition before allowing a floor to relax the high-water. A CLI-reported timestamp alone is insufficient under the current submission trust model.

Validation:

- Frontend suite: 1,028 tests passed.
- TypeScript and ESLint on all changed files passed.
- PostgreSQL 16 migration replay and 27 reconciliation/moderation integration tests passed.
- Calendar, message-warning, and moved-cell regressions were reproduced before their fixes.

### DIFF
--- a/packages/frontend/__tests__/api/submitAntigravityCliHighWater.test.ts
+++ b/packages/frontend/__tests__/api/submitAntigravityCliHighWater.test.ts
@@ -648,16 +648,15 @@ describe("POST /api/submit antigravity-cli re-attribution high-water", () => {
     ).toBe(true);
   });
 
-  it("credits the new window when the reported retention floor clears the stored days", async () => {
+  it("keeps the lifetime bound when an unverified retention floor clears the stored days", async () => {
     const store = newStore();
     installTx(store);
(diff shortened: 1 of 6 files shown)


junhoyeo (approved):
Thanks a lot!

PR #1318, opened 2026-09-08
feat(clients): add Meept session parsing

bhodgens:
# feat(clients): add Meept session parsing

## Summary

Adds [Meept](https://github.com/bhodgens/meept) (a Go agent daemon) as a supported tokscale client. Meept records one `llm_calls` row per LLM provider call in a SQLite database at `~/.meept/metrics.db`, with per-call session attribution, agent identity, and the full token breakdown including reasoning and cache-creation tokens.

- **Client registry**: `Meept` (`id: "meept"`, root `~/.meept`, file `metrics.db`, `submit_default: true`)
- **Parser**: `sessions/meept.rs` reads `llm_calls` read-only and projects it to `UnifiedMessage`:
  - `tokens_sent` → input, `tokens_received` → output, `tokens_cached` → cache_read, `cache_creation_tokens` → cache_write, `reasoning_tokens` → reasoning
  - error rows (`error = 1`) excluded; they carry no usage by construction, and the filter stays explicit so a schema change cannot silently inflate totals
  - dedup keys derive from the monotonic row id (`meept:<id>`), so identical calls are distinct work
  - ISO-8601 UTC timestamps parsed defensively; malformed/legacy shapes fall back to epoch rather than dropping the row
  - a pre-migration database (no session_id/reasoning/cache-creation columns) yields no messages instead of logging a failure per install
  - agent_id flows into the UnifiedMessage agent field for per-agent reporting
- **Cache identity**: explicit `parser_version` arm at v1, per the intentional no-`_ => 1` contract in `message_cache.rs`
- **CLI**: `ClientFilter::Meept` with `--client meept`, TUI hotkey `W`
- **Frontend**: `SUPPORTED_CLIENT_TYPES` + `SOURCE_DISPLAY_NAMES` / `SOURCE_LOGOS` / `SOURCE_COLORS` (amber `#F59E0B`, matching the logo), so submissions containing meept pass the submit schema
- **Assets**: `.github/assets/client-meept.png` (400×400, square-cropped from the meept repo's art) mirrored at `packages/frontend/public/assets/logos/meept.png`
- **Docs**: client table, platform matrix, multi-platform lists, and `--client` values in README.md plus the ko/ja/zh-cn localized READMEs

## Verification

- Parser unit tests cover the fixture contract, row-scoped dedup keys, agent/session pass-through, timestamp parsing (including malformed input), and legacy-schema tolerance
- Validated end-to-end against a live meept database (3,550 rows, 22.0M input / 4.8M output tokens): every reported total matches direct SQL `SUM()` exactly, per-session grouping attributes task/plan/chat sessions correctly, and hourly/monthly reports bucket on the parsed timestamps
- `cargo fmt --all -- --check`, `cargo clippy --locked --workspace --all-features -- -D warnings`, `cargo test --workspace --all-features`, `bun run --cwd packages/frontend test -- __tests__/lib/clientRegistry.test.ts`, and `bun run --cwd packages/frontend typecheck` all pass

---
## Summary by cubic
Adds Meept as a supported client by parsing per-call LLM usage from `~/.meept/metrics.db`, which was previously untracked. Reads rows read-only, maps token columns to the standard breakdown, tolerates legacy schemas without logging failures, and routes through the SQLite source cache so warm scans skip re-reading the database.

- Registers `meept` in the client registry, CLI `--client meept`, TUI hotkey `W`, and frontend constants with amber `#F59E0B` branding.
- Dedup keys derive from the monotonic row id so identical calls count as distinct work; malformed timestamps fall back to epoch instead of dropping rows.
- Requires a complete scan; incomplete results yield no messages rather than reporting undercounted totals.
- Agent and session ids flow into the unified message for per-agent reporting.
- Locks the parser cache identity at version 1.
- Documents the client across all READMEs and adds the logo asset.

<sup>Written for commit f250e272fe3b8caf735a81d79c95f0e1a252d0c3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/junhoyeo/tokscale/pull/1318?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bhodgens:
Addressed all review findings in f250e272:

- **P1 timestamp panic** (meept.rs): the length gate is now char-bounded (`chars().count() >= 19`), so multi-byte garbage falls back to epoch instead of panicking on slice. Regression test with a Japanese-string timestamp added.
- **P1 partial scan** (meept.rs): `scan.ran()` → `scan.completed()` — an Incomplete prefix no longer converts to messages that would undercount totals.
- **P2 clamped-invalid timestamps** (meept.rs): out-of-range fields (`2026-13-40T25:99:99Z`) are now rejected to the epoch fallback rather than silently clamped to a wrong-but-valid instant. Regression tests added.
- **P2 uncached parse lane** (lib.rs): the meept dispatch now routes through `load_or_parse_sqlite_source` with `CacheIdentity::for_client(ClientId::Meept)`, so warm scans skip the DB when its fingerprint is unchanged. Verified identical totals post-change (3,550 msgs / 30,281,909 tokens vs direct SQL).
- **P3 hotkey comment** (client_ui.rs): rationale corrected.
- **P3 zh-cn Windows matrix**: Meept row added — and same gap closed in README.ko.md and README.ja.md, which had it too.

All gates re-run green: fmt, clippy -D warnings, workspace tests, meept unit suite (6).

Issue #851, opened 2026-07-10
feat: Support Codex Desktop App (ChatGPT.app) usage tracking


hellokun985:
## Feature Description

Support tracking token usage from the **Codex Desktop App** (`com.openai.codex`, marketed as "ChatGPT" on macOS) as a separate client source, in addition to the existing Codex CLI (`codex`) support.

## Motivation

The Codex Desktop App (ChatGPT.app) is a separate product from the Codex CLI. Many users run both, but tokscale currently only tracks CLI sessions via `~/.codex/sessions/` JSONL files. Desktop App usage data is stored in `~/.codex/state_5.sqlite` (a SQLite database with `threads` table containing `tokens_used`, `model`, `model_provider`, etc.) but tokscale does not read this data source.

This means a significant portion of Codex usage is invisible in the tokscale dashboard — the "Agents" tab in TUI shows zero for `codexapp` even though the user actively uses the desktop app.

## Proposed Solution

1. Add a new client identifier (e.g. `codexapp` or `codex-desktop`) that reads from `~/.codex/state_5.sqlite`
2. Parse the `threads` table schema:
   - `id` (TEXT PRIMARY KEY)
   - `title` (TEXT)
   - `source` (TEXT) — values include `vscode`, `exec`, etc.
   - `model_provider` (TEXT)
   - `model` (TEXT)
   - `tokens_used` (INTEGER)
   - `created_at` / `updated_at` (INTEGER, unix timestamps)
3. Attribute token usage to the correct model and provider
4. Distinguish desktop app sessions from CLI sessions (both share `~/.codex/` directory but use different storage formats)

```bash
# Example: see current Codex App data
sqlite3 ~/.codex/state_5.sqlite "SELECT id, title, model, tokens_used FROM threads WHERE tokens_used > 0 ORDER BY updated_at DESC LIMIT 5;"
```

```
┌─────────────────────────────┬──────────────────────┬────────────┬────────────┐
│ id                          │ title                │ model      │ tokens_used│
├─────────────────────────────┼──────────────────────┼────────────┼────────────┤
│ 019f49d8-...                │ 判断是否卸载重复...    │ gpt-5.6-sol│     492453 │
│ 019f3599-...                │ 排查 herdr 安装问题   │ gpt-5.5    │     206280 │
│ 019f3179-...                │ appstoreagent和dasd  │ gpt-5.5    │    4301464 │
└─────────────────────────────┴──────────────────────┴────────────┴────────────┘
```

[...]

PR #1353, opened 2026-09-20
fix(mcode): parse stream-json completion envelopes

### LINKED ISSUE #1352: mcode: headless capture parses 0 messages on MiniMax Code 0.4.12; proposal to source from the local runtime store
## Summary

Two findings and one proposal.

1. **Bug.** The headless `mcode` client returns zero messages against a real `mcode exec --output-format stream-json` capture from MiniMax Code 0.4.12. The parser matches top-level `{"type":"message"}` and `{"type":"exec.result"}` events; 0.4.12 emits envelope events (`item.*`, `turn.completed`, `exec.completed`) with usage at `turn.completed.usage` and the result nested at `exec.completed.result`. It fails silently — no error, just zero tokens.
2. **The store looks like a more complete and durable runtime source, but it does not preserve the originating surface.** The desktop app and the `mcode` CLI write the same local runtime store, whose per-message records carry `model`, `provider` and usage; aggregating them by `turn_id` reproduced the stream's per-turn numbers on all three turns we could compare. It is data the app writes for itself — its layout is versioned too (see §2) — and it cannot say which surface a session came from.
3. **Proposal.** Make the store the primary source for `mcode`, keep the headless path as a second scan root under the same client, and settle the ownership rule with you rather than resting it on token equality (§3). This is a request for the integration boundary you described in #851, not a claim that we already have the right shape.

We are working on this locally and would rather get your read on the boundary before we send a PR. The details, the raw samples (sanitized) and our planned direction are below, and the five questions we would like your call on are in §3.

## Environment

- tokscale `4.17.0` (`d8fd670a`) against `junhoyeo/tokscale` main
- MiniMax Code desktop `3.0.73` (`com.minimax.agent.cn`, macOS 26.3.2 arm64)
- `mcode` CLI `0.4.12` from npm `@minimax-ai/code` (published 2026-09-18), installed via `npm i -g`

### How we obtained this

- Everything below comes from **one macOS machine** running the **CN build** of the desktop app, with the CLI at region `cn`. We have not checked the global build, another platform, or a machine with a long-lived store — the desktop install here is from the same day, so we have no samples of migrated or legacy sessions.
- The static claims (paths, formats, which field holds what) come from unpacking the desktop app's `app.asar` with the official `@electron/asar` and unpacking the npm tarball, then reading the shipped runtime code. No MiniMax source or SDK was needed.
- The empirical claims come from that machine's own store: six session directories in total — one from the desktop app, four from `mcode exec` (a plain reply, a tool-using turn, one run continued with `--continue`, and the run that spawned the sub-agent), and the sub-agent's own `task` session — plus the SQLite table.
- The 0-message result comes from running the released parser against a capture we produced ourselves; that capture is what the wrapper records, since it injects the same `--output-format stream-json` flag.
- Nothing in tokscale was modified to gather this. One throwaway integration test file was used to call the parser and was deleted afterwards.
- Windows is the one thing we state from code rather than observation, and it is flagged as such below.

## 1. The headless capture parses 0 messages

<details>
<summary>Reproduction, the envelope format it emits, and a minimal fix</summary>

### Reproduction

```sh
mcode exec --output-format stream-json --cwd /tmp/probe "say hi in three words" > stream.jsonl
```

`tokscale headless mcode exec ...` injects `--output-format stream-json` (`crates/tokscale-cli/src/main.rs:7085`), so this is the shape the wrapper captures.

Running the released parser over that capture (`crates/tokscale-core/src/sessions/mcode.rs`):

```
PROBE_RESULT: parsed 0 message(s)
```

(That was run through a throwaway integration test calling `parse_mcode_file` on the capture, in an otherwise unmodified checkout.)

The wrapper itself still reports success: it prints `✓ Saved headless output to <path>` and exits 0, so the loss only surfaces later, as usage that never appears in a report.

### Why

Every line in our capture is an envelope:

```json
{"schemaVersion":1,"sequence":15,"timestampMs":1789890580193,"runId":"exec_turn_...","sessionId":"mvs_...","turnId":"turn_...","type":"exec.completed","result":{...}}
```

`type` values observed in one capture: `exec.started`, `session.started`, `turn.started`, `item.started`, `item.updated`, `item.completed`, `turn.completed`, `exec.completed`. There is no top-level `message` and no top-level `exec.result`.

In the captures we took, usage appears in two event locations:

- `usage` on the `turn.completed` event (a direct field of the envelope) — per turn, e.g. `{"inputTokens":12710,"outputTokens":24,"reasoningTokens":0,"cacheReadTokens":128,"cacheWriteTokens":0,"totalTokens":12734}`
- `usage` under `exec.completed.result` — the same numbers, next to `exec.completed.result.model` = `{"providerId":"minimax","modelId":"MiniMax-M3","variant":"thinking","providerSource":"provider","providerKind":"minimax-managed"}`

The parser switches on the top-level `type`, so `Some("message")` and `Some("exec.result")` never match and the function returns an empty vector. The unit-test fixtures in `sessions/mcode.rs` use the flat shape (top-level `{"type":"message","message":{...usage...}}` and top-level `{"type":"exec.result",...}`), which is presumably what an earlier CLI emitted. We could not date the change: the CLI's bundled changelog does not mention the output shape, and we have no older CLI build to compare against.

While investigating we also noticed the pattern that `prepare_headless_args` routes every non-Codex source through an `else` branch — the same concern raised during #1136 review — so this fix and any future headless client touch the same seam.

### Minimal fix if the headless path stays

Map the new shape onto the existing message model: read `usage` from `turn.completed` (or `exec.completed.result.usage`), take `providerId`/`modelId` from `exec.completed.result.model`, and key by `sessionId`/`turnId`, which the envelopes already carry at top level. The four token buckets map one-to-one; `reasoningTokens` is present in the envelope and is currently hardcoded to `0` in the parser. Existing captures may use the flat shape, while 0.4.12 captures use envelopes, so the fix should keep reading both. Otherwise upgrading tokscale could silently drop captures a user already has, and we would rather not trade one silent zero for another.

</details>

## 2. What the store actually contains

<details>
<summary>Paths, record shape, and the evidence that desktop and CLI share one store</summary>

### Location

Root is `~/.minimax`, resolved as `MINIMAX_DATA_DIR` → `MAVIS_DATA_DIR` → `~/.minimax`. On Windows it is `%USERPROFILE%\.minimax`: neither the desktop app nor the CLI has a platform branch for the data root (`LOCALAPPDATA` appears only for the CLI install root), and neither branches on region, so the CN and global builds should agree here. The desktop documentation states the same default.

One wrinkle: the desktop can relocate the directory. It persists `localRuntimeDataParentDir` in its own config and joins the same `.minimax` name onto that parent, so `~/.minimax` is not the only possible root (the machine we tested has never relocated and the field is absent from its config). A scanner that wants to cover relocated installs has to read that setting.

```
<root>/v2/sessions/<YYYY>/<MM>/<DD>/<HH-MM-SS-mmm>-session_<base64url(sessionId), unpadded>/
├── manifest.json                 {"schemaVersion":1,"sessionId":"mvs_...","createdAtMs":...,"source":"local-runtime","layout":"v2-final-dated-session","paths":{...}}
├── messages.jsonl                canonical history, one JSON object per line
├── history-catalog.json          {"activeGeneration":0,"activeRevision":"sha256:...","artifacts":[{... "fileName":"messages.jsonl","byteLength":...,"messageCount":...}]}
├── user-message-locators.jsonl   byte offsets for user messages
└── llm-call.json                 request snapshot (api/model/provider/systemPrompt/tools), no usage
<root>/v2/sqlite/runtime-state.sqlite   → table local_runtime_token_usage (WAL)
```

`manifest.json` also declares `ledger.jsonl`, `display.jsonl`, `snapshot.json` and `reports/`; none of them existed in the sessions we sampled. The runtime's own released-history reader treats `ledger.jsonl` plus `snapshot.json` as an older history format it replays for migration and recovery, so a parser should not expect them to be present.

The layout is versioned and still moving, so a parser should key off what it finds rather than assume one shape. The 3.0.73 we tested writes `v2/sessions` plus `v2/sqlite/runtime-state.sqlite`, while the same app's bundled code already declares a newer directory contract for `v2` whose database is `v2/chats/local-runtime.sqlite` with `ledgers/` and `snapshots/` beside it — directories this build does not create.

### Message record

Only assistant rows carry usage. From a real session:

```json
{"message_id":"msg-...","turn_id":"turn_...","message":{"role":"assistant","model":"MiniMax-M3","provider":"minimax","api":"anthropic-messages","content":[{"type":"thinking"},{"type":"text"}],"usage":{"input":12710,"output":24,"cacheRead":128,"cacheWrite":0,"totalTokens":12862,"cost":{"input":0,"output":0,"cacheRead":0,"cacheWrite":0,"total":0}},"timestamp":1789890578370,"responseId":"...","stopReason":"stop"}}
```

Two schema notes worth recording:

- `cost` is zero in every sample we took; whatever pricing the app performs is not persisted here, so cost has to come from tokscale's own tables (the headless parser already passes `0.0`).
- `totalTokens` is not the same quantity as the stream's `totalTokens`: here it includes `cacheRead` (`12710 + 24 + 128 = 12862`), while the stream's excludes it (`12734`). A parser should use the four buckets and ignore `totalTokens`.

### The SQLite table, for completeness

```sql
CREATE TABLE local_runtime_token_usage (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  session_id TEXT NOT NULL, agent_name TEXT NOT NULL, framework_type TEXT NOT NULL,
  turn_id TEXT, model TEXT, ts INTEGER NOT NULL,
  input_tokens INTEGER NOT NULL, output_tokens INTEGER NOT NULL, reasoning_tokens INTEGER NOT NULL,
  cache_read_tokens INTEGER NOT NULL, cache_write_tokens INTEGER NOT NULL,
  cost_usd REAL, raw TEXT);
```

It is a best-effort projection of the committed history (the runtime's own `usage/README.md` says a process that exits between the history commit and the usage append is allowed to leave the row missing, with no durable watermark or startup reconciliation). We also observed `model` as `NULL` in every row we saw — eleven rows at the time of writing, spanning both surfaces and both agents. So the JSONL is the better source: `messages.jsonl` is described in the runtime's own code as the canonical history, and it carries `model` and `provider` per record.

### Desktop and CLI write the same store

Three independent levels of evidence:

- **Same files.** A desktop session and a CLI session produce the same five files with the same shapes, and `manifest.json.source` is hardcoded `"local-runtime"` in both (the writer is shared runtime code).
- **Same records.** The assistant rows are field-for-field identical in structure.
- **Same prompt text.** The first user message in both contains a byte-identical `<system-reminder><agent-context>` block (`agent: Mavis`, `agentRole: orchestrator`, `SESSION ROLE: root`), differing only in session id and date. The npm CLI ships its own bundle, but it embeds the same local runtime code — the manifest writer, the session layout constants and the usage projector are recognisably the same code — so the artifacts come out identical and there is no textual difference to key on either.

**Consequence: the store cannot attribute a session to desktop versus CLI.** We looked for a marker everywhere one might live:

- `manifest.json` — only `source: "local-runtime"`, hardcoded by the shared writer
- the session record — `sessionOrigin` is only `local-runtime | legacy-opencode | unknown`
- `extra_data_json` — `origin: "user"`, `appMode: "coding"`
- the message rows — `source: "api"`
- `framework_type` — `pi-agent`
- `llm-call.json` — identical top-level keys on both surfaces
- on disk — `runtime-owner-leases/<uuid>.owner` holds only `{"pid":...,"startToken":...}`

The runtime does have the concept of a runtime owner kind (`electron` / `cli` / `tui`), but it is a host option held in memory and never persisted.

If the CLI and desktop must be reported separately, that requires an upstream change from MiniMax. Our reading is that they should be one client, and we would rather hear your view before assuming it.

### Granularity, and why it lines up with the stream

The store records one row per assistant message; the stream reports one usage per turn. A tool-using turn therefore produces several rows. Summing them by `turn_id` reproduces the stream's `turn.completed.usage` exactly — verified on three turns where we have both sides:

| turn | store rows | store sum (in/out/cacheRead/cacheWrite) | stream `turn.completed.usage` |
| --- | --- | --- | --- |
| plain reply | 1 | 12710 / 24 / 128 / 0 | 12710 / 24 / 128 / 0 |
| tool call | 2 | 12727+151 = 12878 / 60+416 = 476 / 128+12855 = 12983 / 0 | 12878 / 476 / 12983 / 0 |
| `--continue` follow-up | 1 | 423 / 99 / 13561 / 0 | 423 / 99 / 13561 / 0 |

That equivalence is what makes the dedup argument below work for these captures — §3 says where it does not generalise.

One invariant we have not verified: that every assistant row within a turn carries the same `model`/`provider`. If a turn can mix identities (a BYOK turn, a fallback, a mid-turn switch), then "sum the buckets and take the record's identity" is not well defined, and the parser needs an explicit rule — we would either take the identity of the row that carried the usage or split by identity instead of summing. Worth settling before anyone writes the parser.

### Sub-agent and task sessions

They are ordinary sessions with their own directory and their own usage rows. A sub-agent run we triggered produced a session with `agent_name = explore`, `session_kind = task`, `parent_session_id` set, and its own usage rows in the same table. On the JSONL side the same distinction appears as text in the first user message: `SESSION ROLE: root` for user-initiated sessions versus `SESSION ROLE: branch` for the task session (SQLite additionally has `session_kind`, `parent_session_id`, `visibility`).

We think these should be **counted**: the tokens were really spent, and dropping them would under-report the tool. Classification by `agent_name` or `SESSION ROLE` can come later without changing the ingest.

</details>

## 3. Proposal and the boundary question

Reading your reply in #851 — integration boundary narrower than a new client, no ingesting of a shared store that double-counts what tokscale already parses, and a Desktop-specific client only once a product-origin marker exists — we think MiniMax Code can meet the "one owner" requirement with machinery that already exists, and we would like your view on whether that is enough given there is **no** product-origin marker.

### Reuse

- One client id with several scan roots is already the idiom (`ClientId::Codex` has three: `sessions`, `archived_sessions`, plus the headless capture root). `mcode` can gain a store root the same way.
- `scan_directory(root, pattern)` walks recursively and matches on filename, so `pattern: "messages.jsonl"` works as-is; Codex's `~/.codex/sessions/**/*.jsonl` is the same dated-directory shape.
- Parsing helpers are shared: `lossy_lines`, `file_modified_timestamp_ms`, `normalize_timestamp`, `UnifiedMessage::new_with_dedup`.
- The source cache, pricing, reports and submit paths are untouched: the record already carries `model` and `provider`, so nothing has to wait for a terminating event.

### One owner

`mcode` currently runs on `parse_cached_lane` (`crates/tokscale-core/src/lib.rs:2457`), which does not dedup across files, while `parse_cached_lane_deduped` (`lib.rs:1571`) does — its contract says parsers using it must supply globally stable dedup keys and that the pass is first-wins in scan-path order, with keys surviving a warm cache hit.

The existing `mcode` parser already emits `mcode:{session_id}:{turn_id}:{index}:{in}:{out}:{cache_read}:{cache_write}`. Once the headless side reads `turn.completed` — one usage per turn, so index 0 there — a store parser that aggregates by `turn_id` produces the same key for the same call: index 0 on both sides, with the four buckets equal as shown above.

But that key carries the payload, not just the identity, and that is where we think the argument needs care:

- **A one-token disagreement breaks it.** If a future runtime version counts a token differently, or commits at a different moment, the two sources produce `input=12878` and `input=12879` for the same call. The keys then differ, and the copy is counted twice instead of collapsing — the mechanism fails in the direction that inflates totals.
- **Flat-shape captures may not participate.** A flat-shape capture can emit several messages for one turn (`index` 0, 1, …) with partial usage, while a store parser would emit one aggregate at `index` 0. Their keys cannot match, so those copies would all survive.

So the honest statement is narrower than our first draft: for new envelope captures where both sources expose one aggregate per turn, the existing key collapsed the copies in every sample we measured; for legacy captures, and for any case where the two sources disagree numerically, it does not. We would rather not build "one owner" on token equality at all. The options we see are: an identity-based key (`session_id` + `turn_id`) applied only after both sources are normalised to one record per turn — on its own it would collapse a flat capture's several real calls within one turn against each other; keeping the store parser at message granularity so it lines up with flat captures; or declaring a source precedence and filtering the losing source instead of deduplicating by value. This is the fifth question below.

### What we would like your call on

1. **Boundary.** Is "store as primary + headless as a second root on the same client id, deduped by key" acceptable in the absence of a product-origin marker, or would you rather see the headless path retired so the store is the single source?
2. **Fix order.** We read the 0.4.12 parse failure as a separate, reportable bug. Would you prefer a small fix issue/PR first (envelope support), with the store work following, or one combined change?
3. **Sub-sessions.** We plan to count `session_kind = task` usage. If you would rather they be filtered or separately labelled, that changes the parser's emit shape and we would rather know now.
4. **Client id.** Reuse `mcode` / "MiniMax Code" for the store source, or introduce a separate id?
5. **Ownership rule.** If the store becomes primary, would you accept an identity-based key (`session_id` + `turn_id`) applied after both sources are normalised to one record per turn, a store parser that stays at message granularity to line up with flat captures, or a declared source precedence (one source wins, the other is filtered)? We would rather decide this than assume it.

## 4. What we are doing locally

<details>
<summary>The changes we have in progress</summary>

- A second scan root for `ClientId::Mcode` pointing at `<data-dir>/v2/sessions` with `pattern: "messages.jsonl"`, where the data dir resolves as `MINIMAX_DATA_DIR` → `MAVIS_DATA_DIR` → `~/.minimax`. The existing `PathRoot::EnvVar` takes a single variable, so that needs a small extension, and relocated desktop installs would need the persisted parent dir as well.
- Two wiring points need the new root: the main scan (`lib.rs:2457`) and the second parse inside `parse_local_clients` (`lib.rs:6050`), which both `tokscale clients` and the report wiki population go through.
- A per-turn store parser: collect assistant rows, group by `turn_id`, sum the four buckets, and take `model`/`provider` from the record (supporting several `providerId` values, not just `minimax`). The session id has to come from the sibling `manifest.json` — the row itself carries only `message_id` and `turn_id` — and not from decoding the directory name or scraping the prompt text. The emit shape's dedup key is question 5, so we have deliberately left it unfixed here.
- A headless fix for the envelope format, so the current CLI stops reporting zero.

We have not opened a PR for any of this yet. Any guidance on the five questions above — especially the boundary and the client id — is welcome before we do, and we are happy to split this into separate issues if that suits the tracker better.

</details>

## Appendix — sanitized samples

<details>
<summary>Raw envelope, store record, prompt block and SQLite rows</summary>

Envelope events (one capture, `type` values in order):

```
exec.started → session.started → turn.started
→ per-item lifecycle: item.started … item.updated ×N … item.completed, one per reasoning / agent_message item
→ turn.completed{durationMs,model,usage} → exec.completed{result{…, model, usage}}
```

`turn.completed` (sanitized):

```json
{"schemaVersion":1,"sequence":14,"timestampMs":1789890580193,"runId":"exec_turn_...","sessionId":"mvs_<redacted>","turnId":"turn_<redacted>","type":"turn.completed","durationMs":2078,"model":{"providerId":"minimax","modelId":"MiniMax-M3","variant":"thinking","providerSource":"provider","providerKind":"minimax-managed"},"usage":{"inputTokens":12710,"outputTokens":24,"reasoningTokens":0,"cacheReadTokens":128,"cacheWriteTokens":0,"totalTokens":12734}}
```

Store assistant record for the same turn (sanitized; paths and ids redacted):

```json
{"message_id":"msg-<redacted>","turn_id":"turn_<redacted>","message":{"role":"assistant","model":"MiniMax-M3","provider":"minimax","api":"anthropic-messages","content":[{"type":"thinking"},{"type":"text"}],"usage":{"input":12710,"output":24,"cacheRead":128,"cacheWrite":0,"totalTokens":12862,"cost":{"input":0,"output":0,"cacheRead":0,"cacheWrite":0,"total":0}},"timestamp":1789890578370,"responseId":"<redacted>","stopReason":"stop"}}
```

First user message, both surfaces (sanitized; the block is byte-identical apart from id and date):

```
<system-reminder>
<agent-context>
  agent: <agent name>
  user: <user name>
  agentName: mavis
  agentRole: orchestrator
  SESSION ROLE: root
  YOUR SESSION ID: mvs_<redacted>
  date: <timestamp>
</agent-context>
```

SQLite rows (sanitized), showing `model` NULL on every row:

```
('mvs_<redacted>', 'mavis',   'pi-agent', NULL, 1789888602489, 24548, 136,   0, 128,   0, 0.0)
('mvs_<redacted>', 'mavis',   'pi-agent', NULL, 1789890715319,   151, 416,   0, 12855, 0, 0.0)
('mvs_<redacted>', 'explore', 'pi-agent', NULL, 1789890747749,  3907, 180,   0, 128,   0, 0.0)
```

Not covered by our sampling, in case it matters to you: interactive TUI sessions, cron or IM-channel sessions, BYOK providers other than MiniMax, and any Windows or Linux machine — the Windows path above is derived from the shipped code rather than observed.

</details>

REMIXPLAYER:
## Summary

- Parse the final `exec.completed` envelope emitted by MiniMax Code 0.4.12, reading provider/model and input, output, cache-read, and cache-write token counts from its nested `result`.
- Preserve the legacy `message` plus top-level `exec.result` path introduced in #1136.
- Continue ignoring incomplete or unattributed streams instead of guessing a model.

## Root cause and fix

The released parser expects top-level `message` events followed by a top-level `exec.result`. MiniMax Code 0.4.12 instead projects envelope events and finishes a successful run with `turn.completed` followed by `exec.completed`, with the actual `exec.result` nested under `exec.completed.result`. The old branches therefore never match and a valid capture parses as zero messages.

This change reads model identity and usage directly from the nested final result. It intentionally ignores `turn.completed`: 0.4.12 projects the same turn usage into both completion events, so consuming both would double-count one run. The legacy path is otherwise unchanged. Both formats still require a final provider/model; the envelope path additionally requires usage in the nested result.

`reasoningTokens` handling is unchanged and remains reported as zero, matching the existing mcode parser. Changing token-accounting semantics is outside this compatibility fix.

## Investigation and verification

1. Reproduced the zero-message result with a real `mcode exec --output-format stream-json` capture.
2. Inspected the shipped `@minimax-ai/code@0.4.12` projector and output encoder to confirm that `stream-json` emits the nested final result and does not mix the legacy flat events into the same valid stream.
3. Cross-checked a plain reply, a tool-using turn, and a `--continue` turn; in each case, `turn.completed.usage` and `exec.completed.result.usage` matched.
4. Added a regression test that confirms `turn.completed` alone emits nothing and the following `exec.completed` emits exactly one message.
5. Ran a fresh one-step request returning `hi`: mcode emitted `MiniMax-M3` usage of 12,718 input, 21 output, 128 cache-read, and 0 cache-write tokens, and the patched Tokscale CLI parsed exactly one message with the same four buckets.

## Scope

Addresses only the headless parsing bug described in #1352. This PR does not add runtime-store scanning, change scan roots or client identity, or choose an ownership/deduplication policy. Those boundary questions remain open, so this PR intentionally does not close #1352.

## Validation

- `cargo test --locked --workspace --all-features`
- Fresh `mcode exec --output-format stream-json` capture parsed end-to-end by `tokscale --no-spinner --json --client mcode`


junhoyeo (approved):
Reviewed the completion-envelope compatibility change and its upstream event contract. The current head has passing Rust tests, lint, rustdoc, native builds, launcher and container checks, with no unresolved review threads.

Approving the scoped compatibility fix. Independent GPT-6 Astra review also identified a pre-existing cross-file capture deduplication gap in the aggregation paths; it will be addressed with regression coverage in the requested consolidated follow-up PR alongside #1354 and #1349.

REMIXPLAYER:
Thanks for the review and merge — really appreciate you taking the compatibility fix.

One quick follow-up on #1352: the runtime-store / ownership-boundary part was intentionally kept out of this PR, since #1353 only fixes the completion-envelope parsing regression.

I noticed #1352 was closed after the merge, so I just wanted to make sure we're interpreting that correctly: does the closure mean you'd prefer not to pursue the shared runtime-store source for now, given the origin/ownership constraint from #851, or was it simply closing out the parsing bug resolved here?

We mainly want to avoid taking the next step in a direction that doesn't fit the project.

junhoyeo:
The requested contributor PRs (#1353, #1354, and #1349) are now merged. Consolidated fixes and regression tests from the independent GPT-6 Astra reviews are in https://github.com/junhoyeo/tokscale/pull/1355.

The follow-up preserves the original features and covers MiMo history/submission accounting and additional scan roots, MiniMax copied-capture deduplication, and Muse provider/pricing attribution. It is intentionally left open for review; deploy its receiving changes before publishing the MiMo desktop split.

junhoyeo:
Thanks @REMIXPLAYER for fixing MiniMax Code's completion-envelope parsing, and for the detailed investigation against real captures and the upstream event format! This is now merged into `main`. 🙌

I also opened #1355 as a consolidated follow-up. For MiniMax Code, it addresses the pre-existing aggregation gap where copies of the same captured stream could be counted twice, with regression coverage for both report and count paths.

Really appreciate the contribution! If Tokscale has been useful to you, a ⭐ on [the repo](https://github.com/junhoyeo/tokscale) would also mean a lot—it helps others discover the project.

